Issue #58, opened 2013-02-26
Create Token from customer id


agrigg:
In order to create a shared customer for Stripe Connect the ability is needed to create a token using a customer id instead of the card info as outline on

https://stripe.com/docs/connect/shared-customers

The Stipe call is supposed to look something like:

Stripe::Token.create(
  {:customer => CUSTOMER_ID},
  ACCESS_TOKEN # user's access token from the Stripe Connect flow
)

Seems like the StipeTokenCreateOptions needs to be updated to allow a customer id to be specified.

PR #60, opened 2013-03-01
Add CustomerId to StripeTokenCreateOptions for generating tokens for shared customers

agrigg:
The CustomerId was missing from the StripeTokenCreateOptions so that tokens can be created for share customers using the Stripe create token api.

https://stripe.com/docs/api#create_token

This is to address Issue #58 which I submitted.


nhwilly:
Thanks!  Was looking for this one.

Issue #135, opened 2014-01-21
Add support for "Stripe-Version" parameter


ghost:
See https://stripe.com/docs/api#versioning. It mentions a "Stripe-Version" parameter which can be used to specify the version of the API call, without the need to change the version account-wide.

PR #159, opened 2014-04-21
Add Stripe-Version header based on API supported by library

richardlawley:
This is an alternate suggestion for #135.  I disagree with specifying the supported version in a config file in the calling app (as used in #147), as users will not necessarily know that they need to change this following an update, and at the end of the day it's this library that's compatible with a particular Stripe API version, not their calling app.

This change should allow a user to update Stripe.net without the needing to upgrade their account in their Stripe dashboard.  Unit test included to verify that specifying a lower version changes the response from Stripe.  Unable to verify that specifying a higher version than supported works, since my test accounts are on the latest version and you can't create one on an older version.

JSON.net has been included in the tests to work around a strange problem where ncrunch would not build the test project.

### DIFF
--- a/src/Stripe.Tests/Stripe.Tests.csproj
+++ b/src/Stripe.Tests/Stripe.Tests.csproj
@@ -98,6 +98,7 @@
     <Compile Include="recipients\when_listing_recipients.cs" />
     <Compile Include="recipients\when_updating_a_recipient.cs" />
     <Compile Include="requestor\when_creating_a_request.cs" />
+    <Compile Include="requestor\when_requesting_an_old_api_version.cs" />
     <Compile Include="requestor\when_requesting_a_blacklisted_certificate.cs" />
     <Compile Include="requestor\when_requesting_a_mismatched_certificate.cs" />
     <Compile Include="tokens\when_creating_a_token_with_a_customer.cs" />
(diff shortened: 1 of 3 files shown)


benwmills:
I see right now that the referenced version is 2014-06-13.  That means that Stripe.net is not designed to use any of the new features since then right?  And upgrading the API version in the website will make no difference.  This isn't a problem.  I just wanted to make sure that I'm understanding correctly.

jaymedavis:
Correct. That's a bit outdated, eh? I need to get some time in! A lot of the updated features have been added to Stripe.net, it is just not verified as a whole releasel.

And also correct-  upgrading the API on the site will make no difference, because Stripe.net will just force it to go down to a previous version.

benwmills:
Do you think there might be a potential issue with the web hooks?  I still have version 2011-11-17 set in Stripe.  I wonder if it passes different information in the web hook call than it would if I updated the API version in Stripe.  It's not like regular API calls where you can specify the exact version of the API.

In my case, I'm looking at an example of the invoice.payment_succeeded web hook that was sent to my web site.  The JSON sent in the call seems to have a structure which might not match Stripe.net 2.3.4 (see #249).  Will the Stripe.Mapper.MapFromJson call work?  Will it just map what it can match?  If I can at least get the Invoice ID, then I can make my own call to get the invoice with all the fields in the new version.

Issue #114, opened 2013-11-21
Strongly Named Assembly


rbowness:
Would it be possible to for the stripe.net assembly to be signed with a strong name key for use in signed projects?

PR #183, opened 2014-06-26
Enabled signing of the Stripe.net DLLs

sidshetye:
This signs the Stripe.NET binaries and resolves ticket #114

Please note that we don't have the commercial testing framework used (NCrunch) so can't run the tests.


jaymedavis:
Lately i have discussed creating a monetized (improved functionality) version of stripe.net and have discussed protecting the dll. I hear rumors the signing of dll's is going away or things like that. Are you familiar with this? 

Not to hi-jack your original thread, but I was hoping a strong name signing could be a mechanism for protecting a product key or something similar in the library - but I don't want to waste time doing this if it's going away. 

Are you still in need for this? Is it a company policty for you to use only signed dll's?

sidshetye:
Hi Jayme,

Signing won't prevent disassembly - it's kinda like a header that only the holder of the private key can generate. Anyone can strip the original one (whose public key may be advertised on the product website) with their own key or just totally strip it off. So it's more for authenticity (if found) than privacy.

Having said that, we recently switched from strong named dlls to regular ones because some private dlls from MSFT were being sent to us without strong name signatures (and some heavy resistance from them on being signed). Not sure if that is indicative of the fact that they are being retired but it swayed our decision.

I still think you should be having strong names on your dll because without it you're forcing your users to disable strong names on their projects (and I think final users/developers should have that freedom of "choice")

Regards
Sid

jaymedavis:
So it seems like - while it's not secure, I could generate a private key (theoretically) for every single user of stripe.net (the enhanced edition). While it doesn't offer security from distributing it to other users, I could at least name the private key your company name and sign it with that private key instead of a just a single key I have for myself.

Is that crazy?

Thanks for the help, Sid!

sidshetye:
Honestly, that wouldn’t be my first approach. Took some time to write this in my free time but I like your library so like helping.

 

Seems your objective is to distribute a premium Strip.NET library which works only when some check passes that it’s “paid for”.

 

So zooming back, a better approach would be to leave the .NET strong signing mechanisms aside since they provide no security assurances other than proving origin of the binary. Having seen a few solutions, here is what I suggest:

 

Server side registration

1.       Create an elliptic curve certificate public-private key pair (eg. via OpenSSL)
openssl ecparam -name prime256v1 -genkey -param_enc explicit -out private-key.pem

openssl req -new -x509 -key private-key.pem -out server.pem -days 730

2.       User registers and pays you

3.       Upon confirmed payment, use ECDSA to sign a “user identifier” (say username@server.com) creating a signature ( <http://msdn.microsoft.com/en-us/library/system.security.cryptography.ecdsacng.signdata%28v=vs.110%29.aspx> MSDN link) . A 256 bit EC key should give you 64 bytes of signature (say WhKI0ojbph1JS24bhltGUub2KHwiWiI2T+oyzc/HxojbJ3w/HklagP8pKZstrX91OBh/ENVfnDOZxnecSBNM6g== in base64)

a.       Suggest having a very public but terse “user string” like email. 

4.       Email the result back to user at their email address. Email the full sample usage to minimize user error (see clientside-1-a)

 

Client side

1.       User puts the received string in their app.config file (similar to how they currently add the Stripe API key). 

a.       For example <add key="StripeNetRegistration" value=" username@server.com:WhKI0ojbph1JS24bhltGUub2KHwiWiI2T+oyzc/HxojbJ3w/HklagP8pKZstrX91OBh/ENVfnDOZxnecSBNM6g==" />

2.       When the library comes up check is StripeNetRegistration exists. 

a.       If no => community/free features => done with this part

b.      If exists, check signature in step 4

3.       Signature checks:

a.       Add the certificate (and therefore the public key) as an embedded resource to your DLL

b.      Load this certificate/resource at runtime from the DLL => Load to EcDsaCng class

c.       Use VerifyData method to confirm the signature ( <http://msdn.microsoft.com/en-us/library/system.security.cryptography.ecdsacng.verifydata%28v=vs.110%29.aspx> MSDN link)

 

 

Notes:

a)      Using EC @ 256 bits since the resulting signature is much smaller than RSA 2048 (despite equivalent strength of at RSA 16K). See NSA Suite B. This helps users in copy-pasting the signature and in visual tidiness.

b)      Ignore the digital crackers. If someone *really* wants to hack your DLL, then can always replace your genuine certs with their fake ones. Or just invert the if condition in clientside-2-a above. Or a ton of other things since the DLL is 100% in their control after delivery/download. Without server side checks that you perform, there is little you can do when the code and execution platform is out of your control. I’d suggest ignoring such people, they wouldn’t be your “customers” anyway and you’re wasting your time on an effort that won’t bring any revenue.

c)       You could actually simplify the above if you went with HMAC-SHA256 (MSDN link <http://msdn.microsoft.com/en-us/library/system.security.cryptography.hmacsha256%28v=vs.110%29.aspx> ) rather than public elliptic curve (EC) crypto. E.g. signature=HMAC-SHA256(secret-key, user-email). The issue is this facilitates key-generators (vs crackers) because the ‘secret-key’ is shared and lives in both the server signature generator as well as the client side (for verification). Anyone can trivially see the secret-key inside the DLL and then forge a key-gen. EC raises the bar a bit, to b) above

 

Feel free to email me at sid314 at outlook dot com if you have any specific questions on the above.

 

Regards,

Sid

jaymedavis:
Sid, 
I might have more questions when it comes time to make this happen, but I cannot thank you enough for this awesome information, It has been very valuable. Should I use this technique or need more of your help along the way, I would be happy to kick over some stripe.net donation money I've received.

Thanks again!

paulirwin:
Just chiming in on behalf of the #nomorestrongnaming community... Strong names create problems including needing binding redirects when multiple assemblies reference slightly different versions. The community as a whole is definitely moving away from strong-named assemblies (just search twitter for #nomorestrongnaming) and ASP.NET 5 gets rid of it altogether AFAICT.

I'm a member of the Lucene.NET PMC, and they've discussed creating two Nuget packages - one strong named and one not, and the default would be the non-strong-named version. That way, people that really need strong naming can use it, but it doesn't burden people that don't need it.

jaymedavis:
Thanks @paulirwin for chiming in. 

@sidshetye - what is the reason you need this lib signed?

sidshetye:
Hey Jayme,

We originally needed it because we had signed binaries ourselves. But since then we ourselves have switched to unsigned binaries to simplify our processes (for reasons Paul mentioned too)

As such we don't need strong named binaries anymore but that's just us.

Cheers
Sid

jaymedavis:
closing this - thanks for the advice on the security.

PR #187, opened 2014-07-10
Update API to 2014-06-17 and other minor improvements

CodeAnimal:
I have updated the library to work with the Stripe API 2014-06-17. I have also added full refund support, with a refund service and list support. There is now a Refund method on StripeCharge entities also. All changes are backwards compatible and should have tests.

I have also changed some tests so that they get the currency from AppSettings. Also StripeEventListOptions now inherits from StripeListOptions.

No version numbers have been updated except the API version in StripeConfiguration.


src/Stripe/Entities/StripeRefund.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                ExpandableProperty<StripeCharge>.Map(value, s => ChargeId = s, o => Charge = o);
+            }
+        }
+
+        [JsonProperty("metadata")]
+        public Dictionary<string, string> Metadata { get; set; }
+
+        public static implicit operator StripeCharge(StripeRefund r)
+        {
```
jaymedavis: what is this for?

jaymedavis:
So I realize this PR is pretty old, but I'm trying to get the good parts merged in. I left a comment on one section I wasn't sure what it was for - 

To move to the "2014-06-17" version, the stripe docs state: 
*Change refunds property on charge responses from an array to a sublist object, which contains the total_count, has_more, url, and data parameters.*

This code was almost there, but it only included the data. It didn't have total_count, has_more, or url. Also, due to other bugs, the expandable properties have to have the public type marked as JsonIgnore (there was no way you could have known about that one). 

Anyway, I'm about to check in a new PR that pulls a litlte bit more of this PR in to make it compatible to 2014-06-17 with the above changes. I have not merged the test. 

Also, I think I would rather avoid the currency flag. I don't really find the value in testing it for any other types at the expense of having users have to use it in their config for testing. If you test for USD or not, it doesn't really matter - does that make sense? 

Anyway, checking coming soon if you want to see what I had to change from the version in this PR to get StripeRefundList setup the correct way. StripeCharge has changed quite a bit, but that's just over time.

Would love your feedback after I checkin - thanks!

jaymedavis:
I am going to close this since it appears refunds are the only thing left (which still need to be implemented). The library has changed so much since this PR, I'm not sure what kind of a headache it will be to merge in master and submit only the parts for the refund service. 

I'll leave it up to you, but I'll close it and reference it in a new request to be completed. (there are quite a few things - like file uploads - that I have not had time to implement yet)

jaymedavis:
closing - the remainder of this will be address in #299

PR #237, opened 2014-12-04
Async support

paulirwin:
I modified Requestor to have async methods for all of the operations, and added async versions of all of the service methods. I started work on this before the .NET 3.5 support was merged in, so my second commit merges that in and adds "#if NET40" around all the async methods. This uses the Microsoft.Bcl.Async NuGet package so that it works on .NET 4.0 as well as .NET 4.5 and later, while the #if statements allow the 3.5 build to compile.

One thing I noticed is that the .NET 4 and .NET 3.5 builds output to the same folder -- we may want to consider putting them in platform-specific output folders.


jaymedavis:
Readme updates would definitely help. Tests help too. 

One of the things I'm struggling with thinking about on this PR is how worth it keeping 3.5 support is. It seems like a lot of hurdles to keep support for 3.5. Unfortunately, there is no clear way to know how used 3.5 is to my knowledge.

jaymedavis:
If I end up dropping 3.5, I don't expect you to redo all of your work on this pull request, though. Thanks again for your help. Just sharing what's holding me up.

paulirwin:
I'm fine either way, although maintaining 3.5 support was not that big of a deal.

I'm a little tied up today and this weekend but I'll work on README updates next week. Any interest in creating a Wiki here on GitHub for documentation?

rvoyentzie:
Hi Jayme,
Do you plan on supporting ACH charges?  Right now it's in beta only.
I'm sure you already have this info, but here's a link to the documentation.https://gist.github.com/raycmorgan/166e899cbb359768ad29
Great job on the code by the way.  I'm new to stripe and just getting started, but if we end up using the service I'll be donating for sure.  Thanks for all the work you have done to make Stripe integration easy.
-Rich
Date: Thu, 8 Jan 2015 19:03:55 -0800

jaymedavis:
Maintaining it for more experienced developers is easy, but I worry for newcomers that having two projects (Stripe and Stripe35) is confusing (people have asked me if they need to make changes to both places when it's actually just symlinks). Perhaps the easy thing to do here is just remove the project from the solution if it still compiles fine. My other way is bloating the codebase with #if NET_40 statements for what might be 5 people using 3.5 (again, I wish nuget had stats on this - I am going to reach out to the folks at NuGet to see if there are any stats here). 

As for the wiki, yes. The readme is getting a bit cumbersome. Some of my plans here are to create a much more usable site for documentation, and perhaps even include XML comments in the codebase for easier consumption. Of course, this is starting to get into more and more time, so I might looking into monetizing a version of the library somehow (maybe with a debug console or somethign). Nothing in stone yet. 

I would also like to have a nice github page for documentation, probably similar to the Stripe documentation for ease of consumption. 

Thanks for your concern/help!

jaymedavis:
Hey @rvoyentzie - thanks for the kind words! 

My goal is typically to support everything once it reaches https://stripe.com/docs/api (out of beta). This looks like a fairly bigger feature, so it may take some time - but it will eventually make it!

rvoyentzie:
awesome, thanks Jayme!  Look forward to checking it out.  I'm sure I'll be in touch...
-Rich

Date: Fri, 9 Jan 2015 09:20:51 -0800

paulirwin:
Jayme, I've added async documentation to the Readme.md file. Let me know what you think.

paulirwin:
Jayme, I have merged the latest changes from master into this pull request. Please let me know if you need anything else.

jaymedavis:
Awesome Paul, thanks so much.

The only reason this is in yet is because I am so far behind with verifying my entities/services match Stripe's. If you look at StripeConfiguration, you will see I am here "2014-07-22". Look here https://stripe.com/docs/upgrades and you will see I am way behind. :)

I just wanted to give you an explanation as to why so much work got done recently w/out your pull request being in there. But merging in master and adding that to the readme makes my life much easier. I appreciate it.

readme.md
```diff
@@ -771,3 +945,8 @@ Any errors that occur on any of the services will throw a StripeException with t
 
 The StripeException contains and HttpStatusCode and a StripeError entity. The StripeError entity contains the type, message, code and param. For more infomation, review the Errors section
 of stripe here: https://stripe.com/docs/api#errors
+
+Async Support
+-------------
+
+All methods have Async versions that are non-blocking so that you can use them on web applications for higher throughput, or GUI applications to prevent the UI from locking up while waiting on a response. Async versions are not available in the .NET 3.5 build.
```
jaymedavis: the readme looks great - I assume nothing special is required using the await methods? You can just continue coding like you were writing blocking code? 

Sorry, I'm an await noob. :)

src/Stripe.Tests/infrastructure/camel_case_contract_resolver/when_mapping_invoice_item.cs
```diff
@@ -21,7 +21,7 @@ static when_mapping_invoice_item()
         Because of = () =>
         {
             var path = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "../../infrastructure", "test_data", "invoice_item.json");
-            var json = File.ReadAllText(path);
+            var json = File.ReadAllText(path);
             _invoice = Mapper<StripeInvoiceLineItem>.MapFromJson(json);
         };
```
jaymedavis: what's changed here?

src/Stripe.Tests/invoiceitems/invoiceitem_behaviors.cs
```diff
@@ -6,7 +6,7 @@ namespace Stripe.Tests
     [Behaviors]
     public class invoiceitem_behaviors
     {
-        protected static StripeInvoiceItemCreateOptions StripeInvoiceItemCreateOptions;
+        protected static StripeInvoiceItemCreateOptions StripeInvoiceItemCreateOptions;
         protected static StripeInvoiceLineItem StripeInvoiceItem;
         protected static string StripeInvoiceItemId;
```
jaymedavis: looks like no changes here

src/Stripe.Tests/invoiceitems/when_creating_an_invoiceitem.cs
```diff
@@ -4,7 +4,7 @@ namespace Stripe.Tests
 {
     public class when_creating_an_invoiceitem
     {
-        protected static StripeInvoiceItemCreateOptions StripeInvoiceItemCreateOptions;
+        protected static StripeInvoiceItemCreateOptions StripeInvoiceItemCreateOptions;
         protected static StripeInvoiceLineItem StripeInvoiceItem;
         protected static string StripeInvoiceItemId;
```
jaymedavis: looks like no changes here

src/Stripe.Tests/invoiceitems/when_creating_an_invoiceitem_with_an_api_key.cs
```diff
@@ -5,7 +5,7 @@ namespace Stripe.Tests
 {
     public class when_creating_an_invoiceitem_with_an_api_key
     {
-        protected static StripeInvoiceItemCreateOptions StripeInvoiceItemCreateOptions;
+        protected static StripeInvoiceItemCreateOptions StripeInvoiceItemCreateOptions;
         protected static StripeInvoiceLineItem StripeInvoiceItem;
         protected static string StripeInvoiceItemId;
```
jaymedavis: looks like no changes here

src/Stripe.Tests/invoiceitems/when_getting_an_invoiceitem.cs
```diff
@@ -4,7 +4,7 @@ namespace Stripe.Tests
 {
     public class when_getting_an_invoiceitem
     {
-        protected static StripeInvoiceItemCreateOptions StripeInvoiceItemCreateOptions;
+        protected static StripeInvoiceItemCreateOptions StripeInvoiceItemCreateOptions;
         protected static StripeInvoiceLineItem StripeInvoiceItem;
         protected static string StripeInvoiceItemId;
```
jaymedavis: looks like no changes here

src/Stripe.Tests/invoiceitems/when_listing_invoiceitems.cs
```diff
@@ -5,7 +5,7 @@
 namespace Stripe.Tests
 {
     public class when_listing_invoiceitems
-    {
+    {
         private static List<StripeInvoiceLineItem> _stripeInvoiceItemList;
         private static StripeInvoiceItemService _stripeInvoiceItemService;
```
jaymedavis: looks like no changes here

src/Stripe/Entities/StripeAccount.cs
```diff
@@ -29,7 +29,7 @@ public class StripeAccount : StripeObject
         public string DisplayName { get; set; }
 
         [JsonProperty("email")]
-        public string Email { get; set; }
+        public string Email { get; set; }
 
         [JsonProperty("statement_descriptor")]
         public string StatementDescriptor { get; set; }
```
jaymedavis: you may be able to remove your changes here entirely - I recently updated this file with stripe's docs

src/Stripe/Entities/StripeEventData.cs
```diff
@@ -4,18 +4,18 @@ namespace Stripe
...
-		public object Object { get; set; }
+#else
+		[JsonProperty("previous_attributes")]
+		public object PreviousAttributes { get; set; }
+
+		[JsonProperty("object")]
+		public object Object { get; set; }
 #endif
 	}
 }
```
jaymedavis: what's different?

src/Stripe/Entities/StripeInvoiceLineItem.cs
```diff
@@ -1,77 +1,77 @@
...
+        public StripePlan Plan { get; set; }
+
+        [JsonProperty("quantity")]
         public int? Quantity { get; set; }
 
-        [JsonProperty("subscription")]
+        [JsonProperty("subscription")]
         public string SubscriptionId { get; set; }
     }
 }
```
jaymedavis: Another one I recently updated, you may be able to discard your changes

src/Stripe/Entities/StripePageable.cs
```diff
@@ -3,17 +3,17 @@
...
+        [JsonProperty("has_more")]
+        public bool HasMore { get; set; }
+
+        [JsonProperty("url")]
+        public string Url { get; set; }
+
+        [JsonProperty("total_count")]
         public int TotalCount { get; set; }
     }
 }
```
jaymedavis: what's changed?

src/Stripe/Entities/StripeRefund.cs
```diff
@@ -1,63 +1,63 @@
...
+        [JsonProperty("reason")]
+        public string Reason { get; set; }
+
+        [JsonProperty("receipt_number")]
+        public string ReceiptNumber { get; set; }
+
+        [JsonProperty("description")]
+        public string Description { get; set; }
+    }
 }
```
jaymedavis: what's different? you may use tabs, that might be getting all these diffs? :)

src/Stripe/Entities/StripeRefundList.cs
```diff
@@ -1,23 +1,23 @@
...
+        [JsonProperty("has_more")]
+        public bool HasMore { get; set; }
+
+        [JsonProperty("url")]
+        public string Url { get; set; }
+
+        [JsonProperty("total_count")]
+        public int TotalCount { get; set; }
+    }
 }
```
jaymedavis: what's different?

src/Stripe/Stripe3.5.csproj
```diff
@@ -1,178 +1,178 @@
...
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <PropertyGroup>
+    <PostBuildEvent>
+    </PostBuildEvent>
+  </PropertyGroup>
+  <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
+</Project>
```
jaymedavis: this is hard to diff

src/Stripe/Stripe.csproj
```diff
@@ -158,4 +178,16 @@
...
+    <Error Condition="Exists('..\packages\Microsoft.Bcl.Build.1.0.14\tools\Microsoft.Bcl.Build.targets')" Text="The build restored NuGet packages. Build the project again to include these packages in the build. For more information, see http://go.microsoft.com/fwlink/?LinkID=317568." HelpKeyword="BCLBUILD2002" />
+  </Target>
+  <!-- To modify your build process, add your task inside one of the targets below and uncomment it.
+       Other similar extension points exist, see Microsoft.Common.targets.
+  <Target Name="BeforeBuild">
+  </Target>
+  <Target Name="AfterBuild">
+  </Target>
+  -->
+</Project>
```
jaymedavis: I need to learn about the three Bcl libraries included and what each purpose is for. Can you maybe tl;dr that and what this does for hte project for me?

jaymedavis:
I left a few comments here for questions (mostly about bcl) and some cleanup things to do. I like the FormatListUrl(listOptions) stuff you did, but it still seems like there is a way we could reduce repeating ourselves. But I'm not as worried about that right now. There also seems like some ways we could implement more DRY techniques on the last call in the services (since they match identically). I think this PR is an awesome start though and it's definitely a great PR to work from. Good job.

readme.md
paulirwin: Yes, that's the idea. Mark the method with async, return Task or Task<T> (or void for event handlers), and use the await keyword to await the async operation. Overview here: https://msdn.microsoft.com/en-us/library/hh191443.aspx

paulirwin:
Most of the "no changes here" was from the merge, and I believe that has to do with line endings. This is due to this repository not having a .gitattributes file specifying the line endings. The best option will be to do a one-time commit to clean up the line endings with a .gitattributes file. See here: https://help.github.com/articles/dealing-with-line-endings/

I'd be happy to add a line endings commit to this PR if you would like, using the default .gitattributes from GitHub for Windows. That will result in a lot of files that look like they've changed, but they haven't. But otherwise, you can feel free to ignore the changes here for now (again, most were from the merge from master, not from me) and do a gitattributes cleanup later.

src/Stripe/Stripe.csproj
paulirwin: The BCL libraries are required to support async on .NET 4, since the project targets .NET 4 and not .NET 4.5. See here: http://blogs.msdn.com/b/bclteam/archive/2012/10/22/using-async-await-without-net-framework-4-5.aspx

jaymedavis:
Sweet, thanks! I will stash my current work (finally fixing the list methods to return a StripePageable object) and check in a few small fixes I have. This might require one more rebase/merge on your part, but it's all really small stuff if so. But I will go ahead and do a release asap to get this file in. This problem has been bugging me for ages. :)

jaymedavis:
I will just follow the instructions in the link you sent me, I quickly ran over it, but looks easy enough. _famous last words_

jaymedavis:
doesn't look like anything is different - do you need to redo the PR?

src/Stripe/Infrastructure/Requestor.cs
```diff
@@ -1,106 +1,182 @@
...
+using System.Threading.Tasks;
+#endif
+
+namespace Stripe
+{
+    internal static class Requestor
+    {
+        public static string GetString(string url, string apiKey = null)
+        {
+            var wr = GetWebRequest(url, "GET", apiKey);
```
jt000: Hey @paulirwin . Nice change! I was about to do this myself when I came across your pull request. One note that might help out @jaymedavis or any future developers in this code is to call the async call from the sync call to prevent code duplication.

For example:
```c#
return GetStringAsync(url, apiKey).GetAwaiter().GetResult();
```

Note that you can't just call ```task.Result``` or ```task.Wait()``` or exceptions will [bubble up as AggregateException rather than the original exception](http://stackoverflow.com/questions/17284517/is-task-result-the-same-as-getawaiter-getresult).

Thanks for doing this!

src/Stripe/Stripe.csproj
jt000: Should Stripe.net.nuspec be updated to automatically include the reference for BCL on 3.5?

src/Stripe/Infrastructure/Requestor.cs
jaymedavis: The entire thing is disabled in 3.5 anyway. I might drop support for 3.5 altogether, it's a pain to manage.

src/Stripe/Stripe.csproj
jaymedavis: Yes.

jt000: Sorry, I meant add to nuspec for 4.0 (not needed for 4.5 since it was added in that version & not available for 3.5)

src/Stripe/Infrastructure/Requestor.cs
paulirwin: @jt000 Not only does that not work with 3.5, but it is best not to mix async and blocking code, in particular calling Result or GetResult() from synchronous code without paying attention to the captured context can result in a deadlock. More info here: https://msdn.microsoft.com/en-us/magazine/jj991977.aspx

paulirwin:
I have updated this PR again with the latest changes from master.

jaymedavis:
It's not safe to merge yet. BCL does not work with the version of nuget in stripe.net, so it wont compile. Also, there are no dependencies listed in the nuget spec, so it wont download/compile for other users.

I'd like to do some other small cleanup items as well, but I will just move this to another branch and work on it a bit.

jaymedavis:
On second though, probably best to handle the items on this branch to keep it easy. Here are still some things I see:

1) It wont compile, needs the new nuget.exe from https://nuget.codeplex.com/releases/view/58939
2) Is the extra stuff in app.config necessary? I would prefer to just have the key in there
3) I tried to DRY some of the FormatGetUrl/ListUrl stuff (good idea btw), but it turned out to be more complicated than I thought, so I left it for now.
4) The BCL dependencies need to be added to the nuspec file under json.net - and for that, only Microsoft.Bcl.Async needs to be added. It has a dependency on Bcl, which has a dependency on Bcl.Build, so there is no need to specify all 3.
5) Are all of the added assemblies necessary in Stripe.csproj? Or are maybe just one or two required? I like to keep the library as small as possible. None of them are needed in Stripe.Tests due to no tests.

Thanks

paulirwin:
I think creating a branch for this would be a good idea. Then I can submit minor PRs separately against that branch instead of master, until this is fully complete. 

Regarding #2, the app.config changes were done by the BCL NuGet packages, and since the app.config is not deployed with the Stripe.net NuGet package, this doesn't matter. I'd leave it so that the BCL NuGet doesn't keep changing it later.

Regarding #4, yes, the BCL Async package needs to be added as a dependency to the nuspec file. I did not include this because that is more of a concern at release time.

Regarding #5, anyone using the Stripe.net NuGet package will have to install them anyways, so it doesn't matter if they're referenced here since it will be a nuspec dependency. For the test project, yes, the references are needed because the test project wasn't building without them for me. If you can get the tests project to build without the references, that's fine. But again, there's no harm here since the libraries are a dependency (might be better for the test project to reference all dependencies actually, IMHO).

jaymedavis:
I added a list of some things that needed to be fixed above. I assume (thought i'm not an expert with async), there are ways individual implementations could add async functionality on their own without it being part of this library. That's not to say this wont be merged in, but it hasn't been priority. It's been a struggle just to keep up with stripe's changes (which i'm still behind).

I disagree with Paul's notes above about things 'not mattering' in the codebase. He is right in the fact that it technically wont break anything, but it is still bloat that is not needed for the library to function, and IMO that is harmful to any codebase. If a line of code is not needed (especially in user facing files like app.config), it does not belong. 

I'm not against supporting async, I just haven't personally had the time to go through this and fix it up and test it.

jamesthurley:
+1 for this.  I don't think there is an easy way to implement async functionality efficiently outside the library itself. You could create async wrappers to the synchronous calls, but underneath there will still be a blocked thread being tied up by the synchronous call in the library.

b9chris:
Yes, agree with James; it's common to assume you can just wrap sync calls
in async wrappers and get just as good a performance benefit, but this is
very much not the case. All you do is peg the CPU on some other thread.
Really if you wanted this kind of hands off, low bloat ethos, the right
implementation today is async everywhere. Then those who wanted sync could
indeed wrap without losing anything they'd have gotten with a pure sync
limitation.

jbarker4682:
+1 for this... I'm a little rusty on my .Net, but what about using extension methods to add asynch functionality?

paulirwin:
@jbarker4682 See @b9chris' comment. It is best to go async all the way down rather than "fake" async by wrapping it somehow. The only real way to wrap sync with async is to do Task.Run or similar, but that is only beneficial for CPU-bound tasks, not for IO-bound tasks like HTTP calls in this case.

paulirwin:
See this post: http://blog.stephencleary.com/2013/10/taskrun-etiquette-and-proper-usage.html

> One common mistake is to try to make asynchronous “wrappers” around existing synchronous methods.
> ... I call such methods “fake-asynchronous methods” because they look asynchronous but are really just faking it by doing synchronous work on a background thread.
> ... If your code is ever used on ASP.NET, a fake-asynchronous method leads developers down the wrong path. The goal of `async` on the server side is scalability, and fake-asynchronous methods are less scalable than just using synchronous methods.

jbarker4682:
@paulirwin Hey, yeah... I actually agree with you. What I meant to say is that we could add asynchronous versions of each method (all the way down) via extensions, if we wanted to keep the 3.5, 4.0 and 4.5 code independent from each other, rather than using conditional compilation or the BCL

paulirwin:
@jbarker4682 Gotcha. But I'm not entirely sure what the benefit would be of using extension methods within this library rather than partial classes, for example. (Note that I am no longer advocating adopting my original code in this PR. This PR is now very out-of-date but could be updated with a new approach or started from scratch.)

b9chris:
Yeah I actually tried merging the PR and gave up. Wrote a narrow lib of my own that's async-only. With async avail since 2012 and our starting a new project, it didn't make any sense to peg the CPU during net calls. Don't know if that's any incentive to get async going here... . But there ya go.

jbarker4682:
@paulirwin: I only suggested it because I thought a main concern was keeping support for .NET 3.5 and/or VS 2012. If true, I believe the async methods could be added as part of an extension library, so everyone still using 3.5 / VS 2012 can just use the existing code, while anyone on 4.0+ could download the extensions.  Personally, I would drop the 3.5 support all together, but that's just me...

jbarker4682:
@b9chris I'm actually doing the same thing... I created a fork and I'm adding async methods, as well as other general refactoring / code-cleanup.  Also going to see if I can add in the Connect / Managed Accounts functionality as well...

jaymedavis:
Thanks for the link @paulirwin. The biggest hurdle/smell to this has been 3.5 support. I want to drop it worse than anyone else, lol. I will be looking into dropping it, but it wouldn't be here if there wasn't a lot of request for it.

As for managed accounts, you can search other forks and fine them complete in quite a few places. Forks spin up in a ton of places, but that work doesn't necessarily make it back as a PR.

I think the best plan for async is ditching 3.5. The hard part of that is knowing what the demand of 3.5 is - perhaps it has fallen off. But I do agree it's a PITA and makes a lot of things more difficult.

b9chris:
@jbarker4682 Interesting. I'll open source what I've put together once I've got our private company libs abstracted out. We can compare notes!

adaam2:
Is there any update on async support?

jaymedavis:
No, this PR has been pretty quiet relative to other things. Thanks for letting me know you're interested in this.

jaymedavis:
.NET 3.5 support was removed, which makes this somewhat easier, but many other platforms are now supported which makes it somewhat more complicated. With the recent addition of .NET Core, Xamarin (iOS, Android), a PCL, etc.. most of which do not support async. We actually tried when we were working on the PCL, but Xamarin just hung. The 4.5 build should be able to do this though, preferably without a bunch of #if's.

adaam2:
@jaymedavis That's no problem. For me at least, it would be a "nice to have" as opposed to a critical feature. Great project btw!

jaymedavis:
@adaam2 Thanks!

I think a lot of the existing users from this PR went with their own forks, so I assumed it wasn't as big of a deal. I can start getting something working though if you would like to have it. I might have to suck it up with the #if statements. I'll research this a bit.

b9chris:
Yeah we ended up borrowing a very small amount of XamarinStripe for our own
Async library for Stripe. I'll try to clean it up and post it if anyone
here is interested.

It has way less functionality than Stripe.Net though. The performance
impact of long-waiting sync operations was just too much, so that portion
of our lib is written from the ground up entirely in .Net 4.0 async, to
ensure there's no fake-async creeping in.

b9chris:
Here ya go

https://github.com/b9chris/StripeAsync

jaymedavis:
Appreciate it, @b9chris. 

Check out the `async` branch, where I have committed the first part of getting this done for the AccountService, here is the diff:
https://github.com/jaymedavis/stripe.net/commit/e6559b563a9a407a611715b10726fb7e177c5e73

The things I'd love some feedback on:
1) The Requestor now supports Async - this was pretty easy after BCL was introduced with the PCL version of Stripe.net recently. It was a little different than @paulirwin's original PR, just because it uses HttpClient now and modernhttpclient to support handlers for multiple platforms (originally it was WebRequest). It's basically the same though. :smile: 

2) The AccountService has been updated to support Async for everything. I was able to make the code look a lot cleaner (imo) by shrinking the code down. This will make it look less copy/pasta, imo.

3) One thing I'm not sure about is how I handled the exceptions, which you can view on the Requestor. If the call was successful, it's async all the way down. If there was an exception, I immediately return the result. There might be a better way to handle this, but since it needs to populate `StripeException`, I'm not sure exactly how I should handle this.

4) I wrote a couple of tests to verify it works, but I obviously need more to test Update/Delete too.

I would really appreciate feedback on the architecture before I start grinding through all of the services updating them to be async (especially on the exception part).

Thanks again everyone who has worked on this so far and for all of the code, especially @paulirwin. Many apologies for the 1.5 year wait to get this in the library. I owe you all a beer.

jaymedavis:
closing this PR - this is being worked on. The branch is async (I think I have a way to handle the exceptions being fully async as well, so it's end-to-end async) if you want to review it. I will be tracking this work on #467, so if you have any feedback or comments, please leave them there. Also feel free to submit PR's to the async branch. A lot of the work is going to be grinding through the services and doing the same thing I did for the AccountService.

jaymedavis:
FYI - this just went live. :+1:

PR #253, opened 2015-01-16
Fixed bug in StripeCustomerService Get with missing url re-assignment

kspearrin:
Simple fix. I noticed this when I wasn't getting the expanded `StripeDefaultCard` back with my customer `Get`.


jaymedavis:
oh snap, nice catch.

PR #262, opened 2015-02-01
Const string of all Event Types

jt000:
Adding constant strings defining all possible stripe event types as defined at https://stripe.com/docs/api#event_object. I found this incredibly helpful so I don't have to keep looking them up online & thought you might want to add it to Stripe.Net...

PR #263, opened 2015-02-01
Adding Verified property to Recipient

jt000:
"Verified" is not explicitly mentioned in [API docs](https://stripe.com/docs/api/curl#recipient_object), but appears in examples and mentioned in [testing recipient](https://stripe.com/docs/testing). It is necessary if you only want to accept recipients who have been underwritten (suggested to help prevent fraud).

*Note: This PR also includes changes in #262. Please look at that PR first, then you can ignore the duplicates in this PR (or if you'd prefer to do them together, then I can close #262 & we can just use this one).*


jt000:
I'm going to close & re-open this without the strings (I have another branch w/ only this change)

PR #264, opened 2015-02-03
Invoice expandable fix

jslaybaugh:
Alright, for what its worth, this definitely seems to be better on the final diff here. I do get a lot of noise in my terminal when I `git add` or `git commit` though: 
screen shot 2015-02-02 at 7 47 52 pm

Also, Since I use tabs instead of spaces, I have to manually space it over. Strangely even if I copy from one of your lines that clearly has spaces, when I paste, VS turns them into tabs right away.

Either way, should be good now. If I have anything else, I'll try to keep these things in mind or if you have any advice (except for trying to convert me to spaces -- you'll have to pry my tabs from my cold dead hands!!!), I'd love to hear it.


jaymedavis:
awesome, thanks dude!

Issue #182, opened 2014-06-24
Recipient Creation with Token procured through Stripe.js


adamjzeuske:
Wonderful library!

I want to be able to use Stripe.js to get a bank account token, and then use it to create a recipient.  The following link suggests it should be possible to have the bank_account be a token (string) instead of the child collection of three account fields.  This would allow me to not post bank account information to my server.

https://stripe.com/docs/tutorials/sending-transfers

In looking at the code, I was thinking a new base class BankAccountToken and then perhaps a RecipentTokenCreateOptions, with the service knowing to just serialize the token instead of the child object, as seen in the API.

What are your thoughts?

Respectfully,
Adam

jaymedavis:
Sorry this has been lingering for so long - 

It looks like with all the changes now, the class needs a StripeBankAccount and StripeCard or something similar. I think the BankAccountOptions might not work here anymore, or there is definitely a way to clean it up.

jaymedavis:
This has been fixed in 2.7.0. When creating/updating a recipient, both options are now supported (card and bank account).

PR #265, opened 2015-02-03
Adding ability to set Bank Account token from Stripe.js

jt000:
This change includes the following modifications
* Updated the parameter builder to allow traversing of child objects (so you don't have to use inheritance to add re-usable child properties)
* Added a QueryStringParameterConverter to allow custom conversion of types to query string value (required for BankAccountToken, so it doesn't show as a sub-value)
* Added interface IBankAccountOption to allow both the original BankAccountOption (with account, routing, country) and BankAccountToken (with token from stripe.js).

*Note: Should not include any breaking changes since the original BankAccount values are piped to the new BankAccountOption class with an ObsoleteAttribute to push people to switch*

This provides a fix for Issue #182

### DIFF
--- a/src/Stripe.Tests/infrastructure/test_data/sample_object.cs
+++ b/src/Stripe.Tests/infrastructure/test_data/sample_object.cs
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 using System.Text;
 using Newtonsoft.Json;
+using Stripe.Infrastructure;
 
 namespace Stripe.Tests.test_data
 {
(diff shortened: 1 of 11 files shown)


jaymedavis:
would you mind merging master then resubmitting this pull request? I've added .gitattributes which should help with the whitespace/line endings, but as it stands I cannot tell what is different in parameterbuilder, etc

jt000:
Resubmitted in PR #274.

I didn't see anything specifically with whitespace\line endings. There are a lot of changes in the creation of the QueryString (pulled it out into its own method, so I could better see what it was doing). Maybe that is what you're seeing?

The QueryString code could be further cleaned up, but didn't want to introduce unnecessary changes.

jaymedavis:
oh maybe it was - my mistake. thanks so much, Jason!

PR #268, opened 2015-02-05
Stripe Recipient Create Options Update to Include Debit Card. Stripe Transfer Options Update to Include Recipient Card. Stripe Card Service Update to include Recipient URL.

Giovani-Brady:


jaymedavis:
can you summarize what all is going on here?

a couple of things off hand:

1) I'm not a big fan of the properties, I'd rather just not repeat typing the same thing if that's possible.
2) The interface option just means you would have to implement that in a lot of different places, probably why you did the number 1. There are lots of classes that would need this interface.

Giovani-Brady:
Updated implementation.

Stripe API Recipient Services supports the use of a debit card. (https://stripe.com/docs/api#create_recipient) Updated StripeRecipientCreateOptions and StripeRecipientUpdateOptions to include card properties. (https://stripe.com/docs/api#create_card) Stripe API Card Services support CRUD opertions of Customer credit cards and Recipient Debit Cards. Updates to StripeCardService to support these operations. Stripe API Transfer Services support transfers to recipient bank accounts and debit cards. (https://stripe.com/docs/api#create_transfer) Updated StripeTransferCreateOptions to include card property.

jaymedavis:
I just pushed a big update (2.7.0) that addresses the majority of this. Let me know if anything is missing!

Bank accounts and cards are now supported together. Thanks for your initial pull request, but I had something in flight already to solve the problem. :)

Giovani-Brady:
2.7.0 updates are missing StripeCardService Recipient debit card create, retrieve, and update API endpoints. This commit includes those changes.

jaymedavis:
Can you be more specific?

Giovani-Brady:
ending thread.

PR #271, opened 2015-02-09
Adding verified to StripeRecipient

jt000:
Note: "Verified" is not explicitly mentioned in https://stripe.com/docs/api/curl#recipient_object, but appears in examples and mentioned in https://stripe.com/docs/testing.

(re-opened from #263 to remove the event strings)

### DIFF
--- a/src/Stripe/Entities/StripeRecipient.cs
+++ b/src/Stripe/Entities/StripeRecipient.cs
@@ -38,6 +38,9 @@ namespace Stripe
         [JsonProperty("cards")]
         public StripeCardList StripeCardList { get; set; }
 
+        [JsonProperty("verified")]
+        public bool Verified { get; set; }
+
         public string StripeDefaultCardId { get; set; }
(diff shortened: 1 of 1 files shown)


jaymedavis:
this could mean stripe is working on removing it. there are a few places they have removed verified, disabled, stuff like that and a 'status' field has shown up. Check with them to make sure it's just missing from the docs, otherwise i'll just end up having to remove it soon. :)

jaymedavis:
crap, I was going to just slip this in to 2.7.0. :(

Is this urgent for you? I can make a 2.7.1 if you really need it.

jaymedavis:
had to patch, snuck it in. :)

jt000:
Thanks Jayme. I verified that the property isn't scheduled for removal. It was actually recently added & hadn't been documented yet.

From mail with support@stripe.com
>  The recipient object is still pretty new and not all properties are explicitly documented just yet but you should still be able to use it as expected.

jaymedavis:
perfect! thanks!

PR #274, opened 2015-02-14
Adding ability to set Bank Account token from Stripe.js

jt000:
This change includes the following modifications
* Updated the parameter builder to allow traversing of child objects (so you don't have to use inheritance to add re-usable child properties)
* Added a QueryStringParameterConverter to allow custom conversion of types to query string value (required for BankAccountToken, so it doesn't show as a sub-value)
* Added interface IBankAccountOption to allow both the original BankAccountOption (with account, routing, country) and BankAccountToken (with token from stripe.js).

*Note: Should not include any breaking changes since the original BankAccount values are piped to the new BankAccountOption class with an ObsoleteAttribute to push people to switch*

This provides a fix for Issue #182 
Resubmitting PR #265


jaymedavis:
I ended up going a different route with this. I didn't want to make a lot of changes to ParameterBuilder, but that's mostly because I'm not really that thrilled with it's implementation to begin with. I hope to break that class up a lot in the future. 

Here are the changes I ended up making to this particular class, tried to keep it simple:
https://github.com/jaymedavis/stripe.net/commit/5be8cd5b80be3d6caad79275397aa495242dbf41#diff-6

To get this entire feature in, there were a lot of other changes that needed to be made (mostly around testing). None the less, you can now use bank accounts and credit cards wherever you need to. :)

Thanks so much for the pull request, I just had something in flight already to solve this problem. You can review the entire diff here: https://github.com/jaymedavis/stripe.net/commit/5be8cd5b80be3d6caad79275397aa495242dbf41. 

Version 2.7.0 is live now.

jt000:
No problem. Your fix looks good, too. I'll merge it into my code & try it out today or tomorrow. Thanks!

PR #279, opened 2015-02-17
Stripe Card Service API Endpoint Updates to support CRUD operations of Recipient Debit Cards

Giovani-Brady:
Makes Stripe API Endpoints available for recipient card creation, retrieval, listing, updating, and deleting.

### DIFF
--- a/src/Stripe/Infrastructure/Urls.cs
+++ b/src/Stripe/Infrastructure/Urls.cs
@@ -52,9 +52,14 @@
             get { return BaseUrl + "/customers"; }
         }
 
-        public static string Cards
+        public static string CustomerCards
         {
             get { return BaseUrl + "/customers/{0}/cards"; }
+        }
+
+        public static string RecipientCards
+        {
+            get { return BaseUrl + "recipients/{0}/cards"; }
         }
 
         public static string Events
--- a/src/Stripe/Infrastructure/Urls.cs
+++ b/src/Stripe/Infrastructure/Urls.cs
@@ -59,7 +59,7 @@
 
         public static string RecipientCards
         {
-            get { return BaseUrl + "recipients/{0}/cards"; }
+            get { return BaseUrl + "/recipients/{0}/cards"; }
         }
 
         public static string Events
--- a/src/Stripe/Services/Cards/StripeCardService.cs
+++ b/src/Stripe/Services/Cards/StripeCardService.cs
@@ -9,9 +9,9 @@ namespace Stripe
         public bool ExpandCustomer { get; set; }
         public bool ExpandRecipient { get; set; }
 
-        public virtual StripeCard Create(string customerId, StripeCardCreateOptions createOptions)
+        public virtual StripeCard CreateCustomerCard(string customerId, StripeCardCreateOptions createOptions)
         {
-            var url = string.Format(Urls.Cards, customerId);
+            var url = string.Format(Urls.CustomerCards, customerId);
             url = this.ApplyAllParameters(createOptions, url, false);
 
             var response = Requestor.PostString(url, ApiKey);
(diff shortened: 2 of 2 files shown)


jaymedavis:
Recipients now support creating and updating bank accounts and cards. Can you point me to the documentation on stripe's site so I can learn more about this?

Giovani-Brady:
https://stripe.com/docs/api#create_card observe on the right hand side just above the definition there is a toggle between "Customer" and "Recipient"

jaymedavis:
AH. I never caught that. :)

Does this pull request address everything for this? Can the other issues be closed or are they unrelated?

Giovani-Brady:
The two pull requests are directly related but can be integrated separately. this pull requests is for debit card management, the other is for transfers into debit cards. these two updates to 2.7.0 make it a much more complete and whole package. now we can brag that we fully support all basic forms of e-payment.

jaymedavis:
haha. Thanks!

jaymedavis:
what about #280 ? is that addressed in one of the two pull requests?

jaymedavis:
Thanks, I think I finally get it. :+1: 

Btw, I lol'd at your spelling of recipient in the code as a parameter. I spell that word right 1 out of 20 times. :)

Giovani-Brady:
yeah, I've got a horrible spelling problem lol

Giovani-Brady:
Yes Issue #280 is addressed in Pull Request #281

Giovani-Brady:
Note: there was a bug, missed the "/" at the beginning of the Recipients end point API. updated and fixed.

jaymedavis:
it looks like this is really the exact same code, just hits a different endpoint. I might try to find a way to do it a little easier, but starting from your code. (something like having the methods share a single method and passing the either recipient or customer). It seems like some duplication could be fixed. But that's really small, this is plenty to get me started. Thanks again.

jaymedavis:
Let me know if anything else is missing in the library for your use! I'd love to hear about what you are building too sometime if you wouldn't mind sharing. jaymedavis@gmail.com if you want to share. :)

Giovani-Brady:
Updated the implementation to be backwards compatible. Let me know what you think.

jaymedavis:
I don't have a lot of time during the workday to look into this, but from a quick glance it looks awesome. :)

jaymedavis:
Is this the last thing missing in the library for it to be full-purpose for you?

Giovani-Brady:
I believe so. It's fine, I'm using a build from a fork for now. I'd rather the code go through your usual process then be expedited with bugs and errors.

jaymedavis:
Thanks man, that gave me a big smile. :)

jaymedavis:
Couple things here I changed - wanted to check with you to make sure it should be fine:

1) CustomerCards is no longer trailing with /cards, looks like they changed it to sources

        public static string CustomerCards
        {
            get { return BaseUrl + "/customers/{0}/sources"; }
        }

2) I changed stripeId to customerOrRecipientId - I thought it was just a little more intuitive.

jaymedavis:
On second thought, I'll probably just stick with /cards. If it works, it works. It would be ideal to talk with stripe about this though. 

Also I changed IsNullOrWhitespace to IsNullOrEmpty, since that method didn't exist.

Giovani-Brady:
IsNullOrWhitespace was introduced as part of the 4.0 .Net framework so the Stripe3.5 project may not compile correctly but the Stripe project targeting the latest framework will compile correctly.

Recommend we use both where appropriate as IsNullOrWhitespace is the replacement for IsNullOrEmpty.

Yes it seems the API endpoint has been updated from "cards" to "sources" with the latest version of the API. It's been updated everywhere but the Retrieve A Card API endpoint has "cards" instead of the new "sources". It would be worth a quick email to verify that the change is global and also shed some light on backwards compatibility "cards" vs "sources"

Also I thought this package had a hard coded version of the API in the project? has this been updated?

jaymedavis:
It still has a hardcoded version of the API in StripeConfiguration - so as long as I don't update that, we should be ok for now with cards. I will release this before I release those updates anyway.

I think i'll stick with IsNullOrEmpty for now. I agree the other is preferable (for the obvious reason it catches whitespace), but the only way around that right now is to to some #if statements in the code, and that makes it a mess. The way the 3.5 project is setup right now, there is no way to have a separate file for each version, it's just #if blocks.

jaymedavis:
Ok, this is all checked in locally and will be part of the next release. I updated the readme as well, to let users know that isRecipient is optional now too. Thanks for the help!

Giovani-Brady:
No problem, let me know if there's anything else hanging to the way side. I'd like to be a regular contributor.

PR #288, opened 2015-02-20
Const string of all Event Types

jt000:
Re-opened from #262.

On a side-note, I think we should do this for all Stripe strings to help with version compatibility (or better yet, convert them to enums). Better to have a build break, than to have a runtime error...


jt000:
Removed the comments from consts as requested in other PR

jaymedavis:
I am just going to copy this file manually and add it for a release - the 16 commits makes the commit log look pretty wild. :)

I think the only thing missing from this is updating the readme under the events section, it still has a hardcoded string. I can take care of that as well.

jaymedavis:
I've got all that done. The only change I made was I moved the class from Entities to a new folder called Types. There is going to be quite a few of these going forward, and I think it made more sense to separate them into another folder (everything still has the same root namespace). This will make it out in the next release.

PR #289, opened 2015-02-23
Adding const strings for comparison with StripeTransfer.FailureCode

jt000:
Using const strings allows for better backcompat (build-breaks instead of functional breaks) and helps avoid simple mistakes like typos.


jaymedavis:
Same as the last one with this one, I'm moving it to the Types folder. Ideally, I'd prefer to have an example of how this is used in event_hooks_handler.cs and/or the readme. This will make it out in the next release. If you have some sample code on how to use it, feel free to just drop it in a comment and I can add it.

PR #290, opened 2015-02-24
Adding Transfer Reversals list

jt000:
The [transfer_object](https://stripe.com/docs/api#transfer_object) contains a list of reversals along with reversed and amount_reversed properties. Adding these to the Stripe.NET api


jt000:
Accidentally merged other changes into this branch... Need to fix (temporarily closing PR)

jt000:
Resolves #313

src/Stripe.Tests/transfers/test_data/stripe_transfer_create_options.cs
```diff
@@ -5,13 +5,13 @@ namespace Stripe.Tests.test_data
 {
     public class stripe_transfer_create_options
     {
-        public static StripeTransferCreateOptions Valid()
+        public static StripeTransferCreateOptions Valid(string recipient = "self")
```
jaymedavis: What are the other options here? This should probably be a static class in the Types folder to follow the convention.

jaymedavis:
This will need to be rebased and readme changed updated. I can handle this in the next release if you don't have time.

jaymedavis:
few small changes:
1) renamed Transfer to TransferId (so it matches BalanceTransactionId). both of these will need to be expandable properties at some point, but there's no reason that has to block this PR.
2) moved the new tests from when_listing_transfers to when_creating_a_failed_transfer
3) instead of the amount 50 being hardcoded in the test, I made it *test_data.stripe_transfer_create_options.Valid(_recipientId).Amount*

Ignore the AppVeyor message above, that was something I was trying but decided to punt on for now. Since the test are not mocked and actually hit stripe, I would have to publish my secret test key in order to get the tests to all pass. That doesn't seem like a good idea.
 
Anyway, this is done and it will be in the next release. I'm going to try and get more stuff completed before I push it up, I'm a bit behind.

Thanks for the help!

PR #291, opened 2015-02-24
Added TaxPercent on CustomerCreate when a Plan is passed.

pedropaf:
Added Billing_Cycle_Anchor.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -204,7 +204,8 @@ a credit card or token, and various meta data.
 	}
 
 	myCustomer.PlanId = *planId*;                          // only if you have a plan
-	myCustomer.Coupon = *couponId*;                        // only if you have a coupon
+	myCustomer.TaxPercent = 20;                            // only if you are passing a plan, this tax percent will be added to the price.
+    myCustomer.Coupon = *couponId*;                        // only if you have a coupon
 	myCustomer.TrialEnd = DateTime.UtcNow.AddMonths(1);    // when the customers trial ends (overrides the plan if applicable)
 	myCustomer.Quantity = 1;                               // optional, defaults to 1
 
--- a/readme.md
+++ b/readme.md
@@ -138,22 +138,22 @@ customer or a charge, but only used once.
 	myToken.Card = new StripeCreditCardOptions()
 	{
 		// set this property if using a token
-		TokenId = *tokenId*;
+		TokenId = *tokenId*,
 
 		// set these properties if passing full card details (do not
 		// set these properties if you set TokenId)
-		CardNumber = "4242424242424242";
-		CardExpirationYear = "2022";
-		CardExpirationMonth = "10";
-		CardAddressCountry = "US";                // optional
-		CardAddressLine1 = "24 Beef Flank St";    // optional
-		CardAddressLine2 = "Apt 24";              // optional
-		CardAddressCity = "Biggie Smalls";        // optional
-		CardAddressState = "NC";                  // optional
-		CardAddressZip = "27617";                 // optional
-		CardName = "Joe Meatballs";               // optional
-		CardCvc = "1223";                         // optional
-	}
+		CardNumber = "4242424242424242",
+		CardExpirationYear = "2022",
+		CardExpirationMonth = "10",
+		CardAddressCountry = "US",                // optional
+		CardAddressLine1 = "24 Beef Flank St",    // optional
+		CardAddressLine2 = "Apt 24",              // optional
+		CardAddressCity = "Biggie Smalls",        // optional
+		CardAddressState = "NC",                  // optional
+		CardAddressZip = "27617",                 // optional
+		CardName = "Joe Meatballs",               // optional
+		CardCvc = "1223",                         // optional
+	};
 
 	// set this property if using a customer (stripe connect only)
 	myToken.CustomerId = *customerId*;
--- a/src/Stripe/Services/Customers/StripeCustomerCreateOptions.cs
+++ b/src/Stripe/Services/Customers/StripeCustomerCreateOptions.cs
@@ -28,6 +28,9 @@ namespace Stripe
         [JsonProperty("plan")]
         public string PlanId { get; set; }
 
+        [JsonProperty("tax_percent")]
+        public decimal? TaxPercent { get; set; }
+
         [JsonProperty("quantity")]
         public int? Quantity { get; set; }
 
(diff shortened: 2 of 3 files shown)


src/Stripe/Services/Customers/StripeCustomerCreateOptions.cs
```diff
@@ -28,6 +28,9 @@ public class StripeCustomerCreateOptions
         [JsonProperty("plan")]
         public string PlanId { get; set; }
 
+        [JsonProperty("tax_percent")]
+        public decimal? TaxPercent { get; set; }
```
jt000: I don't see this property in the [Stripe API Reference](https://stripe.com/docs/api#create_customer). Where did you find this?

src/Stripe/Services/Subscriptions/StripeSubscriptionCreateOptions.cs
```diff
@@ -42,6 +42,24 @@ internal string TrialEndInternal
         [JsonProperty("tax_percent")]
         public decimal? TaxPercent { get; set; }
 
+        public DateTime? BillingCycleAnchor { get; set; }
+
+        public bool BillingCycleAnchorNow { get; set; }
+
+        [JsonProperty("billing_cycle_anchor")]
+        internal string BillingCycleAnchorInternal
```
jt000: I'm not seeing this one either https://stripe.com/docs/api/curl#create_subscription. Maybe I'm looking in the wrong place?

src/Stripe/Services/Customers/StripeCustomerCreateOptions.cs
pedropaf: Stripe Internal Documentation, the guys from Support told me that because without this I need two API calls to crate a customer + subscription + tax percent. With this property it can be all done in one call.

src/Stripe/Services/Subscriptions/StripeSubscriptionCreateOptions.cs
pedropaf: Same thing, also Internal Stripe documentation. But this production ready and supported by Stripe, but they haven't made this public yet. Also another library in PHP supports this: https://github.com/laravel/cashier/pull/114

src/Stripe/Services/Customers/StripeCustomerCreateOptions.cs
pedropaf: What do you think about this pull request? It's all officially supported by Stripe. I have also tested all this, and it works perfectly. Could you please get it merged for the next release? Thanks

readme.md
```diff
@@ -138,22 +138,22 @@ customer or a charge, but only used once.
...
+		CardExpirationMonth = "10",
+		CardAddressCountry = "US",                // optional
+		CardAddressLine1 = "24 Beef Flank St",    // optional
+		CardAddressLine2 = "Apt 24",              // optional
+		CardAddressCity = "Biggie Smalls",        // optional
+		CardAddressState = "NC",                  // optional
+		CardAddressZip = "27617",                 // optional
+		CardName = "Joe Meatballs",               // optional
+		CardCvc = "1223",                         // optional
+	};
```
jaymedavis: oh snap, I need to fix a bunch of these probably!

src/Stripe/Services/Subscriptions/StripeSubscriptionCreateOptions.cs
jaymedavis: What is this?

pedropaf: ## Stripe Internal Documentation

At creation time, passing a non-now value for billing_cycle_anchor allows you to set up a subscription to have its first "regular" billing date be at a specified point in the future. In particular (and unlike trial_end) billing_cycle_anchor lets you make a prorated charge for a fraction of the month, and then "snap" the billing cycle to the day that you desire.

So, for instance, if on the 20th of January you did a create subscription request with billing_cycle_anchor equal to February 1st, your customer would immediately be charged for about 1/3 of a month, and then subsequently be charged the full amount on the 1st of each month, going forward.

Additionally, when updating an existing subscription, you can pass billing_cycle_anchor=now (you can pass the string "now", just like for trial_end) to force the billing cycle to "reset" and be anchored "now", the way that it would if you were changing intervals.

Concretely, if you had a subscription that billed on the 1st of the month and then on the 18th of February you made an update request with billing_cycle_anchor=now, the subscription would immediately bill in full for the month of February 18th - March 18th, and then bill on the 18th of the month going forward.

src/Stripe/Services/Subscriptions/StripeSubscriptionCreateOptions.cs
```diff
@@ -42,6 +42,24 @@ internal string TrialEndInternal
...
+        {
+            get
+            {
+                if (EndTrialNow)
```
jaymedavis: Looks like this should be BillingCycleAnchorNow?

jaymedavis:
I'm typically careful adding things that aren't in the api docs, but based on the other libraries support and the documentation you provided I'm fine with this. Thanks for documenting TaxPercent in the readme. I don't think BillingCycleAnchor needs to be documented. People that don't know what it is, probably wont need it. :) ... that sure is a weird property name.

I left one comment on the code that looks like a potential bug (probably from copy/paste on trial_end). Let me know if that's the case and I'll get this added to the next release.

src/Stripe/Services/Subscriptions/StripeSubscriptionCreateOptions.cs
pedropaf: yes, a copy paste bug, can you fix it when you merge it? I can send a new pull request otherwise. Thanks!

pedropaf:
@jaymedavis it sounds all good, I replied to your comment, I introduced a copy / paste bug. About BillingCycleAnchor I guess that it's an edge case and the guys from Stripe didn't want to add it to the documentation for that reason. But it's being used and supported. Thanks for merging this!

src/Stripe/Services/Subscriptions/StripeSubscriptionCreateOptions.cs
jaymedavis: sure thing - I got it.

src/Stripe/Services/Customers/StripeCustomerCreateOptions.cs
btriad: Hi.
   I tried to use TaxPercent but I got an error: "stripe net Invalid decimal: 23.0000 must contain at maximum two decimal places."
I downloaded the source code, changed TaxPercent from decimal to string and problem solved since I was able to send formatted string.

PR #296, opened 2015-02-28
Change closed and forgiven to nullable for StripeInvoiceUpdateOptions.

dabster:


jaymedavis:
this is completed - it will be released when 3.0 goes live. thanks!!

PR #297, opened 2015-03-02
Add support for the 'source' parameter when creating customers

ryan-codingintrigue:
As per the [Saving credit card details for later](https://stripe.com/docs/tutorials/charges#saving-credit-card-details-for-later) article on the docs, the `source` parameter is required when creating a customer so that a customer can be both charged and saved to an internal database using a single Stripe.JS/Checkout token.

[API Docs for "source"](https://stripe.com/docs/api#create_customer)

Here's a C# version of the code in the linked article, using the new `SourceTokenId` included.
```c#
// Create a Customer
var customer = _customerService.Create(new StripeCustomerCreateOptions
{
    SourceTokenId = tokenId, // tokenId retrieved from Stripe.JS/Checkout
    Description = "payinguser@example.com"
});

// Charge the Customer instead of the card
var charge = _chargeService.Create(new StripeChargeCreateOptions
{
    CustomerId = customer.Id,
    Amount = 1000, // in cents
    Currency = "usd"
});

// Save the customer ID in your database so you can use it later
CustomerDb.Add("test@email.com", customer.Id);

// Later...
var customerId = CustomerDb.GetStripeCustomerId("test@email.com");
var chargeLater = _chargeService.Create(new StripeChargeCreateOptions
{
    CustomerId = customer.Id,
    Amount = 1500, // $15.00 this time
    Currency = "usd"
});
```

I've *attempted* to add unit tests for this. I'm sorry if they don't work, but I'm not familiar with writing MSpec tests,

### DIFF
--- a/src/Stripe.Tests/customers/when_creating_a_customer_with_a_source_token_id.cs
+++ b/src/Stripe.Tests/customers/when_creating_a_customer_with_a_source_token_id.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using Machine.Specifications;
+
+namespace Stripe.Tests
+{
+    public class when_creating_a_customer_with_a_source_token_id
+    {
+        protected static StripeCustomerCreateOptions StripeCustomerCreateOptions;
+        protected static StripeCustomer StripeCustomer;
+        protected static StripePlan StripePlan;
+        protected static StripeCoupon StripeCoupon;
+        protected static StripeCard StripeCard;
+
+        private static StripeToken _stripeToken;
+        private static StripeCustomerService _stripeCustomerService;
+
+        Establish context = () =>
+        {
+            var _stripePlanService = new StripePlanService();
+            StripePlan = _stripePlanService.Create(test_data.stripe_plan_create_options.Valid());
+
+            var _stripeCouponService = new StripeCouponService();
+            StripeCoupon = _stripeCouponService.Create(test_data.stripe_coupon_create_options.Valid());
+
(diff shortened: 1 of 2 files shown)


jt000:
I could be mistanken, but wasn't ```source``` added in [2015-02-18](https://stripe.com/docs/upgrades#2015-02-18)? Stripe.NET is currently supporting "2015-02-10" (not sure @jaymedavis's timeline on making the leap, but I assume this would be included in that upgrade)

ryan-codingintrigue:
@jt000 you're correct, I should've checked the changelog first. This is something we were working on in the office anyway so thought I'd contribute.

jt000:
I think for v2015-02-10 you can use [```Customer.Card```](https://github.com/jaymedavis/stripe.net/blob/master/src/Stripe/Services/Customers/StripeCustomerCreateOptions.cs#L14). 

*Just to note: I'm not the owner of the repository, just making an observation. Jayme has the word on what goes in or not...*

jaymedavis:
This api jump will be pretty big. The documentation states that the StripeCharge object will have "source" now as well.

This PR handles the change in StripeCustomerCreateOptions, but it will also be required in StripeCustomerUpdateOptions, StripeChargeCreateOptions, and probably other places.

StripeRecipients don't appear to be updated, but that's probably because they are getting rid of recipients?

Also, StripeToken still has "card". 

A lot more research/planning will need to be done before any changes in that api release can be added. Thanks for getting the ball rolling, but a lot more work needs to be done in this area.

PR #299, opened 2015-03-04
Adding StripeRefundService

jt000:
Adding StripeRefundService to support Create\Update\Get\List of StripeRefund objects

### DIFF
--- a/src/Stripe.Tests/Stripe.Tests.csproj
+++ b/src/Stripe.Tests/Stripe.Tests.csproj
@@ -122,6 +122,11 @@
     <Compile Include="recipients\when_getting_a_recipient.cs" />
     <Compile Include="recipients\when_listing_recipients.cs" />
     <Compile Include="recipients\when_updating_a_recipient.cs" />
+    <Compile Include="refunds\when_listing_refunds.cs" />
+    <Compile Include="refunds\when_getting_a_refund.cs" />
+    <Compile Include="refunds\when_updating_a_refund.cs" />
+    <Compile Include="refunds\when_partially_refunding_a_charge.cs" />
+    <Compile Include="refunds\when_refunding_a_charge.cs" />
     <Compile Include="requestor\when_creating_a_request.cs" />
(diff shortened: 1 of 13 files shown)


jaymedavis:
This looks very well done! The only thing I'm iffy on is the Reason object. I have generally tried to avoid enums or constants stripe defines, just so maintenance basically doesn't exist. :)

jt000:
I assume you're referring to the StripeRefundReasons class. This is very similar to what I did to provide constants to Events & in other PRs. 

Personally, I'd suggest having everything specified in enums. That's part of the allure of having type-safety is to get build breaks more often if you do something incorrect. The looser the coupling the more something could change & cause runtime issues in production instead of compile-time during development.

Take for example the change in [2015-02-16](https://stripe.com/docs/upgrades#2015-02-16) where the event name changed from "transfer.canceled" to "transfer.reversed". This will break a lot of people that unknowingly upgrade the Stripe.NET version that takes that API version. However, if there were type-safety (for example, enums instead of strings or they used a constant field), then they'd get a build-break before they introduced that change into their product.

Just my 2 cents... I can remove the constants class if you'd prefer

xeolin:
I totally agree with jt000, Enums are the way to go imho (it really helps with automated refactoring too).

jaymedavis:
Good points. :)

I'm ok with leaving it in, but I think it's important to slowly start converting other parts of the library (that don't use classes/enums) for consistency.

I think the static class is very easy to implement, so I like it for that reason. Did you expirement with an enum instead by chance?

jaymedavis:
I also appreciated that you marked the previous method Obsolete, and I can review it in the next breaking release. I need to get better about that. :)

jt000:
Thanks.

When I did the events list I shortly played around with enums, but figured that'd be a breaking change & would be a little tricky to implement in the conversion to form-data so it didn't make it much farther than the drawing board.

jaymedavis:
nice work. moved StripeRefundReaasons to constants (folder previously named types), but that's about it. will release soon

PR #300, opened 2015-03-09
Change amount to nullable

jslaybaugh:
Amount has to be nullable or if you are updating some other property the amount gets set to zero.


jaymedavis:
This sounds bad. lol

I'll fix it asap.

jaymedavis:
Currency is also on this entity now, which is not listed on the Stripe documentation - we need to figure out if Stripe plans on supporting this in the future or not.

jaymedavis:
this has been merged in and will be part of 3.0 - thanks!

Issue #315, opened 2015-03-24
Invoice Create - multiple parameter support


Chizzl:
At the moment we van only pass in the customer parameter to create an invoice immediately

https://github.com/jaymedavis/stripe.net/blob/master/src/Stripe/Services/Invoices/StripeInvoiceService.cs

But Stripe can have multiple parameters > https://stripe.com/docs/api#create_invoice

E.g. I can not set a correct tax % on the invoice I want to create after plan change.

jaymedavis:
this will be fixed with the next release (addressed in the pr)

PR #318, opened 2015-03-24
invoice create - adding optional parameters

Chizzl:
Fix for issue #315 

Excuse me for the 2 previous invalid pull requests.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -465,11 +465,11 @@ Invoices
 ### Create a customer invoice
 
 	var invoiceService = new StripeInvoiceService();
-	StripeInvoice response = invoiceService.Create(*customerId*);
+	StripeInvoice response = invoiceService.Create(*customerId*, stripeInvoiceCreateOptions);
 
 ### Updating a customer invoice
 
-	var stripeInvoiceUpdateOptions = new StripeInvoiceUpdateOptions();
+	var stripeInvoiceUpdateOptions = new StripeInvoiceUpdateOptions(*invoiceId*);
 	stripeInvoiceUpdateOptions.Closed = true;
 
 	var invoiceService = new StripeInvoiceService();
--- a/src/Stripe/Services/Invoices/StripeInvoiceCreateOptions.cs
+++ b/src/Stripe/Services/Invoices/StripeInvoiceCreateOptions.cs
@@ -1,0 +1,30 @@
+﻿using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Stripe.Services.Invoices
+{
+    public class StripeInvoiceCreateOptions
+    {
+
+        [JsonProperty("customer")]
+        public bool CustomerId { get; set; }
+
+        [JsonProperty("application_fee")]
+        public int? ApplicationFee { get; set; }
+
+        [JsonProperty("description")]
+        public bool Description { get; set; }
+
+        [JsonProperty("metadata")]
+        public Dictionary<string, string> Metadata { get; set; }
+
+        [JsonProperty("statement_descriptor")]
+        public string StatementDescriptor { get; set; }
+
+        [JsonProperty("subscription")]
+        public string SubscriptionId { get; set; }
+
+        [JsonProperty("tax_percent")]
+        public int? TaxPercent { get; set; }
+    }
+}
--- a/src/Stripe/Services/Invoices/StripeInvoiceCreateOptions.cs
+++ b/src/Stripe/Services/Invoices/StripeInvoiceCreateOptions.cs
@@ -5,10 +5,6 @@ namespace Stripe.Services.Invoices
 {
     public class StripeInvoiceCreateOptions
     {
-
-        [JsonProperty("customer")]
-        public bool CustomerId { get; set; }
-
         [JsonProperty("application_fee")]
         public int? ApplicationFee { get; set; }
 
--- a/src/Stripe/Services/Invoices/StripeInvoiceService.cs
+++ b/src/Stripe/Services/Invoices/StripeInvoiceService.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using Stripe.Services.Invoices;
 
 namespace Stripe
 {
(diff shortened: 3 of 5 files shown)


src/Stripe/Services/Invoices/StripeInvoiceService.cs
```diff
@@ -63,10 +64,11 @@ public virtual IEnumerable<StripeInvoice> List(StripeInvoiceListOptions listOpti
             return Mapper<StripeInvoice>.MapCollectionFromJson(response);
         }
 
-        public virtual StripeInvoice Create(string customerId)
+        public virtual StripeInvoice Create(string customerId, StripeInvoiceCreateOptions createOptions = null)
```
jt000: I'd suggest creating a separate overload for Create. One that takes the customerId (for backcompat) and one that takes only the createOptions. That way there isn't a duplicate CustomerId and this method follows convention of the other Create methods in services...

...or maybe remove CustomerId from StripeInvoiceCreateOptions.

Chizzl:
The customerId wasn't supposed to be in there.  Thanks for mentioning.

src/Stripe/Services/Invoices/StripeInvoiceService.cs
Giovani-Brady: Yes i agree, while declaring the createOptions parameter null does make i backwards compatibility for all server side implementations, client side implementations still read and require that optional parameter. So basically this change will break every JS-based implementation.

As far as web goes.

jaymedavis: Off the top of my head, I think i'd remove the customer id from StripeInvoiceCreateOptions. Without looking through the other services, I believe most of them follow the convention of passing some sort of id before an options object. I would have to look through to verify I'm remembering correctly.

Chizzl:
Any actions required from my side to close this issue?

src/Stripe/Services/Invoices/StripeInvoiceService.cs
```diff
@@ -63,10 +64,11 @@ public virtual IEnumerable<StripeInvoice> List(StripeInvoiceListOptions listOpti
...
         {
             var url = Urls.Invoices;
             url = ParameterBuilder.ApplyParameterToUrl(url, "customer", customerId);
+            url = this.ApplyAllParameters(createOptions, url, true);
```
jaymedavis: there should probably be a null check here before ApplyAllParameters is called

jaymedavis:
I think this implementation is fine, but it will need to be in a 3.0 release since it's a breaking change (for the reason @Giovani-Brady mentioned). There's a lot of PR's that will be breaking changes, I'm going to try and get a lot of them (or all of them) in 3.0.

src/Stripe/Services/Invoices/StripeInvoiceService.cs
jaymedavis: this last parameter should be false, (isListMethod) - I've fixed it, just fyi

src/Stripe/Services/Invoices/StripeInvoiceCreateOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Stripe.Services.Invoices
+{
+    public class StripeInvoiceCreateOptions
+    {
+        [JsonProperty("application_fee")]
+        public int? ApplicationFee { get; set; }
+
+        [JsonProperty("description")]
+        public bool Description { get; set; }
```
jaymedavis: this should be string, i've fixed, fyi

src/Stripe/Services/Invoices/StripeInvoiceCreateOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+        public Dictionary<string, string> Metadata { get; set; }
+
+        [JsonProperty("statement_descriptor")]
+        public string StatementDescriptor { get; set; }
+
+        [JsonProperty("subscription")]
+        public string SubscriptionId { get; set; }
+
+        [JsonProperty("tax_percent")]
+        public int? TaxPercent { get; set; }
```
jaymedavis: this should be decimal, i've fixed. fyi

readme.md
```diff
@@ -465,11 +465,11 @@ Invoices
 ### Create a customer invoice
 
 	var invoiceService = new StripeInvoiceService();
-	StripeInvoice response = invoiceService.Create(*customerId*);
+	StripeInvoice response = invoiceService.Create(*customerId*, stripeInvoiceCreateOptions);
```
jaymedavis: modified this to be like the other ones (convention):
StripeInvoice response = invoiceService.Create(*customerId*); // optional StripeInvoiceCreateOptions

readme.md
```diff
@@ -465,11 +465,11 @@ Invoices
...
 
 ### Updating a customer invoice
 
-	var stripeInvoiceUpdateOptions = new StripeInvoiceUpdateOptions();
+	var stripeInvoiceUpdateOptions = new StripeInvoiceUpdateOptions(*invoiceId*);
```
jaymedavis: this was not needed

jaymedavis:
this is merged in. there were a few fixes, but it will be part of 3.0. thanks!

jt000:
Where was this merged? I don't see it at https://github.com/jaymedavis/stripe.net/tree/master/src/Stripe/Services/Invoices and there doesn't appear to be any sub-branches... Is there a different repository for 3.0?

jaymedavis:
it will be part of 3.0 - that hasn't been released yet. If you want to see everything coming in the next release, I've started keeping the next release on a draft: https://github.com/jaymedavis/stripe.net/releases/tag/untagged-64c6401be66a38ad66f2

jaymedavis:
I might go ahead and get a release out this weekend. I'm trying to not clutter the codebase with obsolete items, but unless I want to be at release 15.0 in a few months I might have to start doing that.

jt000:
That link returned a 404 for me. Have you pushed it yet? The [3.0 tag](https://github.com/jaymedavis/stripe.net/releases/tag/v3.0.0) looks like it only has a single file change that isn't this PR

jaymedavis:
Ah, thanks for pointing that out. I had to publish the release as pre-release, i was just saving it as a draft.

PR #338, opened 2015-05-19
Add destination field to create charge options - allows charges to be…

aidanbebbington:
… created on behalf of connected accounts


jaymedavis:
this has been added and will be part of the 3.0 release. thanks!

Issue #328, opened 2015-04-15
Support for Stripe-Account header for connected accounts


sandeep-at-airlift:
As of now in Stripe.net the [suggested way](https://github.com/jaymedavis/stripe.net/pull/43#issuecomment-10903921) to support connected accounts for a platform is to use access_tokens for the connected account.

This can be cumbersome to manage for many connected accounts. Also could be a security hole if the access token is presentable by itself (have **not** verified this).

Stripe documentation [recommends](https://stripe.com/docs/connect/authentication#authentication-via-the-stripe-account-header) using the platform secret key along with the Stripe-Account header set to the account id of the connected account. 

I could see adding a property to `StripeService` for the Stripe-Account id and then in `Requestor::GetWebRequest()` where the `WebRequest` is created set the header if this property is set. 

Only problem is there will be a proliferation of overloads on all services and new methods on `Requestor` as we can't support api-key and/or  Stripe-account-id as optional parameters.

Thoughts on how to fix this?

jbarker4682:
I would create a RequestOptions class of some type, with these properties and then replace all references to any optional parameters with this class instead (ex: Foo(string apiKey, string accountId = null) becomes Foo(RequestOptions options) { ... }

jaymedavis:
The one I'm floating around right now is having each service have properties like    UseConnectedAccountId = ... UseIdempotencyKey = .... 

With that said, I think this is a great idea and I will definitely experiment with that or a combination of the two. For this pull request, I'm going to close it in favor of managing the work on PR# #348

PR #348, opened 2015-06-02
add support for Stripe-Account header - to make requests on behalf of connected accounts

aidanbebbington:
Feels a bit wrong that I changed so much stuff to set one header, but just a little hesitant to make large structural changes to the project at this stage, and this seemed like it followed the existing structure best.  Let me know if you want it refactored though.

### DIFF
--- a/src/Stripe.Tests/infrastructure/test_data/test_service.cs
+++ b/src/Stripe.Tests/infrastructure/test_data/test_service.cs
@@ -8,7 +8,7 @@ namespace Stripe.Tests.test_data
 {
     public class test_service : StripeService
     {
-        public test_service() : base(null)
+        public test_service() : base(null, null)
         {
         }
     }
--- a/src/Stripe.Tests/infrastructure/when_building_parameters_with_expandable_properties.cs
+++ b/src/Stripe.Tests/infrastructure/when_building_parameters_with_expandable_properties.cs
@@ -10,7 +10,7 @@ namespace Stripe.Tests
     {
         public class TestService : StripeService
         {
-            public TestService() : base(null) { }
+            public TestService() : base(null, null) { }
             public bool ExpandSimple { get; set; }
             public bool ExpandMultiWordProperty { get; set; }
         }
--- a/src/Stripe/Infrastructure/Requestor.cs
+++ b/src/Stripe/Infrastructure/Requestor.cs
@@ -7,35 +7,35 @@ namespace Stripe
 {
     internal static class Requestor
     {
-        public static string GetString(string url, string apiKey = null)
+        public static string GetString(string url, string apiKey = null, string stripeAccountId = null)
         {
-            var wr = GetWebRequest(url, "GET", apiKey);
+            var wr = GetWebRequest(url, "GET", apiKey, stripeAccountId);
 
             return ExecuteWebRequest(wr);
         }
 
-        public static string PostString(string url, string apiKey = null)
+        public static string PostString(string url, string apiKey = null, string stripeAccountId = null)
         {
-            var wr = GetWebRequest(url, "POST", apiKey);
+            var wr = GetWebRequest(url, "POST", apiKey, stripeAccountId);
 
             return ExecuteWebRequest(wr);
         }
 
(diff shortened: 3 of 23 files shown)


hkenuam:
Hi, I'm setting up Connected Accounts and really could use this commit. Any eta on when it will be merged?

jaymedavis:
I think it would be more intuitive to call the variable something like UseConnectedAccountId on the StripeService base class (otherwise, it's hard to know what StripeAccountId is from just looking at the code). 

I see why you had to do it the way you did, but I wonder if there is an easier way (with reflection or otherwise) to check the caller and look for the property. I would see it being used sort of like this:

var service = new StripeCustomerService();
service.UseConnectedAccountId = "blahblah";
service.DoSomething()

Then maybe the requestor could determine from the caller if that variable was set (by convention) - would save the need for a lot of code. 

I'm not sure how feasible this is, just a thought. :)

jaymedavis:
I will be investigating this more in the next few days to see if this is the best way to go about it, or if there is something else we can do. With the idempotency key too, I don't want to get in a situation where the services all take a lot of parameters. With that said, I completely understand why it was developed this way. :)

I will spend some time trying to learn more about Stripe Connect.

jbarker4682:
+1.  It looks like there are numerous things in the works, so not sure where I can jump into help... But let me know if there is anything I can do...

jaymedavis:
I think the method @jbarker4682 mentioned on #328 would be a great way to handle this. 

As is, with the problem we need to solve:
1) Every service now takes an apikey as an optional parameter.
2) All of the services are going to need a few options in addition to this parameter:
* api key
* idempotency key
* Stripe-Account
* Use Bearer (for oauth)

I think a good solution to this would be to mark the api key option obsolete for the next version, and for now add another option to each service that take a RequestOptions object. This object would give us the available functionality needed for all of these issues.

Something like this:

var service = new CustomerService();
service.RequestOptions = new RequestOptions { ApiKey = 'blah blah', StripeConnectedAccountId = 'blah blah' };

service.DoYourThing()

This is a pretty big change, but ultimately seems like the best route to give us the ability to support all of these cases.

Feel free to chime in. :)

jaymedavis:
Actually, this might really dirty up the codebase to mark so many things obsolete. This will have to touch every single service method, so I may just make this a major release.

jbarker4682:
Hi Jayme,

To clarify, anything that would only be set once per life if a service should ideally be injected into the service constructor as an options param.  Anything that can change from one call to the next should be passed in to the method directly as its own options param.  Also, you can maintain backwards compatability by leaving your existing method signatures as they are and creating new overloads that accepth the new options object...  to prevent code duplication, move the body of each existing function inside its newer version, and call the new version from the old version...

jaymedavis:
That makes sense. I didn't consider the fact that it would be annoying to have to re-construct the service if you want to do something different with the request options. That would be especially annoying for anyone using dependency injection.

For the second part, that sounds like it would be a little bit of bloat, but I guess it's not too bad. Doing it that way would also not require a major version change, since it's just a new method signature with the options.

I like it, Thanks!

jaymedavis:
FYI - I'm actually going to skip the overloads and just add StripeRequestOptions = null on the existing signatures. This may be a breaking change for some, but most people should not have to make any changes it all. It also reduces the code for future use in building async, etc.

jaymedavis:
@aidanbebbington did you have this working? I cannot seem to get it to go. I keep getting an authorization error when I try to access the connected account with the account_token returned from the oauth request. 

Short of figuring this out (I've reached out to Stripe), this is pretty much ready to go.

aidanbebbington:
Hi Jayme,

Sorry for my lack of attention on this. Been busy at work. Yep, it is
currently powering pitchi.com! I didn't encounter any errors. Can you send
a code sample?

Cheers
Aidan

jaymedavis:
I think I have the problem worked out - I was sending the secret key on the response as the stripe-account header instead the acct_* value. I'll let you know if I still need help after this - thanks!!

jaymedavis:
Cool site, btw - love the hero image.

PR #351, opened 2015-06-09
add support for unchanged billing_cycle_anchor

jdc0589:
add support for "unchanged" billing_cycle_anchor".

From IRC: 
```
<jdc0589> can someone confirm that the valid values for "billing_cycle_anchor" for a subscription update are undefined, "now", and "unchanged" ?
[11:09] <jdc0589> and does undefined have the same behavior as "unchanged" /
[11:10] <praboud> jdc0589: now and unchanged are the possible values (as well as not passing it at all)
[11:11] <praboud> unchanged will force the anchor to not change when it otherwise would
[11:11] <praboud> like if you go from an unpaid subscription to a paid one, you'd usually reanchor the billing cycle
```


jaymedavis:
Looks good - I'll get this added asap. :)

PR #366, opened 2015-07-24
StripeSourceOptions updated with correct name

SimonAntony:
Previously was creditCardOptions - this was a fix as a result of missing code in the downloaded zip, however it does appear to be in the repo so not sure why they are different.


jaymedavis:
whoops, thanks :)

PR #368, opened 2015-07-27
expose action to fetch an invoice's items

aidanbebbington:
https://stripe.com/docs/api#invoice_lines

### DIFF
--- a/src/Stripe/Infrastructure/Urls.cs
+++ b/src/Stripe/Infrastructure/Urls.cs
@@ -12,6 +12,11 @@
             get { return BaseUrl + "/invoiceitems"; }
         }
 
+        public static string InvoiceItemsForInvoice
+        {
+            get { return BaseUrl + "/invoices/{0}/lines"; }
+        }
+
         public static string Tokens
         {
             get { return BaseUrl + "/tokens"; }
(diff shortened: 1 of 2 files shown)


jaymedavis:
This will need to be rebased and have StripeRequestOptions as a parameter: if you can that, great - if not I can handle it when I merge it. thanks!

jaymedavis:
few more things in addition to StripeRequestOptions

1) This needs to be on the invoice service instead of the invoice item service, that's where it's listed in the docs on Stripe
2) This needs to return an ienumerable, StripeList is type used on the entities, but the services all return ienumerable's

Other than that, I think it would work fine, but tests would be nice to have as well.

daniel-cohen:
@jaymedavis I've created a PR (#525) for this because as far as I can see it has never been implemented before. I'd appreciate it if you could have a look and merge it if you're happy with it.

jaymedavis:
Thanks very much @daniel-cohen. The PR you sent looks good, and I will work from there and close this PR. The only thing I noticed missing off the top of my head was async. I can add that in when I merge though. This looks pretty complete (thanks for testing!) and I will try to get this released soon. Thanks!

PR #371, opened 2015-08-08
Port to Windows 10 Universal

chickenham:
I ported stripe.net for Windows 10 Universal app development.  I created a new csproj, Stripe.Universal.csproj.  The major change I had to make was to create a replacement for Requestor.cs (which I call Requestor.Universal.cs) that uses Windows.Web.Http.HttpClient.


rmisiak:
My PR upgrades the sln to VS 2015 - I can submit a new PR that leaves the original SLN intact and has a separate VS 2015 sln if you'd like.

jaymedavis:
I quickly looked through this to get a feel for it, does this make every request async? Also, will this *only* work on windows 10 or is it possible to just bake in directly?

rmisiak:
It doesn't make every request async.  It runs on Microsoft's new Windows Universal App Platform (https://msdn.microsoft.com/en-us/library/Dn894631.aspx).  This is MS's new platform rolled out with Windows 10 that allows easy development of device-independent (PC, tablet, phone, even Xbox) apps.  It does require Windows 10 and Visual Studio 2015.  I will submit a new PR with changes to the sln files.

rmisiak:
Ok, I pushed the sln change.

jaymedavis:
Thanks so much for the PR - I'll give this some more time as soon as I can. There are a few things in front of it (dropping 3.5, etc). 

Thanks again for he help!

rmisiak:
Sure no problem.  I'm building an app for Windows 10 desktop+phone that utilizes it - I'll send the info on the app when it's available in case you're curious.  In the meantime I'll try to stay on top of merging any commits.

jaymedavis:
Awesome, thanks.

rmisiak:
I merged the latest stripe.net master with my PR.  Do you have any sense for when you will merge this PR?

FYI, our app using Stripe.net Universal code is live in the Windows store: https://www.microsoft.com/en-us/store/apps/ring-doorbell/9nblggh1qwd4

jaymedavis:
that's awesome! your app looks really nice :)

I still have to install visual studio (latest version) on my windows 10 VM (new) - I am halfway there, it's just a matter of finding time to finish that up. I need to review the code a bit, but it seems like it works pretty well for you since your app is live. Mainly, I need to test backwards compatiblility. I'll shoot for this in the next release since you've been kind enough to keep the PR up to date. Hopefully, I'll have a little time this weekend to knock it out.

jaymedavis:
can you summarize what's different in the requestor class/es? it looks almost identical from a quick look. Just trying to make it as lean as possible. thanks

rmisiak:
it's basically identical.  The only difference is that it's using Windows.Web.Http.HttpClient, a different brand of HttpClient which is used in WinRT.  WinRT doesn't fully support WebRequest.Create() like you are doing it in the original Requestor.cs

jaymedavis:
Would it be nasty to stick a few if windows_uwp's in there instead of a separate class? I would prefer to avoid the duplication of its not too messy

rmisiak:
I started out that way but the code became what I thought was an unreadable mess, that's when I moved over to the current version.  The reason it became unreadable is virtually all methods in the class have different code, although I maintained the same method names and signatures.  However, if you'd prefer it that way, I can implement it as such.  Let me know.

rmisiak:
Alternatively, to make things a little more clear, I can change both classes to be instance classes and create an IRequestor interface.  Then make the classes named something like DefaultRequestor or UniversalRequestor.

jaymedavis:
That sounds promising. By doing that, you could also share some code on a base class to reduce the duplication as well (maybe pass Func's to the methods?) - the Func being either the current way or the uwp way

rmisiak:
I'm on it.  I should have a revised PR for you by tomorrow.

jaymedavis:
Awesome. The nice part will be there is lots of test coverage in that area seeing as everything uses it :)

rmisiak:
Ok, I implemented the changes as discussed.  There is now an IRequestor interface, and a Requestor abstract class which contains shared code used by two new classes, DefaultRequestor and UniversalRequestor.  Also I was able to eliminate ParameterBuilder.Universal.cs by making a simple change to the default ParameterBuilder.cs file.

jaymedavis:
I started to pull this down and test, but am getting several build errors. Some were simple, just problems in the 3.5 library. Some are places where I think Requestor just needs to use Requestor.Instance. I need to get a bug fix release out, but I'll revisit this soon. Just FYI

advancedrei:
Any word on this? I just ported a bunch of my stuff over to start getting ready for future architectures, and I'm now stuck on this. Anything I can do to help bring it on home?

jaymedavis:
It will get in eventually, but a few things need to be updated:
1) I couldn't build it
2) I would prefer (if possible) to keep one project with another build target.
3) Requestor.Instance is a pretty big change, I'm hoping for someway to minimize this, but I don't have a solution yet.

1 and 2 are the big things, otherwise it's just time to get to this. At the moment the focus is Managed Accounts.

advancedrei:
I'll see what I can do with a new PR from your current codebase. The problem with Windows 10 is that you now almost *have* to have two projects in the same solution with multiple sets of binary outputs. Because the project systems are vastly different. The problem **I'm** currently having with this branch is that it ONLY targets Windows 10, and it won't work for Portable projects that target any part of Microsoft's new .NET stack.

I think the best solution is to make this project Portable, and take a dependency on the System.Net.Http NuGet package. Then we can have one solution, but multiple projects targeting different outputs, and a batch build configuration that dumps them all. This would give you the widest coverage and device support.

I'd also make changes to the NuSpec file to support these multiple configurations, as well as add Source Packages support to aid with debugging. It would be a similar structure to what I've implemented with the PortableRest project.

If you're OK with the concept of the end result, how would you like me to approach getting it to you o it doesn't mess up your current work, and doesn't create one giant PR? LMK and I'll open up a new issue for it and get this knocked out in the next day or so. Thanks man :)

jaymedavis:
The NuGet issues with how to release this are important to me. As it is, I think I have it wrong because the build versions should go in a folder (I haven't researched this much yet). It's obviously very important to me that existing users have 0 issues with the new bits. If you have a vision of the entire solution, that would be much appreciated. :)

The easiest way for me to handle this is small pull requests that don't interfere with existing functionality. As small as you can logically break the pieces down, the easier it is on me to test and merge them. Even if it's not a fully working version at first, at least it will be heading that way in master.

advancedrei:
Do you want one PR per checkin? Or multiple checkins but one PR? I'm ready to start sending you code one way or another.

jaymedavis:
you can do the checkins however you prefer, I just look at the diff anyway. When it goes to master, if it's too many commits I'll just rebase it anyway.

jaymedavis:
Just do one PR for now with your little bit of work.. When we get that squared away, you can do another. :)

jaymedavis:
this is completed - thank you guys so much for starting this ball rolling. @chickenham - I love my Ring, and signed up for a year! :)

PR #372, opened 2015-08-15
Adjust documentation for Charge

MrDesjardins:
As discussed by Twitter https://twitter.com/jaymed/status/632571689877008384 I think this need to be adjusted to reflect changes in Stripe.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -137,10 +137,10 @@ Tokens
 A token can be used anywhere on Stripe where you would normally pass a card. Once it's created, it can be used on a
 customer or a charge, but only used once.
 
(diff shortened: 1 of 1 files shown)


jaymedavis:
This class is for creating a token, not for creating a charge. The modification to the documentation here would cause you to create a StripeCharge, not a StripeToken. For creating a StripeCharge, you can reference the documentation [here](https://github.com/jaymedavis/stripe.net#charges).

If you are having trouble creating a token, please email me your code jaymedavis at gmail. I can take a look and see what is going on. I still don't understand if it's a token you want, or a charge.

PR #376, opened 2015-08-25
Set StripeSourceOptions.Object from TokenId instead of requiring specification

jbparker:
Hey @jaymedavis 
I worked recently from the readme.md and noticed that specification of the Object property wasn't included in the examples.

It looks like this was recently added to address changes in Stripe's API - since only cards can (apparently) be specified on charges / customer accounts without a token, this seemed to be a good change to automatically determine the Object value if the token wasn't present.

Let me know if you have any concerns about this.  All tests passed as well after this change.

### DIFF
--- a/src/Stripe.Tests/customers/test_data/stripe_customer_create_options.cs
+++ b/src/Stripe.Tests/customers/test_data/stripe_customer_create_options.cs
@@ -9,7 +9,6 @@ namespace Stripe.Tests.test_data
         {
             var cardOptions = new StripeSourceOptions()
             {
-                Object = "card",
                 AddressCountry = "US",
                 AddressLine1 = "234 Bacon St",
                 AddressLine2 = "Apt 1",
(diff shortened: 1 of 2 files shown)


jaymedavis:
Actually, I would prefer this to stay with specifying card and just updating the documentation. Tucking away the object type into a hardcoded string has bit me before (twice) in this library - Stripe typically exposes it like that for a reason. Being that it's a source, there are other types you could create other than a card. Thanks for trying to simplify it and I think it's a cool idea, but I know I will end up pulling this back out. :)

jbparker:
Ok that works, thanks.

Sent from Outlook




On Sun, Sep 13, 2015 at 1:22 PM -0700, "Jayme Davis" <notifications@github.com> wrote:
Actually, I would prefer this to stay with specifying card and just updating the documentation. Tucking away the object type into a hardcoded string has bit me before (twice) in this library - Stripe typically exposes it like that for a reason. Being that it's a source, there are other types you could create other than a card. Thanks for trying to simplify it and I think it's a cool idea, but I know I will end up pulling this back out. :)

---

PR #380, opened 2015-09-04
Ability to delete discounts that have been applied to customers and subscriptions

bcornett:
Currently, there is no way to remove discounts that have been applied to customers or subscriptions.

Stipe has the endpoint implemented already, shown here: https://stripe.com/docs/api#delete_discount

This pull request adds these two endpoints to the respective services as the "DeleteDiscount" method.

### DIFF
--- a/src/Stripe/Services/Customers/StripeCustomerService.cs
+++ b/src/Stripe/Services/Customers/StripeCustomerService.cs
@@ -63,5 +63,14 @@ namespace Stripe
 
             return Mapper<StripeCustomer>.MapCollectionFromJson(response);
         }
+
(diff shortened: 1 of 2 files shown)


jaymedavis:
I think it would be better ti implement these in a Discount folder/Discount service. Since it's listed separately on the Stripe docs, I think it would be confusing having them as part of their respective services (customers/subscriptions). I try to stay as close the the official Stripe docs as possible. I think this is a great start to get there, but needs to be separated into it's own service.

jaymedavis:
I am going to handle this work on #520 - so closing this for now. I will reference for comparison sake. Thanks much for your contribution!

PR #384, opened 2015-09-10
Update StripeCharge.cs

MattSpeakman:
Added a StripeId property that takes the top level "id" field so that this can be recorded in a database


jaymedavis:
Id exist already on StripeObject (inherited)

MattSpeakman:
Jayme,
Thanks, I didn't spot that. I'll adjust my code to match.

Matt

PR #392, opened 2015-09-22
Added the ability to add/update coupon Metadata

sidshetye:
Stripe allows the addition of Metadata to coupons, which is very useful for storing descriptions or other information related to coupons. The current Stripe.NET API didn't allow for Metadata (simply Key-Value pairs), so this patch extends that.


jaymedavis:
thanks, this is on the develop branch and will be in the next release.

PR #395, opened 2015-09-29
add account create function

AvengingSyndrome:
allows a basic account create with simple options - from https://stripe.com/docs/api/curl#create_account

### DIFF
--- a/src/Stripe.Tests/Stripe.Tests.csproj
+++ b/src/Stripe.Tests/Stripe.Tests.csproj
@@ -33,20 +33,21 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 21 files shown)


jaymedavis:
curious where you found the keys value? I recently updated StripeAccount to be 100% current with the Stripe documentation, and I didn't come across this value. Thanks.

AvengingSyndrome:
If you look at the documentation for https://stripe.com/docs/api#create_account under the "Returns" header you'll see this comment: "Returns the account object, with an additional keys dictionary containing secret and publishable keys for that account." This is only available on account creation not account retrieval.

jaymedavis:
Roger that. Thanks!

jaymedavis:
accounts/managed accounts added

PR #415, opened 2015-11-09
Added update method for charges

oystehei:
The StripeChargeUpdateOptions takes 3 parameters, *description*, *metadata* and *receipt_email*. I left out the *shipping* property, as it is not included in the StripeChargeCreateOptions object.

I tried to include the *fraud_details* property as well, but it kept acting up when I tested it against the Stripe API, giving me an error message that read "Invalid hash". I have no idea what that means, so I just removed it altogether as I don't really need this parameter.

### DIFF
--- a/src/Stripe.Tests/Stripe.Tests.csproj
+++ b/src/Stripe.Tests/Stripe.Tests.csproj
@@ -69,6 +69,8 @@
     <Compile Include="balance\when_listing_balancetransactions.cs" />
     <Compile Include="balance\when_listing_balancetransactions_for_charge.cs" />
     <Compile Include="balance\when_retrieving_a_balance.cs" />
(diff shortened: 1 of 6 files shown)


oystehei:
Any timeline for when this can be merged into the library? Right now I have to roll with my own custom version of it, but I would love to see this in the official NuGet package :)

jaymedavis:
no, sorry. with the holidays upon us and vacation, work, etc..  I am not sure. this pull request does look pretty solid from a quick glance so I don't expect much (if anything) to change

jaymedavis:
this was done very well! I don't often get good tests with PR's :) ... at first I was wondering why you didn't include fraud_details and shipping, then I realized it was going to be a lot more work. lol

I've got this merged in and it will go out with the next release - sometime this weekend.

PR #417, opened 2015-11-29
Update .NET version in Nuspec to lib\net46

itsjason:
Tried to install this package to a .NET 4.5 project. It caused a very confusing error, which I tracked down to the fact that the project could not be built using the referenced Stripe.net assembly. I believe this change is what is needed to let NuGet know that it can't install this version to  4.5 projects.

### DIFF
--- a/Stripe.net.nuspec
+++ b/Stripe.net.nuspec
@@ -17,6 +17,6 @@
     </dependencies>
   </metadata>
   <files>
-    <file src="working40\Stripe.net.dll" target="lib\net40" />
+    <file src="working40\Stripe.net.dll" target="lib\net46" />
   </files>
 </package>


jaymedavis:
What was the error?

jaymedavis:
This attribute might be able to be omitted entirely: 

The best way to test is: pull down this repo, run build.cmd, then run build nuget.cmd. This will build the nuget package - that is simply a zip file. You can rename it to .zip, open it up, grab the assembly and try to reference that in your project. This can verify that a) this fix works   or  b) the target is not needed at all.

jaymedavis:
The nuget file did have the wrong json.net version - which could have caused an issue. I also changed the target to lib which might help, but I don't believe these are related. You might need to add the <startup><supportedRuntime....>...</startup> element to your app/web.config. Let me know if the new version gives you the same issue.

itsjason:
Sounds good, I'll give that a try. Also thanks so much for creating this
library.

jaymedavis:
You are very welcome. :)

If enough people have trouble with the 4.6 upgrade, I'll downgrade the version. I really just needed TLS 1.2 for Stripe (they are heading this direction) and I believe that is available in 4.5 as well.

jaymedavis:
closing this - i downgraded to 4.5

PR #424, opened 2016-01-10
add StripeSubscriptionStatuses

rwhitmire:
I have to keep adding this class to my own applications and it feels like it belongs here instead.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -321,6 +321,14 @@ Subscriptions
 
 [StripeListOptions](#stripelistoptions-paging) for paging
 
+### Statuses
+
+    StripeSubscriptionStatuses.Trialing
+    StripeSubscriptionStatuses.Active
+    StripeSubscriptionStatuses.PastDue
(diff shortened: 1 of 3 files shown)


jaymedavis:
looks good - can you update the readme as well to use this and link to the stripe docs so I can verify all of this easily?

Thanks for the PR.

rwhitmire:
Sure thing! https://stripe.com/docs/api#subscription_object-status

jaymedavis:
I was actually hoping for the Subscription docs to be updated a bit more to show creating a Subscription (vs the lack of detail I have now) while using this new feature. I think that would be easier than just putting a list of statuses at the bottom. 

I can update this when I get some time, but wanted to let you know where my head was. :)

jaymedavis:
I went ahead and added this - I'm leaving it out of the readme for now, because some documentation updates are coming really soon. This will take care of you for now though!

Thanks, much! I'll be checking in sometime this evening.

PR #425, opened 2016-01-10
[PCI / Security] Never create a token server-side

matthewarkin:
In regards to PCI compliance creating a token server side and saving a token server-side are equivalent in scope. Tokens should only be made client side using Stripe.js or Stripe Checkout.

Your server should never have access to card data beyond the last 4 and expiration

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -149,38 +149,6 @@ Coupons (queue-pons not coo-pons)
 Tokens
 ------
 
-### Creating a token
-
-A token can be used anywhere on Stripe where you would normally pass a card. Once it's created, it can be used on a
-customer or a charge, but only used once.
-
-	var myToken = new StripeTokenCreateOptions();
-
-	// if you need this...
-	myToken.Card = new StripeCreditCardOptions()
-	{
-		// set these properties if passing full card details (do not
-		// set these properties if you set TokenId)
-		Number = "4242424242424242",
-		ExpirationYear = "2022",
-		ExpirationMonth = "10",
-		AddressCountry = "US",                // optional
-		AddressLine1 = "24 Beef Flank St",    // optional
-		AddressLine2 = "Apt 24",              // optional
-		AddressCity = "Biggie Smalls",        // optional
-		AddressState = "NC",                  // optional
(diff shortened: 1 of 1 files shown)


jaymedavis:
This is not entirely accurate. While client side is typically the recommended approach, it is possible to have your own servers pass and create the token and be PCI compliant. There are various security protocols that need to be in place, but it is entirely possible. The client side approach is the recommended approach - that's why there is a big video on the readme for how to do it, but it is also possible to be PCI compliant via server traffic.

matthewarkin:
I'm not saying it isn't but when a user is new to PCI compliance and Stripe, they just see your docs, and copy and paste code and they think they can just access card data server-side. The easiest level of PCI compliance (SAQ A or A-EP) would be no server-side access. With SAQ D, server-side access and server-side storage is the same SAQ. 

Personally I'd update all the docs to just show token examples.

Also it makes no sense to make a token server-side since they just last for 5 minutes, if you're going to make a token server-side, you should just send the card data to the customers API or to the charges API.

praboud:
@jaymedavis: I think it's entirely reasonable to document the fact that you can create tokens from stripe.net, as there is a valid use for this in tests. However, the documentation should come with the caveat that this is _only_ appropriate for use in test mode (where you want to test code which in production would take a token from stripe.js as input).

Using serverside tokenization in production provides absolutely no compliance value (or actual security benefits). If I really want to pass cards through my server, I should just bite the bullet and pass the card directly. Using serverside tokenization like this is just cargo cult code.

This distinction isn't currently captured in the documentation, and I think is potentially a pitfall for new users. If you think this is reasonable, I'd be happy to draft up a blurb documenting the above.

jaymedavis:
I would be fine with a caveat that this is only valuable if your server is PCI compliant. I'm also fine with saying if you are PCI compliant, you may not want to deal with tokens at all and just pass them directly with ephemeral storage. I'm sure there may be cases of nieve copy/paste from some users that don't understand the consequences, but there's also cases of people that have dealt with PCI compliance for a long time. :)

I closed this request because I believed simply deleting the code from the readme was not the right solution.

PR #426, opened 2016-01-11
Update readme sections about tokens & recipients

praboud:
r? @jaymedavis 

For reference, discussion about tokenization docs in #425.

I also stuck in a quick comment near recipients to note that they've been deprecated. I stuck this in this PR purely in the name of convenience - if you'd prefer to cut that into a different PR to keep them logically separate, I'd be fine with that.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -589,6 +589,9 @@ Disputes
 Recipients
 ----------
 
+**Note: recipients have been deprecated by Stripe - please use
+[Stripe Connnect](https://stripe.com/docs/connect) instead**
+
 ### Creating a recipient
 
 	var myRecipient = new StripeRecipientCreateOptions();
--- a/readme.md
+++ b/readme.md
@@ -154,10 +154,18 @@ Tokens
 A token can be used anywhere on Stripe where you would normally pass a card.
 Once it's created, it can be used on a customer or a charge, but only used once.
 
-In production, you'll want to create tokens with either
+For production usage, you'll almost always want to create tokens with either
 [stripe.js](https://stripe.com/docs/tutorials/forms) or
 [Checkout](https://stripe.com/docs/tutorials/checkout), but it can be useful
 to create tokens with Stripe.net for testing.
+
+You generally wouldn't want to use stripe.net to create tokens in production,
+since creating tokens with your server offers almost no security or compliance
+benefits - it still involves passing raw card data through your server.
+If you're OK with the additional compliance burden, it's usually still simpler
+to pass card data directly to the API.
+However, there are occasionally situations where it would make sense to create
+tokens on your server.
 
 	var myToken = new StripeTokenCreateOptions();
 
@@ -184,12 +192,6 @@ to create tokens with Stripe.net for testing.
 
 	var tokenService = new StripeTokenService();
 	StripeToken stripeToken = tokenService.Create(myToken);
-
-You should never create tokens with Stripe.net with real credit cards, since
-giving your server access to unredacted card data greatly increases your
-PCI-DSS compliance burden.
-(If the extra compliance work isn't a problem, then it's still easier to not
-use tokens at all.)
 
 ### Retrieving a token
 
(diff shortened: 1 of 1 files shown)


jaymedavis:
This is closer, but not there. See my comment on the last PR.

praboud:
@jaymedavis: could you be more specific? I'm not seeing the disconnect.

readme.md
```diff
@@ -151,8 +151,13 @@ Tokens
...
 
-A token can be used anywhere on Stripe where you would normally pass a card. Once it's created, it can be used on a
-customer or a charge, but only used once.
+A token can be used anywhere on Stripe where you would normally pass a card.
+Once it's created, it can be used on a customer or a charge, but only used once.
+
+In production, you'll want to create tokens with either
+[stripe.js](https://stripe.com/docs/tutorials/forms) or
+[Checkout](https://stripe.com/docs/tutorials/checkout), but it can be useful
+to create tokens with Stripe.net for testing.
```
jaymedavis: there are situations where a PCI compliant server might want a token on a checkout page, then might want to use that token on another page. If users decide to not use the two libraries you mentioned above because they are compliant, now we might be recommending they store credit cards in their system. Either session state, or worse, database.

readme.md
```diff
@@ -180,7 +185,11 @@ customer or a charge, but only used once.
 	var tokenService = new StripeTokenService();
 	StripeToken stripeToken = tokenService.Create(myToken);
 
-Tokens are very useful if you don't want to store the customers credit card information on your servers (pci compliance).
+You should never create tokens with Stripe.net with real credit cards, since
+giving your server access to unredacted card data greatly increases your
+PCI-DSS compliance burden.
```
jaymedavis: this line is fine to mention, but the blanket statement of never isn't necessarily better or easier. See previous comment about temporarily storing credit cards.

readme.md
```diff
@@ -580,6 +589,9 @@ Disputes
 Recipients
 ----------
 
+**Note: recipients have been deprecated by Stripe - please use
+[Stripe Connnect](https://stripe.com/docs/connect) instead**
```
jaymedavis: no problem here.

jaymedavis:
left some comments I hope help. I'll have to revisit this later.

readme.md
praboud: I'm fine with getting rid of the "never" wording. However, I'd lean towards targeting this documentation at users that aren't familiar with PCI-DSS, and try to steer them away from common pitfalls. My assumption is that if it's actually your intention to do a SAQ-D, you're fairly familiar with how this stuff works, and need less guidance on how to set up your system in a compliant way.

(As an interesting microcosm of this problem, until recently, the top google result for "stripe.net tutorial" suggested creating tokens through stripe.net. An edit was later made to the tutorial warning readers about the compliance implications of this, and suggesting stripe.js/checkout. https://cmatskas.com/processing-payments-on-your-site-using-stripe-and-net/)

I'm leaning towards some wording along the lines of "from a security / compliance perspective, server-side tokenization has almost no advantages over directly passing unredacted card details to the API."

readme.md
praboud: Related to my other comment, I'd be ok to soften the wording here somewhat, but I still think it should be made clear that the vast majority of users will want to create tokens with stripe.js / checkout.

jaymedavis: 100% agree with using client side whenever possible. But the wording "can be useful for testing" should be changed to "can be useful when you are PCI compliant, and understand the implications."

readme.md
jaymedavis: My assumption was the same, probably to a fault given the article you just posted (I did not see that). And my intention was for it never to come across that you were compliant simply by generating a token on your server. This is something in the works for Stripe.net to make this process easier (interacting with client side). Given your explanation of why this is a problem now, I'm good with your new wording here.

praboud:
Ok, I've tried to make the wording a little bit less absolute, while still pushing people towards client-side tokenization.

jaymedavis:
:+1:

Issue #100, opened 2013-10-16
Testing - Charge


rravenhill:
Hi, Demo shows code to create a charge. Testing using sample as is results in the following exception.

{"Invalid string: {\"number\"=>\"424242******4242\", \"exp_month\"=>\"10\", \"exp_year\"=>\"2016\"}"}

Any idea why?

Thanks
Richard

StefanMilne:
Thanks jcharters.... I fell into this trap too.

jaymedavis:
reopening this to get this in the readme. To be clear, on charges, you can specify card or customer, but not both - I could swear this used to be in the readme, i must have goofed up and removed it at some point.

StefanMilne:
Hi sorry Jayme, didn't realise me commenting automatically reopened this. I am new to Stripe and since the new release of Stripe.net my testing ground was tweaked a little and I am trying to get to grips with it. I am working on a project which supports subscriptions and purchases(hardware) with the 2 being inclusive or exclusive. My logic is to create the subscription first (if needed) then use the stripecustomerid for the charge again if needed. If the user only purchases lets say hardware... I need to create a stripe customer, add a card and then create the charge... Is that correct or is there a better more efficient way?

jaymedavis:
No worries, you didn't actually re-open it, I did. :) ... I wanted to make sure I got this in the readme. 

That sounds fine to me. The only thing I would be aware of, is you might want to use the stripe javascript library to pass over the card details and then get a token back. That will make sure you are PCI compliant (not passing credit cards through your servers).

StefanMilne:
hmm interesting - thanks for that

PR #430, opened 2016-01-19
Delete raw card details in card examples and add comment

jxu577:
review? @jaymedavis 
cc @brandur 

The code examples in the docs pass both a token and the raw card details. A lot of users default to the raw card details and don't even attempt to read about Stripe.js

This PR edits the docs to focus on card tokens and hide the ability to pass raw card details, adding a comment that if you're PCI compliant on your end and want to send raw card details you can.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -215,23 +215,10 @@ a credit card or token, and various meta data.
 	// setting up the card
 	myCustomer.Source = new StripeSourceOptions()
 	{
-		// set this property if using a token
-		TokenId = *tokenId*,
-
-		// set these properties if passing full card details (do not
-		// set these properties if you set TokenId)
-		Object = "card",
-		Number = "4242424242424242",
-		ExpirationYear = "2022",
-		ExpirationMonth = "10",
-		AddressCountry = "US",                // optional
-		AddressLine1 = "24 Beef Flank St",    // optional
-		AddressLine2 = "Apt 24",              // optional
-		AddressCity = "Biggie Smalls",        // optional
-		AddressState = "NC",                  // optional
-		AddressZip = "27617",                 // optional
-		Name = "Joe Meatballs",               // optional
-		Cvc = "1223"                          // optional
+		// a tokenId is a symbolic representation of a card acquired by exchanging credit card information through Stripe.JS or Checkout
+		// if you're fully PCI compliant, you can also pass raw credit card information
+		// such as in https://github.com/jaymedavis/stripe.net#creating-a-card, but this is not recommended
+		TokenId = *tokenId*
 	};
 
 	myCustomer.PlanId = *planId*;                          // only if you have a plan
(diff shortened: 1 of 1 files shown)


brandur:
@jaymedavis Just for a little additional background here: we've noticed on a few inbound support requests that users on the .NET bindings occasionally are posting certain forms with credit card details when a tokenization method would be strongly preferred. Our hypothesis is that it's because the credit card usage in these code samples is so conspicuous and some users are probably just copy/pasting wholesale.

We'd recommend moving the CC usage to somewhere a little less front and center so subtly discourage its use. This pull pitches total removal, but I think there are a number of reasonable alternatives to that as well.

Thanks for taking a look!

/cc @remistr

readme.md
```diff
@@ -638,22 +569,9 @@ Recipients
 	// if you need this...
 	myRecipient.Card = new StripeCreditCardOptions()
 	{
-		// set this property if using a token
-		TokenId = *tokenId*,
-
-		// set these properties if passing full card details (do not
-		// set these properties if you set TokenId)
-		Number = "4242424242424242",
```
brandur: I wonder if it might be worth moving the credit card method to a subsection somewhere just so that if a user really wants to use it, they'll have somewhere that they can reference to see how it's done.

e.g. "A number of Stripe endpoints can take a credit card as an alternative to a token, but you should use this feature sparingly. An example of this usage might look like the following:

``` c#
myRecipient.Card = new StripeCreditCardOptions()
{
    Number = "4242424242424242",
    ...
}
```

readme.md
```diff
@@ -638,22 +569,9 @@ Recipients
...
-		AddressLine1 = "24 Beef Flank St",    // optional
-		AddressLine2 = "Apt 24",              // optional
-		AddressCity = "Biggie Smalls",        // optional
-		AddressState = "NC",                  // optional
-		AddressZip = "27617",                 // optional
-		Name = "Joe Meatballs",               // optional
-		Cvc = "1223"                          // optional
+		// if you are PCI compliant, you can use raw credit card data
+		// instead of a tokenID such as in [this example](https://github.com/jaymedavis/stripe.net#creating-a-token)
+		TokenId = *tokenId*
```
brandur: TBH I totally misread this the first time around. I'd suggest a slightly wordier alternative like:

> A token identifier is acquired by exchanging credit card information through Stripe.JS or Checkout and is a symbolic representation of that card. See https://github.com/jaymedavis/stripe.net#creating-a-token for more details. If you're testing or are sure that you're otherwise fully PCI compliant, you can also consider passing raw credit card information as in <link to section on using raw CC>, but this is generally not recommended.

brandur:
@jxu577 Nice job!

jaymedavis:
i will take a look at this in more detail, as this seems to be an ongoing issue. If you look at the bottom of #100, I explained the proper way to handle this and it seemed 'interesting' to the user. I have a feeling this is more rampant in Stripe.net implementations than expected. Recently another PR was closed that made this wording more obvious @ #425.

I'm not sure what the correct solution is for this now, but this PR contradicts the new wording from #425 (it's probably easier to pass credit card details directly if your server is PCI compliant). Some additional PCI compliance documentation was merged from there you can review. I think your point for adding this was valid in context, but I don't think removing this is the right solution or will fix the underlying problem.

Stripe.net now has
1) A full page video explaining how to properly handle credit cards at the top of the readme
2) Detailed notes under tokens for PCI compliance

Whatever the solution to this problem is, it will involve more code, forcing warnings in the library, or something else altogether to alleviate the issue.

jaymedavis:
The work was merged on #426, my mistake. The predecessor was #425. 

To be clear, I don't think your approach is unreasonable, I just think it won't solve the problem and something more heavy handed is needed. :)

PR #431, opened 2016-01-19
Update StripeAccountVerification.cs

JSchmalenberg:
Make DueBy nullable


jaymedavis:
Thanks - the documentation wasn't clear on if this was nullable or not, but I see from the example response in Stripe that it can be nullable.

PR #434, opened 2016-01-31
Fixed an issue with listing refunds

willous:


jaymedavis:
checkout the `when_listing_refunds.cs` test. This test passes. Can you look at that code and see if perhaps you were invoking it wrong? Or can you explain the issue more?

willous:
I was trying to list refunds for all charges; rather than just listing the refunds against a charge - see https://stripe.com/docs/api#list_refunds.

jaymedavis:
I see - charge used to be required, that would have been the only reason I wrote it that way (required parameters are directly passed, optionals are on options - generally speaking).

Thanks, I'll get this fixed.

willous:
You're right - it has been updated since your implementation. The other calls have also changed slightly, with "v1/charges" being replaced with "v1/refunds" on all the calls. Although, https://stripe.com/docs/api/curl#create_refund, has a mismatch between the definition and example!

POST https://api.stripe.com/v1/charges/{CHARGE_ID}/refunds

curl https://api.stripe.com/v1/refunds \
   -u <key>: \
   -d charge=ch_....

jaymedavis:
The documentation from Stripe can be misleading for sure.

Regarding this work - on master, I have implemented async tokens so this will need to change to support that (pretty easy). Also, how would you feel about making this an overload for now so I can just deprecate the existing method? In 7.0, I will kill the duplication, but if possible I'd like to release this without having to wait.

PR #436, opened 2016-02-02
Update StripeLegalEntity.cs

JSchmalenberg:
Changed AdditionalOwners from a StripeList to a List and added an renamed Strip to Stripe


jaymedavis:
doh'. thanks :)

PR #438, opened 2016-02-02
Create customer with plan - add support for "now" in trial end date

brianweet:
Same problem as issue https://github.com/jaymedavis/stripe.net/issues/91
Like 'Thwaitesy' mentioned on 9 Oct 2015, creating a customer should also support the trial_end "now" value. This way we can create the customer with one request instead of two (one for the customer and one for the subscription).

### DIFF
--- a/src/Stripe.Tests/Stripe.Tests.csproj
+++ b/src/Stripe.Tests/Stripe.Tests.csproj
@@ -56,6 +56,7 @@
     <Reference Include="System.configuration" />
     <Reference Include="System.Core" />
     <Reference Include="System.Web" />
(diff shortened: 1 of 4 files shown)


jaymedavis:
looks good - this probably needs to be fixed on customer update as well. :)

brianweet:
Not really, you will have to update the subscription itself. You have already implemented trialend for subscriptions :)

jaymedavis:
Oh sheez... derp. Thanks again - this looks great.

jaymedavis:
great job on this one. it's done - will checkin at some point today for a new release.

Issue #457, opened 2016-03-16
Stripe Connect - Account Verification: Missing minimum verification requirement fields


brakok:
Hi, I'm currently trying to create a managed account with Stripe .net and I can't find how to add the minimum verification requirements to create an account.

I see that there's an AccountVerification property on StripeAccount with some missing fields, but when creating it with the StripeAccountService and with the StripeAccountCreateOptions, I can't find a way to map the required fields to the missing required fields to have a valid account (nor with the StripeAccountUpdateOptions).

Missing required fields:

legal_entity.dob.day
legal_entity.dob.month
legal_entity.dob.year
legal_entity.first_name
legal_entity.last_name
legal_entity.type
tos_acceptance.date
tos_acceptance.ip

Am I missing something or this has not yet been implemented?

jaymedavis:
Yes - there are a few properties I have not yet implemented. Checkout StripeAccountSharedOptions - it's the base class of the one you mentioned (create and update share properties on Stripe). You will see the few that are missing and need to be implemented in the comments.

Thanks for letting me know you need this so I can put it on the radar.

brakok:
Ok, I see. That's what I thought, because Stripe Connect is still young (beta). Keep up the good work and I will be watching the next versions for this.

jaymedavis:
All of this is on the way - everything except additional_owners. Expect a checkin later today. :)

PR #458, opened 2016-03-17
add legal_entity to stripe managed account

zuologio:
Regarding #457 I've implemented the code necessary in order to have a managed account properly working.

### DIFF
--- a/src/Stripe/Infrastructure/ParameterBuilder.cs
+++ b/src/Stripe/Infrastructure/ParameterBuilder.cs
@@ -77,6 +77,16 @@ namespace Stripe
                             var stripeAccountBankAccountOptions = (StripeAccountBankAccountOptions)value;
                             newUrl = ApplyNestedObjectProperties(newUrl, stripeAccountBankAccountOptions);
                         }
+                        else if (property.PropertyType == typeof(StripeAccountLegalEntityOptions))
+                        {
+                            var sripeAccountLegalEntityOptions = (StripeAccountLegalEntityOptions)value;
+                            newUrl = ApplyNestedObjectProperties(newUrl, sripeAccountLegalEntityOptions);
+                        }
+                        else if (property.PropertyType == typeof(StripeAccountTosAcceptanceOptions))
+                        {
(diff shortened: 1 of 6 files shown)


jaymedavis:
There is still a lot left to do for managed accounts, this is only scratching the surface for the available properties. With that said - it is a good start and your implementation looks very simple.

I would like to at least add the other simple string properties to this, and leave out the harder items (like additional owners), or maybe include some simple stuff like business name. If you have some more time, I would definitely appreciate the help to add some more. :)

If not, thanks for the head start. How pressing is this for you?

src/Stripe/Infrastructure/ParameterBuilder.cs
```diff
@@ -77,6 +77,16 @@ public static string ApplyAllParameters(this StripeService service, object obj,
...
+                        else if (property.PropertyType == typeof(StripeAccountLegalEntityOptions))
+                        {
+                            var sripeAccountLegalEntityOptions = (StripeAccountLegalEntityOptions)value;
+                            newUrl = ApplyNestedObjectProperties(newUrl, sripeAccountLegalEntityOptions);
+                        }
+                        else if (property.PropertyType == typeof(StripeAccountTosAcceptanceOptions))
+                        {
+                            var stripeAccountTosAcceptanceOptions = (StripeAccountTosAcceptanceOptions)value;
+                            newUrl = ApplyNestedObjectProperties(newUrl, stripeAccountTosAcceptanceOptions);
+                        }
```
jaymedavis: you did this correctly, but I need to make a separate class with a list of types or something instead of this madness. adding this note for myself. :)

zuologio:
I agree with you when you say that there is a lot more work to do for managed accounts. The stuff I've done is the bare minimum in order to create a verified account (with all the fields marked as strictly necessary). This implementation is a critical thing for my company therefore I need it. I hope to find the time during the next few days to try to complete the work as you suggested. Thank you!

jaymedavis:
this is done - I went through and added all properties except additional_owners (that will take more thought). I'll get this released today sometime. :)

Issue #342, opened 2015-05-22
Dispute Structured Evidence


toddca:
How do you submit structured evidence for a dispute?

ref: https://stripe.com/docs/guides/dispute-evidence

I can see there is a dispute update method that accepts a single string however the example just shows a simple message/comment type of string being passed. Am I just missing this somehow?

toddca:
Also looking for a way to upload dispute information via file uploads. Cannot seem to find that in this API either.

jaymedavis:
The information you linked to for dispute just needs to be added. When Stripe discount was implemented, that did not exist at the time. StripeDisputeCreateOptions or something like that probably needs to be implemented based on your link.

File uploads has not been developed yet.

jeancroy:
+1 for File uploads

If it help anyone I was able to make it works using the same HttpClient/HttpRequestMessage infrastructure as stripe.net. A lot of what is on the net use the lower level stuff.

https://gist.github.com/jeancroy/c71f8a710734a7d13248f968c778d428

jaymedavis:
Thanks for posting this, it looks less complicated than the other solutions I've seen. Ill definitely keep as a reference - thanks!

What platform are you on? I'm wondering if there's any chance that could work with mono (thinking no because of async - will probably have to roll two versions).

jeancroy:
Hi @jaymedavis 

Basically I followed your model.

Here you create a HttpRequestMessage and set the content property.
https://github.com/jaymedavis/stripe.net/blob/master/src/Stripe/Infrastructure/Requestor.cs#L120-L123


Here I create a HttpRequestMessage 
https://gist.github.com/jeancroy/c71f8a710734a7d13248f968c778d428#file-multipartupload-cs-L6

and set the content property
https://gist.github.com/jeancroy/c71f8a710734a7d13248f968c778d428#file-multipartupload-cs-L34

Main difference is that I use a fancy MultipartFormDataContent instead of your StringContent
https://gist.github.com/jeancroy/c71f8a710734a7d13248f968c778d428#file-multipartupload-cs-L9

---

So to answer your question I use a standard asp.net 4 mvc 5 application.
Most of the async stuff should be similar to yours

The special part may be in MultipartFormDataContent and there seem to have a mono implementation here
https://github.com/mono/mono/blob/master/mcs/class/System.Net.Http/System.Net.Http/MultipartFormDataContent.cs

toddca:
I submitted a pull request for this back in March.

[...]

PR #469, opened 2016-03-28
Improvements for #342 To Add Dispute Structured Evidence

toddca:
This change extends the StripeDisputeService by adding an Update() method that allows one to submit evidence to a dispute given a dispute id, it also adds a List() method which allows for the enumeration of dispute objects. This change also allows for uploading of File based evidence.

### DIFF
--- a/src/Stripe/Infrastructure/ParameterBuilder.cs
+++ b/src/Stripe/Infrastructure/ParameterBuilder.cs
@@ -77,6 +77,10 @@ namespace Stripe
                             var stripeAccountBankAccountOptions = (StripeAccountBankAccountOptions)value;
                             newUrl = ApplyNestedObjectProperties(newUrl, stripeAccountBankAccountOptions);
                         }
+                        else if (obj.GetType() == typeof(StripeEvidence))
+                        {
+                            newUrl = ApplyParameterToUrl(newUrl, "evidence[" + attribute.PropertyName + "]", value.ToString());
+                        }
                         else
                         {
                             newUrl = ApplyParameterToUrl(newUrl, attribute.PropertyName, value.ToString());
@@ -135,5 +139,6 @@ namespace Stripe
 
             return newUrl;
         }
+        
     }
 }
--- a/src/Stripe/Infrastructure/Requestor.cs
+++ b/src/Stripe/Infrastructure/Requestor.cs
@@ -1,6 +1,8 @@
 ﻿using System;
+using System.IO;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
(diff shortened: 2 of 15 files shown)


jaymedavis:
thanks!

src/Stripe/Infrastructure/ParameterBuilder.cs
```diff
@@ -77,6 +77,10 @@ public static string ApplyAllParameters(this StripeService service, object obj,
                             var stripeAccountBankAccountOptions = (StripeAccountBankAccountOptions)value;
                             newUrl = ApplyNestedObjectProperties(newUrl, stripeAccountBankAccountOptions);
                         }
+                        else if (obj.GetType() == typeof(StripeEvidence))
```
jaymedavis: Fortunately, this isn't necessary anymore. Nested options can just be marked with INestedOption instead and it magically works.

src/Stripe/Infrastructure/Requestor.cs
```diff
@@ -29,6 +31,57 @@ public static string PostString(string url, StripeRequestOptions requestOptions)
             return ExecuteRequest(wr);
         }
```
jaymedavis: I had posted from another issue (I believe you commented on), that looked like an easier (less code way) to do the file upload. I'm not sure if it would work the same, but it would be worth investigating to see if we can reduce some code.

jaymedavis:
This is a lot of fantastic work - thanks again! Sorry it's taken me awhile to get to this.

toddca:
Yea it has taken you soooo long -- I want a refund on the licensing I have paid to use this great library. :+1:  :-)

jaymedavis:
You tugged at my :heart: feels. :)

Thanks, man.

jaymedavis:
I've got this started on the `#469-dispute-evidence` branch. I am copy/pasting your code heavily. 👍 

For StripeEvidence, there was a lot of expandable properties that were files/file id's... so I renamed some properties there. I also wrote some tests to start verifying everything. You can check out that branch (only 1 commit so far) and follow along if you'd like. Basically, StripeEvidence is complete and next up I'll be adding the evidence details. Creating a test that generates a dispute was enough tonight - I'll write some more tests tomorrow to start getting the file service going.

I am curious, how did you find the right boundary to send the multipart file? I couldn't find that anywhere - I'm also referencing [this gist[(https://gist.github.com/jeancroy/c71f8a710734a7d13248f968c778d428#file-multipartupload-cs-L34) for file uploading as we discussed in the past on a previous issue. 

Thanks again for your work, it is saving me a lot of time.

PR #501, opened 2016-05-05
add properties for sourced_transfers and destination_payment

drgrieve:
Adding property sourced_transfers to balance transaction and destination_payment to transfer object.

### DIFF
--- a/src/Stripe/Entities/StripeTransfer.cs
+++ b/src/Stripe/Entities/StripeTransfer.cs
@@ -58,6 +58,12 @@ namespace Stripe
         [JsonProperty("description")]
         public string Description { get; set; }
 
+        [JsonProperty("destination")]
+        public string Destination { get; set; }
+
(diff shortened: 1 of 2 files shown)


jaymedavis:
looks good. just need to verify against docs and/or find out why it isn't in them :)

src/Stripe/Entities/StripeTransfer.cs
```diff
@@ -58,6 +58,12 @@ internal object InternalBalanceTransaction
         [JsonProperty("description")]
         public string Description { get; set; }
 
+        [JsonProperty("destination")]
+        public string Destination { get; set; }
```
jaymedavis: This is an expandable property - so the service needs to account for that (with a bool).

jaymedavis:
bah - the expandable property isn't playing nice, this is good for now. Thanks @drgrieve

PR #505, opened 2016-05-12
Update README to use Github style code blocks

khiemtong:
* Adds Github style code blocks for some syntax highlighting
* Found myself mostly referring to the README when looking through the library and thought some syntax highlighting would be helpful


jaymedavis:
VERY nice. Thanks!!

PR #507, opened 2016-05-17
Support for pay_immediately during customer create

JensRasmussen:
Mail correspondence:

A PR or an issue would be great. This is a good question. Typically when things aren't documented, Stripe's stance is 'it's not guaranteed', which is why I avoid it. If you add this email verbatim as a GitHub PR, that will allow me to show them exactly what you are proposing and I can discuss it with them. :)

Jayme

On May 17, 2016, at 3:21 PM, Jens Rasmussen <jlr@cvation.com> wrote:
Hi Jayme
 
I was not sure whether to create an Issue or not (more like a change request).
 
To integrate Stripe with Avalara (handling of taxes – see https://www.avalara.com), Avalara needs to update Stripe invoices before they are paid. This is not immediately possible when creating a Stripe customer with a plan, since Stripe will try to have the first invoice paid as part of the verification process.
 
Stripe does support the scenario if one adds a pay_immediately=false flag to the customer create request (see e.g. https://help.avalara.com/kb/001/Why_are_my_subscription_invoices_in_Stripe_not_calculating_tax_or_posting_to_the_admin_console%3F).
Unfortunately it is not documented by Stripe, so I am not surprised that String.net does not support setting the flag.
 
I have forked Stripe.net and added a bool? PayImmediately property in the StripeCustomerCreateOptions class.
 
If you are interested, I can make a Pull request?
 
Best Regards
Jens

### DIFF
--- a/src/Stripe.Tests/Stripe.Tests.csproj
+++ b/src/Stripe.Tests/Stripe.Tests.csproj
@@ -86,6 +86,8 @@
     <Compile Include="bank_accounts\when_creating_a_bank_account.cs" />
     <Compile Include="card\when_creating_a_card.cs" />
     <Compile Include="charges\when_refunding_and_setting_fraud_details.cs" />
+    <Compile Include="customers\when_creating_a_customer_with_pay_immediately_false.cs" />
+    <Compile Include="customers\when_creating_a_customer_with_pay_immediately_true.cs" />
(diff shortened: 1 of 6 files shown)


jaymedavis:
according to stripe:
` just found some more about that flag -- it's undocumented and part of a legacy feature that's been deprecated`

PR #515, opened 2016-05-27
Update subscriptions to not need customerId

### LINKED ISSUE #514: SubscriptionService Methods - without CustomerId
According to the Stripe API documentation, to do pretty much anything with existing subscriptions, you only need the SubscriptionId. (See https://stripe.com/docs/api/curl#retrieve_subscription) Yet this library requries the customerId.

The method seems to be a leftover from when subscriptions were part of the customer and the way they were retrieved was from the the customer endpoints (https://github.com/jaymedavis/stripe.net/blob/f467706bd0f61c359e4bff09d98cc38402edbda9/src/Stripe/Services/Customers/StripeCustomerService.cs#L63)

The CustomerId can probably be removed from the subscription methods which do not use it. I'd suggest adding an overload which doesn't include the customerId and deprecating the ones which include it but do not need/use it.

TheSharpieOne:
Resolves #514
Use stripe subscriptions endpoints directly (no customer)
Deprecate subscription methods using customerId
Add method to list all subscriptions
Update readme to reflect signature changes

Disclaimer: I am not a .NET developer... but these changes seemed simple enough to do. I may be wrong; review the changes carefully.

### DIFF
--- a/src/Stripe/Services/Subscriptions/StripeSubscriptionService.cs
+++ b/src/Stripe/Services/Subscriptions/StripeSubscriptionService.cs
@@ -1,4 +1,5 @@
 ﻿using Newtonsoft.Json;
+using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 
@@ -10,9 +11,15 @@ namespace Stripe
 
         public bool ExpandCustomer { get; set; }
 
+        [Obsolete("Get with customerId is deprecated, use Get without the customerId.")]
         public virtual StripeSubscription Get(string customerId, string subscriptionId, StripeRequestOptions requestOptions = null)
         {
-            var url = string.Format(Urls.Subscriptions + "/{1}", customerId, subscriptionId);
+            return Get(subscriptionId, requestOptions);
+        }
+
+        public virtual StripeSubscription Get(string subscriptionId, StripeRequestOptions requestOptions = null)
+        {
+            var url = string.Format(Urls.Subscriptions + "/{0}", subscriptionId);
 
             return Mapper<StripeSubscription>.MapFromJson(
                 Requestor.GetString(this.ApplyAllParameters(null, url, false),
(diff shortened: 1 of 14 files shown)


src/Stripe/Services/Subscriptions/StripeSubscriptionService.cs
```diff
@@ -10,40 +11,66 @@ public class StripeSubscriptionService : StripeService
...
                 SetupRequestOptions(requestOptions))
             );
         }
 
+        [Obsolete("Create with customerId and planId is deprecated, use Create without the customerId planId by supplying those values in the StripeSubscriptionCreateOptions.")]
         public virtual StripeSubscription Create(string customerId, string planId, StripeSubscriptionCreateOptions createOptions = null, StripeRequestOptions requestOptions = null)
         {
-            var url = string.Format(Urls.Subscriptions, customerId);
-            url = this.ApplyAllParameters(createOptions, url, false);
+            createOptions.CustomerId = customerId;
```
jaymedavis: you could possibly get a null reference here if they don't pass the create options -  it simply needs a null check first.

src/Stripe/Services/Subscriptions/StripeSubscriptionService.cs
```diff
@@ -10,40 +11,66 @@ public class StripeSubscriptionService : StripeService
...
+            createOptions.PlanId = planId;
```
jaymedavis: same here for null check

src/Stripe/Services/Subscriptions/StripeSubscriptionService.cs
```diff
@@ -10,40 +11,66 @@ public class StripeSubscriptionService : StripeService
...
+
+            return Create(createOptions, requestOptions);
```
jaymedavis: very clever deprecation here :+1:

TheSharpieOne:
I added the null checks and rebased so there are currently no conflicts. Everything should be good to go.

My disclaimer above still stands 😄

jaymedavis:
this will be managed from #726

PR #522, opened 2016-06-03
Overriding the endpoints and the security protocols

sunsided:
Hi, as mentioned in https://github.com/jaymedavis/stripe.net/issues/500#issuecomment-219142516 I added some minor fixes to allow for using different endpoints instead of the original Stripe ones, as well as changing the supported security protocols, e.g. disabling TLSv1.2.

We are running our applications using the `dotnet` CLI (preview1) tooling on mono. Sadly, mono does not support TLS 1.2 at all, so we are using a simple nginx forward proxy as a middle man that does speak plain HTTP on the inside and wraps the actual request to Stripe in TLS. In order to achieve that, we have to point all Stripe related requests to that proxy and also - optionally - turn off TLS 1.2. The last step might be optional.

This is the same with the PayPal API by the way, for which the SDK already supports setting of arbitrary endpoints via a configuration dictionary.

### DIFF
--- a/src/Stripe/Infrastructure/Requestor.cs
+++ b/src/Stripe/Infrastructure/Requestor.cs
@@ -7,10 +7,22 @@ using System.Threading.Tasks;
 
 namespace Stripe
 {
-    internal static class Requestor
+    public static class Requestor
     {
         internal static HttpClient HttpClient { get; private set; }
         internal static string Version { get; }
+
+#if !PORTABLE
+
+        /// <summary>
+        /// Gets or sets the type of the security protocol.
+        /// </summary>
+        /// <value>
+        /// The type of the security protocol.
+        /// </value>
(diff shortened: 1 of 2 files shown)


jaymedavis:
I'm not sure why the url's were made public in the library, as this is something I do not want to expose. Regarding 1.2 - this is a complicated matter. Stripe posted a blog/article recently stating they were moving towards this, and that's all that will be supported in the future. I'm not sure what this means... but this isn't something I can merge in right now. This request is specifically from them to do this. I would contact them and let them know the issue - or drop in IRC for a chat with them. I would prefer the latter (I am truepudding in IRC), because I plan on releasing to Xamarin shortly - and this will be an issue.

sunsided:
I dislike the `public` hack on the `Requestor` as well, but the endpoint configuration override actually saved our lives with the PayPal SDK. (And here as well). 
Would an options based approach be more okay with you - say, `StripeOptions` that go into a constructor - or do you really want to hide the endpoints?

PR #525, opened 2016-06-09
#368  Adding support for listing all invoice line items

daniel-cohen:
See: https://stripe.com/docs/api/curl#invoice_lines

### DIFF
--- a/src/Stripe/Services/Invoices/StripeInvoiceService.cs
+++ b/src/Stripe/Services/Invoices/StripeInvoiceService.cs
@@ -53,6 +53,14 @@ namespace Stripe
             );
         }
 
+        public virtual IEnumerable<StripeInvoiceLineItem> ListLineItems(string invoiceId, StripeInvoiceListLineItemsOptions listOptions = null, StripeRequestOptions requestOptions = null)
+        {
+            return Mapper<StripeInvoiceLineItem>.MapCollectionFromJson(
+                Requestor.GetString(this.ApplyAllParameters(listOptions, $"{Urls.Invoices}/{invoiceId}/lines", true),
+                SetupRequestOptions(requestOptions))
+            );
+        }
+
         public virtual StripeInvoice Create(string customerId, StripeInvoiceCreateOptions createOptions = null, StripeRequestOptions requestOptions = null)
         {
             var url = ParameterBuilder.ApplyParameterToUrl(Urls.Invoices, "customer", customerId);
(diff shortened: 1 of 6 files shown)


jaymedavis:
looks like async still missing? You can just push to your fork - you don't have to close and reopen. I can merge on my end

daniel-cohen:
@jaymedavis. Check it out. I don't have time to test it, But I think that's what you're after.

src/Stripe/Services/Invoices/StripeInvoiceService.cs
```diff
@@ -53,6 +53,14 @@ public virtual IEnumerable<StripeInvoice> List(StripeInvoiceListOptions listOpti
             );
         }
 
+        public virtual IEnumerable<StripeInvoiceLineItem> ListLineItems(string invoiceId, StripeInvoiceListLineItemsOptions listOptions = null, StripeRequestOptions requestOptions = null)
```
jaymedavis: this should just be called List, to match the rest of the library. And these two methods should be moved to a new InvoiceLineItemService - this follows their documentation exactly. If it doesn't - trust me - I get many emails. ;)

src/Stripe/Services/Invoices/StripeInvoiceService.cs
```diff
@@ -106,6 +114,14 @@ public virtual StripeInvoice Create(string customerId, StripeInvoiceCreateOption
             );
         }
 
+        public virtual async Task<IEnumerable<StripeInvoiceLineItem>> ListLineItemsAsync(string invoiceId, StripeInvoiceListLineItemsOptions listOptions = null, StripeRequestOptions requestOptions = null)
```
jaymedavis: same here for List

jaymedavis:
Actually, there is an InvoiceItemService that appears to handle all of this... let me know if that doesn't meet your needs or I am missing something.

jaymedavis:
Scratch that, I found it in the docs. Sorry about that. :)

daniel-cohen:
It's not the same. This is to list all the line items on a specific invoice. since it's limited to 10 or 11 by default when you get the invoice itself.

jaymedavis:
Man - I was totally confused on this PR, disregard my comments. This makes sense now - the invoice line items is just confusing to think about, but I see what you are doing. Thanks!

jaymedavis:
This is on the invoice-line-items branch. I only made a couple small tweaks to the test, and added the subscription plan id to the options. Excellent work, will be in the next release.

daniel-cohen:
@jaymedavis Thanks for merging this !!

Issue #524, opened 2016-06-08
proration_date missing from StripeSubscriptionUpdateOptions


mixtmeta:
Stripe allows for proration based on a specified date. This option does not appear to be available in the StripeSubscriptionUpdateOptions. Would be great if it was added.

jaymedavis:
Thanks for reporting this.

jaymedavis:
will manage this from #526 as the work is done there - thanks

jaymedavis:
sorry about the annoyances with labels here - I'm using waffle.io/jaymedavis/stripe.net and accidentally grabbed the wrong card.

PR #526, opened 2016-06-10
Added Proration Date to StripeSubscriptionUpdateOptions

mixtmeta:
For #524

### DIFF
--- a/src/Stripe/Services/Subscriptions/StripeSubscriptionUpdateOptions.cs
+++ b/src/Stripe/Services/Subscriptions/StripeSubscriptionUpdateOptions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using Newtonsoft.Json;
 using System.Collections.Generic;
+using Stripe.Infrastructure;
 
 namespace Stripe
 {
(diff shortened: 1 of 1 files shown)


jaymedavis:
can you tell me your method for testing this? I'm working on a test for this in stripe.net - thanks.

jaymedavis:
I talked to Stripe about this, because I was unable to see anything working from testing. So how prorate works, is if you just set prorate or the proration date, it will actually not do anything at all. Something else has to change. I am going to get the other properties added for subscription updates so this wont happen, but it's something to be aware of.

jaymedavis:
this is complete. It was a bit tricky to test due to how the proration works, but it will be in the next release - thanks much - your code worked perfectly.

mixtmeta:
Thanks. To test I compiled a local version of the project with the change in it and dumped it into a project I was working on that required proration. Then I checked the stripe log via the dashboard to make sure the proration_date field was received as part of my request.

joshpierce:
@jaymedavis do you have an estimate on when the next release will be that would include this?

We're using stripe.net (and <3 it) and are looking to use this feature as we're currently using the prorate date on our generation of upcoming invoice for previewing a subscription change.

Cheers!

jaymedavis:
Hi @joshpierce - sometime over the next few days I hope to get a release out. There are a couple more things I'm looking at - including a recent PR to get upcoming invoice lines (with pagination).

PR #530, opened 2016-06-22
Adding support for paginated queries of pending invoice line items

circuitrider:


jaymedavis:
this looks pretty awesome @circuitrider. A couple of nits - 1) I'm no longer starting anything with the name Stripe - it is a little too verbose and it was a mistake for even me.   2) This will need to support async as well. Test are nice to have, but if you swear to talk all public flogging, I might let it go. :)

circuitrider:
Hah. I totally meant to but forgot to do async. Mahbad. Will fix and
resubmit in a bit after I get back from dinner.

PR #563, opened 2016-08-19
Fix to targetFramework value for PCL Target Framework in nuspec

duckwaffle:
The targetFramework value for the PCL in the nuspec file is not a valid value. This leads to the nuget package not being able to be consumed for PCL projects.

https://docs.nuget.org/create/targetframeworks


jaymedavis:
Thanks @duckwaffle. wp8 is still also supported, I think we just need to add win8.

duckwaffle:
@jaymedavis  According to the nuget reference document that will change the supported PCL targetframework to profile259. Which also works, as I believe that is the standard profile for Xamarin PCL's!

jaymedavis:
The goal was Profile111 - which it seems you have the correct signature here. I think this was probably just my mistake. 

Cc @flagbug

PR #568, opened 2016-08-23
Add application_fee to transfer objects fixes #552

### LINKED ISSUE #552: Add Transfer Fee to StripeTransfer
It seems like there is quite a bit missing from Transfer according to the api.   I am particularly in need of the application fee.  Now that recipients have been depreciated it also seems to be missing the destination parameter.

dplarina:


src/Stripe.Tests/transfers/test_data/stripe_transfer_create_options.cs
```diff
@@ -18,7 +18,8 @@ public static StripeTransferCreateOptions Valid(string recipient = "self")
                 {
                     { "A", "Value-A" },
                     { "B", "Value-B" }
-                }
+                },
+                ApplicationFee = 10
```
jaymedavis: can we assert this on a test that it is on the StripeTransfer?

jaymedavis: Alternatively, you could just copy/paste a test as a base, then get a Valid object, and set the application fee in that test, then assert in that test. Make sense?

jaymedavis:
You can add the refund application fee if you want here too - based on the link you pasted in the other issue. This is great though - thanks!

src/Stripe.Tests/transfers/test_data/stripe_transfer_create_options.cs
dplarina: I'm not really familiar with the testing framework you're using, but I'll take a stab at it.

src/Stripe.Tests/transfers/when_creating_a_transfer_with_application_fee.cs
```diff
@@ -0,0 +1,29 @@
+﻿// ***** This test needs a bank account setup in order to run correctly, and the ignore attribute wasn't working. Good times.
```
jaymedavis: Ha! I feel ya on the universal "Good times" frustration - I recently created a TestSources class that will give you valid bank account options. You can just create the bank account service and pass that. Also, for testing just the application fee, you don't need to worry about it having the behaviors (the protected types at the top and the behaves_like call). All that can be skipped - we just need to assert this one thing. I've been moving away from behaviors lately, they seem to confuse people - confirm? :)

Thanks for your work - this is great.

dplarina:
Added the test for transfers with application fee. Let me know if that works for you.

src/Stripe.Tests/transfers/when_creating_a_transfer_with_application_fee.cs
dplarina: Sorry that was copypasta from the other test. Realized I didn't rename the class. Removed behaviours. Not sure I can remove the protected types.

src/Stripe.Tests/transfers/when_creating_a_transfer_with_application_fee.cs
```diff
@@ -0,0 +1,27 @@
...
+
+//using Machine.Specifications;
+
+//namespace Stripe.Tests
+//{
+//    public class when_creating_a_transfer_with_application_fee
+//    {
+//        protected static StripeTransferCreateOptions StripeTransferCreateOptions;
+//        protected static StripeTransfer StripeTransfer;
```
jaymedavis: This can be private static - it needs to be global because it's used in the assert.

src/Stripe.Tests/transfers/when_creating_a_transfer_with_application_fee.cs
```diff
@@ -0,0 +1,27 @@
...
+
+//        private static StripeTransferService _stripeTransferService;
+
+//        Establish context = () =>
+//        {
+//            _stripeTransferService = new StripeTransferService();
+//            StripeTransferCreateOptions = test_data.stripe_transfer_create_options.Valid();
```
jaymedavis: This can just be declared inline....  var createOptions = ....  Since it's not used anywhere else. The reason for having them protected at the top is so the behavior can access them for assertions. Make sense?

jaymedavis: The only reason you would leave it global is if you changed your test below to assert that the transfer is equal to the options value (instead of hard coding 10) - then you would still make it private because the behavior is gone. Some options for you :)

dplarina: Yep I believe I understand

dplarina:
This should be good to review.

Issue #333, opened 2015-05-04
source_transaction missing from StripeTransfer


ronytesler:
I want to use the 'source_transaction' field, as described here:
https://stripe.com/docs/api?lang=ruby#create_transfer

ronytesler:
I guess it's this class - StripeTransferCreateOptions

jaymedavis:
Looks like there is a few things missing from this class according to the docs. Feel free to submit a PR or I will update it when I have the chance. Thanks!

cwoolum:
This can be closed now

jaymedavis:
Thanks!

PR #571, opened 2016-08-25
Add property for source transaction

cwoolum:
Added source_transaction property to facilitate Special Case Transfers for Stripe connect.

Related ticket #333

### DIFF
--- a/src/Stripe/Services/Transfers/StripeTransferCreateOptions.cs
+++ b/src/Stripe/Services/Transfers/StripeTransferCreateOptions.cs
@@ -26,6 +26,12 @@ namespace Stripe
         [JsonProperty("statement_descriptor")]
         public string StatementDescriptor { get; set; }
 
+        /// <summary>
+        /// You can use this parameter to transfer funds from a charge (or other transaction) before they are added to your available balance. A pending balance will transfer immediately but the funds will not become available until the original charge becomes available. See the Connect documentation for details.
+        /// </summary>
+        [JsonProperty("source_transaction")]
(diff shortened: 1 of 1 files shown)


src/Stripe/Services/Transfers/StripeTransferCreateOptions.cs
```diff
@@ -26,6 +26,12 @@ public class StripeTransferCreateOptions
         [JsonProperty("statement_descriptor")]
         public string StatementDescriptor { get; set; }
 
+        /// <summary>
+        /// You can use this parameter to transfer funds from a charge (or other transaction) before they are added to your available balance. A pending balance will transfer immediately but the funds will not become available until the original charge becomes available. See the Connect documentation for details.
+        /// </summary>
+        [JsonProperty("source_transaction")]
+        public string SourceTransaction { get; set; }
```
jaymedavis: How do you know this works?

jaymedavis:
This actually looks like it belongs on the StripeTransfer, and not the StripeTransferCreateOptions. I will add it there, this doesn't look like something you set.

cwoolum:
No, it's definitely on create. For Stripe connection, you can't have more than 10% of your transfers be withour a source transaction for compliance reasons

https://stripe.com/docs/api#create_transfer

cwoolum:
Also see https://stripe.com/docs/connect/special-case-transfers#associating-a-transfer-with-a-charge

jaymedavis:
thanks :+1:

PR #573, opened 2016-08-26
Fix xamarin.ios throwing NotImplementedException when deserializing StripeCharge

Julien-Mialon:
Hello,

Just a simple fix to avoid a NotImplementedException on iOS platforms when deserializing StripeCharge object. 
It was due to the call of method ToObject<dynamic>() on JObject because of missing code generation support in Xamarin.iOS

Thanks for your work.


jaymedavis:
Ah-ha! That makes sense, thanks!

PR #574, opened 2016-08-27
Async CancelationTokens

SharpeRAD:
Added cancelationToken's to the async service methods and corrected the async delete methods so they return a Task.

### DIFF
--- a/src/Stripe.Portable/Stripe.Portable.csproj
+++ b/src/Stripe.Portable/Stripe.Portable.csproj
@@ -112,6 +112,9 @@
     <Compile Include="..\Stripe\Entities\StripeDeclineChargeOn.cs">
       <Link>Entities\StripeDeclineChargeOn.cs</Link>
     </Compile>
+    <Compile Include="..\Stripe\Entities\StripeDeleted.cs">
+      <Link>Entities\StripeDeleted.cs</Link>
+    </Compile>
     <Compile Include="..\Stripe\Entities\StripeDiscount.cs">
       <Link>Entities\StripeDiscount.cs</Link>
     </Compile>
--- a/src/Stripe/Entities/StripeDeleted.cs
+++ b/src/Stripe/Entities/StripeDeleted.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using Newtonsoft.Json;
+using Stripe.Infrastructure;
+
+namespace Stripe
+{
+    public class StripeDeleted : StripeObject
+    {
+        [JsonProperty("deleted")]
+        public bool Deleted { get; set; }
+    }
+}
--- a/src/Stripe/Infrastructure/Requestor.cs
+++ b/src/Stripe/Infrastructure/Requestor.cs
@@ -3,6 +3,7 @@ using System.Net;
 using System.Net.Http;
 using System.Reflection;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Stripe.Infrastructure;
 
@@ -18,6 +19,9 @@ namespace Stripe
             HttpClient = new HttpClient();
         }
 
+
+
+        //Sync
         public static string GetString(string url, StripeRequestOptions requestOptions)
         {
             var wr = GetRequestMessage(url, HttpMethod.Get, requestOptions);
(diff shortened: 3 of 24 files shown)


jaymedavis:
Would you mind Obsolete'ing the existing delete methods? I'm trying to avoid breaking changes right now. Thanks for all your hard work here!

jaymedavis:
Unrelated - I was not familiar with Cake - does this work with mspec by chance? Im in desperate need of a better build process. :)

SharpeRAD:
There's no need to obsolete them as the previous methods returned void. At worst the compiler will throw a warning as clients will not have an await in the call, but the important point is no errors so your OK :+1: 

Personally I would change the existing sync methods to extensions of the async implementations that simply call the respective "method().Result" as maintaining two implementations must be painful :cry: 

I don't have any experience of mspec but you should definitely move to :cake:, I was tempted to include it in the PR but it was getting too big already :smile:

SharpeRAD:
You should also look into using [Flurl](https://github.com/tmenier/Flurl) for your client as your requests could be a lot cleaner :wink:

jaymedavis:
Thanks - didn't consider that on Delete. I'll check it out.

Agreed on sync, this is something I want to do. It never happened before only because this library is old and supports a lot of frameworks - calling .Result wasn't always a possibility. But it is now. :)

xt0rted:
I forget what the exact reasons were, but [shawnmclean/Mandrill-dotnet](https://github.com/shawnmclean/Mandrill-dotnet) removed Flurl in favor of HttpClient because of issues users were having when using it. I don't know if all of the discussion around this in the GitHub issues or not, but this is one of them https://github.com/shawnmclean/Mandrill-dotnet/issues/146. Just something to consider/look into before making a change like this.

jaymedavis:
I think my method is fine, honestly. I want to do a couple of formatting changes (remove Urls.cs, tab/space services like The BankAccount/CountrySpec service), but there isn't many things that trouble me or warrant adding dependencies. The bones of Stripe.net isn't very much code at all - it's a big library that does nothing more than build strings. If you take away the cookie cutter parts (services/entities), there is not much left. Seeing as a goal of mine is to run on the Xbox, I'm very careful about any dependencies introduced. Newtonsoft.Json has earned my trust. :)

I do appreciate comments/opinions, and I will check it out. But I'm hesitant to add a dependency unless I have to.

jaymedavis:
Btw, thank you for not including Cake in this PR. I am not a fan of PRs handling too many concerns.

SharpeRAD:
Noticed the reference issue so you've got two for one now :smile:

SharpeRAD:
@xt0rted the issue with Flurl appeared to be just a bad reference but I understand the hesitance to adding another reference, if it ain't broke!

Anyway its beyond the scope of this PR, which I'm hoping will now be accepted :wink:

jaymedavis:
Thanks for that @SharpeRad! This will definitely be accepted, I just want to pull it down and test it. Last week was plenty busy, so I will get to this next week. I'd also like to write an async section in the readme on how to use the cancellation token (and an async example for that matter).

SharpeRAD:
Can you merge this :wink:

jaymedavis:
Yes - I was a bit under the weather last week, and didn't get to everything I had hoped. If you want to add-on and example in the readme for how to use the cancellation token, that would help too.

jaymedavis:
I was reading [more about cancellation tokens] this morning. It seems the typical recommendation is for tasks that take a long time to complete, like file uploads, etc. Many of the methods that now have a token, are so fast, I'm not sure why they need a token (such as a get). I can see a token being needed for create's (or any long running tasks), and even those are not very long running. Just because every async task can use a token, I'm not sure it's necessary to use one (based on my research). Would you disagree with removing tokens from the calls like Get? Or do you actually use cancellation tokens on these calls?

SharpeRAD:
The point of modern web-dev is to async all the way and free up threads as quickly as possible, so it doesn't matter how quickly your task completes if its an async task (which api calls are) and your client or background process cancels the request then you should pass that along the chain using cancellation tokens.

SharpeRAD:
Or to simplify the argument: if the framework provides you a feature that can be added to a client library using optional parameters without breaking existing implementations and someones already written it for you why wouldn't you include it?

SharpeRAD:
As you can tell I don't like your speed argument as cancelationTokens are to cover the times when endpoints aren't responding in a timely fashion.

jaymedavis:
>If the framework provides you a feature that can be added to a client library using optional parameters, without breaking existing implementations why wouldn't you include it?

Libraries like Go exist because other libraries have fluff. If you look through my code, you will see I am not a fan of fluff. Just because something can be added, doesn't mean it should. 

> As you can tell I don't like your speed argument as cancelationTokens are to cover the times when endpoints aren't responding in a timely fashion.

This is a reasonable argument. However, I cannot see a single user implementing a "cancel" button when they are getting a subscription by ID. That is the only way this would be implemented.

jaymedavis:
[This](https://blogs.msdn.microsoft.com/andrewarnottms/2014/03/19/recommended-patterns-for-cancellationtoken/) is another good article on async. I definitely see your argument about a chain - you could be getting a subscription in the middle of a sequence, thus you need the token. This article does warn of the dangers of cancellation tokens too though - it's not like a "rollback" feature, it just stops where it's at, and could potentially leave you in a bad state. I'm not arguing against this PR entirely, I'm just trying to understand the use case.

SharpeRAD:
Its an optional parameter that completes the async implementation of the underlying HttpClient that this library is wrapping. Its not being forced on anyone, nor is it breaking the existing interface.

SharpeRAD:
> Libraries like Go exist because other libraries have fluff. If you look through my code, you will see I am not a fan of fluff. Just because something can be added, doesn't mean it should. 

So why does the library contain two implementations of every method (sync / async) when the underlying request is asynchronous?

jaymedavis:
Can you provide a real-world example of why you would use this on say... Listing Invoices?

I am aware of how the async operation works, and why this is needed. My argument is that it should be in places where it will be used, not just everywhere because it doesn't break the existing interface. I could add many optional arguments that don't break the interface, I have purposely chosen not too. I get a lot of email, and people will want to know what this is and how it works. I can answer that question on creating a charge, I cannot on listing invoices.

xt0rted:
Just a heads up, the optional parameters are a breaking change and should result in a major version bump. More info on that can be found here http://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/

jaymedavis:
Thanks @xt0rted. It appears from this article that the fix is simply to re-compile your application. Is it bad to assume users recompile when they update NuGet packages? Or is it wrong to just add a release note to tell them to recompile?

> What happens when we try to execute the client without recompiling the client application?

Thanks much for bringing this up!

SharpeRAD:
While I will not pass the cancelationToken up the chain for every call (like charging a card), I would pass it up the chain to calls like "listing invoices" as why would you wait for it to complete when you know you can have your thread back ealier?

SharpeRAD:
That said I would expect a wrapper to have a consistent interface, so id implement it on every async method and let the developer decide when to pass the token.

jaymedavis:
Can you write a code sample of listing invoices and getting your thread back earlier? I'm sorry - I'm a bit confused. As far as I know, you can set a timeout on your token, but that's just going to cancel the operation and you have no data.

jaymedavis:
> That said I would expect a wrapper to have a consistent interface, so id implement it on every async method and let the developer decide when to pass the token.

That's fair.

xt0rted:
@jaymedavis I'd say it's mostly safe to assume the application is being rebuilt when updating the NuGet package with the only real exception to that being if the application was using another library that references Stripe.net. In that scenario the other library would need to be rebuilt otherwise the consumer would throw an exception.

Not that I'd like to admit it but I've had times where I needed to drop a new version of a library in the bin folder to quickly fix an issue while working on a perminant fix.

There's probably a very small chance of those cases happening with this library but they're still possible. 

[SaasEcom.Core](https://www.nuget.org/packages/SaasEcom.Core/) has a dependency on Stripe.net, though it's an old one so not an issue here.

jaymedavis:
Appreciate it @xt0rted. There are a few others that have dependencies as well, not to mention someone could have wrapped this them self in a facade or behind their business rules. The only other way I could think of passing a cancellation token would be a property on the service itself, but I haven't seen anything like that in the wild so it would be a bit confusing.

SharpeRAD:
If your calling from a MVC controller you just add the cancelationToken as a parameter to your action and the framework handles everything's for you, its [really simple](http://www.davepaquette.com/archive/2015/07/19/cancelling-long-running-queries-in-asp-net-mvc-and-web-api.aspx)

All without a "cancel" button :smile:

SharpeRAD:
Its so simple that I don't know why anyone would argue against it?

jaymedavis:
Be sure to read the update in that article, it's not as simple as passing the token as a parameter. You have to create a linked token first. :+1: 

Sounds good, though.

SharpeRAD:
It is if you using .net core :smile: 

> Everything works as expected in both ASP.NET Core MVC and in Web API, but for some reason, MVC 5 only supports cancellation if you use the AsyncTimeout attribute.

jaymedavis:
Why use anything else? :)

I kid. Hope this helps someone - I'll link to that article for reference. I don't think anyone would have any long running Stripe queries (since they limit to 10), but someone might have a lot of traffic and be concerned. 

Thanks.

jaymedavis:
As I mentioned, I've been sick and this is the only reason this isn't fully on master. There is a branch where I pushed this work, however. I appreciate you taking the extra time to separate sync/async code for me. I also fixed some methods in Requestor that could be private methods (nothing you did, you just motivated me to clean up some more things). Lastly, I just wrote a test to verify the deleted functionality. One small issue (I'm not worried about it right now as much), is failed deletes still return a StripeDeleted object with the status of true, and I would expect that to be false even with an exception thrown. 

This is almost complete on my end (if not 100% ready), and I hope to bring it into master today. This was really great work - thanks much!

jaymedavis:
Also, I appreciate your patience working through this with me. I'm very adamant about understanding everything in the codebase, so I appreciate the links.

jaymedavis:
The delete was just a test issue, it's not a problem. On master, will be in next release.

PR #581, opened 2016-08-31
Migrate the project from csproj to project.json

flagbug:
Fixes https://github.com/jaymedavis/stripe.net/issues/544


flagbug:
@jaymedavis This should be ready now!

jaymedavis:
Thanks @flagbug!!!

flagbug:
@jaymedavis What's the status of this? No stress, just want to make sure
everything is alright with this PR!

jaymedavis:
From eyeballing through, it looked good. I need to research the net standard stuff a bit (I am really adamant about understanding everything in my code :)). The only reason it's waiting is 1) I got a bit sick and I'm behind. And 2) I think it will be easier to merge the existing PRs first as is before I make this change. Otherwise, it might shown up as new files or be a nasty merge. I will check it out a bit more and just run questions by you as I have them if you don't mind. Many thanks for all your work @flagbug.

.gitignore
```diff
@@ -1,15 +1,261 @@
-# build and binary files
```
jaymedavis: Regarding this file, I prefer to keep only what I need in this file. I reference this same template when I need to, but my mentality is only having lines of code 100% necessary in every file. I'm happy to fix this though. If there are any specific parts of this you need for development, let me know. Curious - are you on Mac or Windows building this? Also, Xamarin or VS? I don't want to make it harder/annoying for users to submit PRs either.

flagbug:
Alright, let me know when you're finished with the PR merges, then I can update this PR with the changes!

build.cmd
```diff
@@ -1,10 +1,5 @@
-copy .\src\Stripe\bin\Release\Stripe.net.dll .\build\net45\Stripe.net.dll /Y
-copy .\src\Stripe\bin\Release\Stripe.net.xml .\build\net45\Stripe.net.xml /Y
-copy .\src\Stripe.Portable\bin\Release\Stripe.net.dll .\build\portable\Stripe.net.dll /Y
-copy .\src\Stripe.Portable\bin\Release\Stripe.net.xml .\build\portable\Stripe.net.xml /Y
+dotnet pack -c Release src\Stripe.net
 
-cd build
-del *.nupkg
-nuget pack Stripe.net.nuspec
+copy .\src\Stripe.net\bin\Release\Stripe.net.*.nupkg .\build\
```
jaymedavis: I like the rename to Stripe.net here - however, the default namespace is Stripe, and that should be preserved. If it is, ignore this. Once I merge this in, I'm looking forward to getting this setup on AppVeyor!

.gitignore
flagbug: Sure, I can revert this, no problem. I actually just needed to exclude a single folder. I'm on Windows using a combination of VS and VSCode

jaymedavis:
Was this PR mostly a matter of just creating a new project and manually copying all of the files into the new project?

build.cmd
flagbug: The namespace is still the same, I just needed to rename the folder, so `dotnet pack` creates the correct NuGet ID, since you can't specify that in `project.json`

flagbug:
I used this handy tool to convert the `csproj` files to `xproj`: https://github.com/OctopusDeploy/CSProjToXProj

Then I migrated the `.nuspec` file to `project.json`, fiddled a bit with the `build.cmd` and voilà!

flagbug:
Just adding a not that I need to fix this in here too: https://github.com/OctopusDeploy/CSProjToXProj/commit/fcf3a5af8e554bd205211991e57fc2246c93a920

.gitignore
jaymedavis: Same here. Shortly after I started this project, I was a node developer for many years, hence why I'm catching up on some .NET goodness. I used atom for a long time, until I found VSCode was actually better and faster (even in Mac). Thanks!

build.cmd
jaymedavis: Gotcha. Thanks

jaymedavis:
<3 Octopus in general! My experience was short lived in my last DevOps role, but I found it a lot more intuitive than a lot of similar tools. Thanks for explaining the process for my learning benefit. :)   Also, thanks for offering to merge the other PR work, how very kind! Let me know if there's anything I can do (services you need from Stripe), to repay the favor.

jaymedavis:
I think regarding the nuget id, you can specify it with the name attribute.

> Source: This comes from the project name – you can optionally declare a name property in project.json (Source). If no name property is specified, the project directory name is used by default. (Source)

source: http://dotnetliberty.com/index.php/2016/01/26/where-does-dotnet-get-nuget-package-metadata/


The reason I bring this up, is I think we may be able to do this sooner rather than later. I'll have to look through the existing PR's again, but I think that's the only thing that will cause a rough merge conflict. I wanted to knock them out and wait, but new PR's just keep coming. :)

flagbug:
@jaymedavis I just tried that, but now the Stripe.Test project can't find the Library project anymore 😕

flagbug:
@jaymedavis Also see https://github.com/aspnet/dnx/issues/1559

jaymedavis:
Bah :(. Thanks man. I'll get the PRs cleaned up soon enough.

jaymedavis:
You've been busy. :)

It seems AppVeyor (and NuGet) really push you towards building your nupkg by pointing to the project when you call NuGet pack, instead of the nuspec. This PR would make my life a lot easier right now.

Did your recent commits catch this up with master?

flagbug:
@jaymedavis Yup, I merged master into this branch. I also had to restructure the folders, because the .NET Core tooling is pretty opinionated when it comes to project structure: https://docs.microsoft.com/en-us/dotnet/articles/core/tutorials/libraries#how-to-use-multiple-projects

I hope this is okay.

One thing I also want to mention, is that the built-in Visual Studio test runner UI doesn't seem to work with the test runner Stripe.net is using, so only `dotnet test` works right now

jaymedavis:
That's no big deal for me - I don't think many people run the tests anyway. I use resharper and AppVeyor will still work (or we can make it run dotnet test).

jaymedavis:
I'll check it out - it seems this standard wants two projects (.Core and .CSharp) - boo. I can see that for a big project as a benefit, but Stripe.net is just a string builder. :)

flagbug:
@jaymedavis I think that's just an example because their project also uses F#. We just need the Stripe.net and Stripe.Test projects

brentdavid2008:
What's the status of this, is there something I can do to try and help with this?  We are making a new project for a client and would like to use netstandard1.6.  i've done a few libraries where i've targeted 4.5, 4.6 and netstandard1.6 in the same project and it just compiles to all the difference versions someone installs.  Stripe is the last library that doesn't have this support.  :)

jaymedavis:
I have had a few requests from Stripe creep up and I've been working pull requests from the bottom, but they keep coming in. I imagine it's going to stay that way unless this gets implemented. I've also been working a lot on the build - which you can see on the other branch with appveyor. Most of that work will carry over just fine since it's just PS scripts.  It is time to go ahead and get this done though.

Here's what I'm fuzzy on:
```
 "netstandard1.1": {
 +      "buildOptions": {
 +        "define": [ "PORTABLE" ]
 +      },
 +      "dependencies": {
 +        "NETStandard.Library": "1.6.0"
 +      }
 +    }
```

netstandard 1.1 lookd right to me. it will give support to everything except silverlight. I'm fuzzy on the dependency of 1.6 and how this block works - can you point me to an article or explain what this dependency is for and why it is 1.6?

jaymedavis:
I guess maybe the question is - since there is a 1.1 offering, will using a 1.6 dependency in my project cause issues for the consumers?

brentdavid2008:
netstandard is the 'next generation' of PCL's. basically replacing them.

This article talks about what it is from a high level: https://docs.microsoft.com/en-us/dotnet/articles/standard/library

1.6 basically is the latest version. I've struggles with other versions older, like 1.1 and so forth, 1.6 is the first that felt more 'normal' and added all the packages that i would normally see in a regular PCL.

This is what you get when you have that as a dependency:
image

Other than that, it's basically just a PCL.  But just because it's a PCL doesn't mean that it supports 4.5 and 4.6, so on the stuff i've done, you still had to have the other frameworks targeted.

here is one project.json i am using

image

brentdavid2008:
obviously that entire netstandard library is only used if the person consuming the app is using the same type.  which is why you still need each of the other frameworks inside it. and when that happens, and there's an namespace conflict, you still have your compiler directives like normal.  in this case, i want to have EF6 and EFCore supported in this library, so I have both of them dependencies in each of their target platforms, but my .cs files have stuff like this:

image

brentdavid2008:
Hopefully that's helpful, I'm not sure what you already know or don't know.  I still think it's necessary to target the others that are important, if you're not using anything in 4.6+ then just the net45 is enough i think since you can install a 4.5 nuget package into a 4.6 project, but i think it's ok to start with netstandard1.6 for the 'Core' customers.  Just my opinion

brentdavid2008:
Not to bombard you with emails from this chat, but i just took your project currently, added a new solution and here's what i would do if it was me making it support both for the project.json file:

image

there are a few places where you are doing some if !PORTABLE and using System.Configuration which doesn't exist in .net  core, but the short of it is, your project will create one package with the dlls needed for 4.5 and netstandard1.6 and the dependencies for each based on this file.  and you'd end up with something like this:  (this is an example from my generic repository)

image

food for thought.

Anyway if you want to go more in dept sometime together, we can do it on some chat platform, that might be easier.

jaymedavis:
Bombard away @brentdavid2008. I appreciate it!  (also - lol at FULLDOTNET - I love it)

So as I understand it, I would need netstandard 1.1 to support .net45 - [look here](https://docs.microsoft.com/en-us/dotnet/articles/standard/library#net-platforms-support). I believe this is why @flagbug defined the target as 1.1. 1.6 would only work for .net core+ consumers. Am I interpreting this wrong?

The part I'm confused about, was this PR has netstandard1.1 as the target, then a dependency on 1.6.  Also, if netstandard1.1 supports net45, why is there a reason for the net45 target at all? Wouldn't netstandard1.1 account for this according to the link I posted?

jaymedavis:
One more message - I would prefer to have the least amount of dependencies possible that the library uses. I'm not sure if you can cherry pick these or not yet, but I do want to keep in mind that it will have to download on the user's server. I have always wanted stripe.net to be a very small wrapper, with a small footprint.

brentdavid2008:
Yeah, i agree.  Can i upload a zip of what i would do instead?  i don't want to do a pull request since i basically want to re-create the project and change a few things.

Your link if you look at the alias for 'netcoreapp'  => this is what the ASP.NET applications (for .NET core) use by default.  here is project.json framwork reference for a brand new ASP.NET app

    "frameworks": {
    "netcoreapp1.0": {
      "imports": [
        "dotnet5.6",
        "portable-net45+win8"
      ]
    }
    },

because of this, i think you should have both 1.1 and 1.6 as separate frameworks.  (assuming i'm reading that chart right, your guess is as good as mine for what -> means lol) A comment on that section has this:

> The way I understand it, the arrow is like the reverse of "backward compatibility", thus "forward compatibility". The simplest way to look at it is: ".NET Core 1.0, which maps to .NET Standard 1.6, is backward compatible all the way down to .NET Standard 1.0.".

brentdavid2008:
I guess moral of the story is, even if you have 3 or 4 dependencies in the netstandard1.6 version, that's really only because .net core makes you reference everything you need, since nothing else is included, so it's probably _more lightweight_ than what you have now anyway, even though it appears to have a 'lot of dependencies'

jaymedavis:
> The way I understand it, the arrow is like the reverse of "backward compatibility", thus "forward compatibility". The simplest way to look at it is: ".NET Core 1.0, which maps to .NET Standard 1.6, is backward compatible all the way down to .NET Standard 1.0.".

Somehow, I missed that! So really, I think this PR could be targeting 1.6? 

Please do upload a zip with what you would do! If you wouldn't mind, drop it in your public dropbox or something like that, and link it here. I really value the opinion of @flagbug - so I'd like him to see  it as well.  👍

brentdavid2008:
No problem, I can do that later this evening and i'll provide a doc as to why i made certain choices to keep/remove/change things as they come up

jaymedavis:
Excellent @brentdavid2008! Thanks much!

flagbug:
@jaymedavis `Netstandard.Library` implements *ALL*  of the .NET Standard Versions. Take a look at the dependencies in https://www.nuget.org/packages/NETStandard.Library/, here you can see that it supports .NET Standard 1.0 and higher, NuGet then automatically selects the correct target.

flagbug:
> Somehow, I missed that! So really, I think this PR could be targeting 1.6?

@jaymedavis No, this would mean that this library is only compatible with .NET Standard 1.6 and higher. If we target .NET Standard 1.1, like I did in the PR, we have a much broader compatibility

jaymedavis:
@flagbug if netstandard 1.1 includes the dependencies needed for .net45, why does there have to be a separate build for net45? Is it because of the TLS issues or something else?

Also, I'm assuming the dependecy can be bumped to netstandard library 1.6.1?

Thanks man.

jaymedavis:
Also, can I pretty much just pull this down and use winmerge to copy over what's missing? hehe

brentdavid2008:
I concur about netstandard1.1, however, i don't think you can get away with doing the full net45 framework target.  I have a solution that references the dll created from a class library that was set to netstandard1.1  The web app (mvc 5) wants to reference netstandard1.1 in order to use it.  you can't do that in a regular .net46 project.  so in order to support those people, you _have_ to do the other framework.

image


if i add the net45 framework to the project.json file in the class library in addition to the netstandard1.1

this works:

image

brentdavid2008:
here's my class library .json file

image

flagbug:
>  if netstandard 1.1 includes the dependencies needed for .net45, why does there have to be a separate build for net45? Is it because of the TLS issues or something else?

If I remember correctly, `System.Configuration` is used by Stripe.NET to retrieve the API key, which isn't included in .NET Standard, so there is a `#if` somewhere in the codebase where we exclude this functionality in the .NET Standard build

brentdavid2008:
> If I remember correctly, System.Configuration is used by Stripe.NET to retrieve the API key, which isn't included in .NET Standard, so there is a #if somewhere in the codebase where we exclude this functionality in the .NET Standard build

Yes that's correct, i would argue that it should be excluded from all builds if you're taking it out. Let the consuming application determine how it gives the Stripe.NET library it's api key.  Consistency reigns supreme in my book.

jaymedavis:
> If I remember correctly, System.Configuration is used by Stripe.NET to retrieve the API key, which isn't included in .NET Standard, so there is a #if somewhere in the codebase where we exclude this functionality in the .NET Standard build

I'll look around, but if this is the only reason for this, nope. It's gotta go. Maybe not right now, but everything is moving away from that anyway. I'll give it a little time - from watching Connect(); the other day, I was thinking that people might be hesitant to change their ways for .net core. I don't want to exclude them just yet. 

Thanks so much @brentdavid2008 for putting that sample together.

jaymedavis:
Thank you guys so much for the help. Based on this thread, it sounds and looks like this PR is good to go - except for having the version updated and the new/changed files copied over.

jaymedavis:
If you check out [branch pr/581](https://github.com/jaymedavis/stripe.net/tree/pr/581), I have the project updated with all of the latest bits. Also, I have appveyor setup to restore/build/package the project. However, the packaging is failing probably because the tests are not all passing. There are some that appear to be silly errors about file paths, but the file upload tests need to be looked into. This branch is pretty cleaned up, as I removed most of the stuff that isn't needed anymore: the build folder, the build.cmd file, the assemblyinfo, and the old Stripe and Stripe.Portable folders. Checkout the code for that branch and the [latest build here](https://ci.appveyor.com/project/JaymeDavis/stripe-net)

The pros so far:
* This is extremely easy and clean
* Version bump in one place (project.json)
* ONE PROJECT
* I can work on my mac again when I'd like
* I can use VSCode easily

The cons:
* mspec support is lacking - I cannot see tests as they run, only when everything is finished
* I'm not entirely sure how to just run one test file right now, since there aren't any IDE tools and `dotnet test --help` doesn't appear to offer anything

None the less, next steps are to fix up these tests to pass. After that, the nupkg should be built and placed on the build as an artifact - this really needs to be tested pretty well, so I would appreciate it if you could add the nupkg locally before it goes live.

Let me know if you have any feedback after reviewing the branch/build/changes. Thanks again!

jaymedavis:
I ended up overhauling Stripe.Tests a bit - to use the environment and not ConfigurationManager, so I could convert it to a .net core 1.* app. The only problem I have now is a lack of understanding mspec with .met core - my build can be found [here](https://ci.appveyor.com/project/JaymeDavis/stripe-net) for the latest build. It says it fails, but there's not other output. It used to output every test and result, but now it seems to all run at once and then not report what actually failed. @ivanz - am I missing something simple here? How can I make mspec product feedback while the tests are running? and feedback about which test failed? Also, I'm not exactly sure how to run a single test. I know this is a work in progress, and I've loved mspec for many years - just curious if there's something I'm missing?

ivanz:
The test failing bit is due to the bellow, no? As in the test failed message is printed by dotnet test, because it can't run the tests I believe. The mspec output doesn't have any `F` (for "failed" - just one `I` for ignored and `.` for success) 

```
The specified framework 'Microsoft.NETCore.App', version '1.1.0' was not found.
  - Check application dependencies and target a framework version installed at:
      C:\Program Files\dotnet\shared\Microsoft.NETCore.App"
```

jaymedavis:
Thanks @ivanz - certainly was a learning curve figuring out the new package.json. :)

I have `netcoreapp1.0`, `net45`, and `netstandard1.1` in my project. I removed `netstandard1.1` because it seemed like mspec did not like it. The build is compiling and finishing now, but I'm still not seeing anything showing passing tests. It shows ~760 specifications, but the output isn't much more. The same thing locally as on the [latest build over here](https://ci.appveyor.com/project/JaymeDavis/stripe-net),

xt0rted:
> certainly was a learning curve figuring out the new package.json. :)

Just wait until you get to replace it all with [msbuild again](https://blogs.msdn.microsoft.com/dotnet/2016/11/16/announcing-net-core-tools-msbuild-alpha/) 😜

jaymedavis:
Bah @xt0rted 😛  ... good times to be had, thanks for link.

@brentdavid2008 the original issue you were trying to solve, I might be able to copy over this project.json structure and get that working. As for building the project itself in .net core, I think we have some things to figure out first w/regards to mspec. I can't get the test runner to work - I've been messing with the source all day, but it's going to take a long time to figure this out, I suspect. Seeing as it should work for .net core now, maybe there's a middleground? I don't want to abandon moving this forward, but I also don't want to invest the time if it's just a little longer for tooling to catch up. Please feel free to pull this branch down and build/run dotnet test to see my woes. :)

jaymedavis:
dotnet-test

Here's a pic of my build locally to clean up the build debris.

brentdavid2008:
I haven't looked extensively into the code to see what it is that might fail in the core version of the app, but i may just download the project and get a version working on core for what we need to do until you get a real version working for nuget.  I've never used mspec, so i can't speak to it, we use SpecFlow and the default implementation for testing (i think that's mstest?) 

Here's a link:
http://www.specflow.org/

This testing platform allows us to define a feature, so in your case, maybe that's charging a card or something, and you can specify scenarios on that charge. i.e. if it's this card number then it should fail but this card number should not, i don't have to create separate test files for that, it's just separate example lines and the specflow library creates the individual tests.

Either way though, i'll probably just download it like i said, remove what i don't think we need, since it's just a string builder like you said or close to it ;) i'll just let you work out the tools as you have time or as they catch up and i'll skip the test project for what i use.

brentdavid2008:
here is an example of one of my features, with two sencarios i'm working on, but 6 total tests because there are 6 total examples, 3 in each.

image

and here's what they call the 'steps' to the scenario, the .cs file that is the actual test bound to those 'human readable descriptions'

image

jaymedavis:
Wow. This reminds me of my coffeescript days. :) .. that looks outstanding. Good test, too! I wasn't even aware it was broken in nuget, I'm so sorry. I have only tested by adding "Stripe.net" do my dependencies in project.json. I will fix this - I am _hoping_ it is something easy.

brentdavid2008:
No biggy, i was mainly referring to waiting until you had something published in nuget for .net core support, i don't think the current version is broken for the full frameworks.

jaymedavis:
I found a little work around in order to generate this. The only questionable thing is the netstandard1.3 I have - which actually includes netcore. Slack Overflow programming. :)

The netcore branch now has a much smaller difference from master (you can diff, very few changes), and it builds for multiple frameworks. Can you grab the nuget file from the build server https://ci.appveyor.com/api/buildjobs/h0tcysgtanc1mj24/artifacts/src%2FStripe%2Fbin%2FRelease%2FStripe.6.12.0.nupkg and let me know if your environment is happy?

Just add a nuget path to somewhere on your disk and try that out @brentdavid2008 - I can still run all of the tests locally this way - all pass. This will probably be vNext. 👍

jaymedavis:
Hey @flagbug :) 

I figured you might know a bit more about this... I'm trying to make sure the branch mentioned above doesn't hose anyone on portable. I created a new iOS project in visual studio and referenced the nuget file, which did work - but prompted to sign a license to agree to a bunch of things - .net core being at the top. Do you think it's necessary to add a .NETPortable reference to project.json as well[1]? Since it references .netstandard1.3 already, and this requires 1.1 i believe, I am thinking it's ok as is. But I'm not entirely sure what you expect to see from the portable side. Thanks man!

By the way, I only found an article _talking_ about making a work around for having a different folder name than packing - lol - you are totally right about that one. :)

[1] 
```
".NETPortable,Version=v4.5,Profile=Profile111": { }
```

duckwaffle:
Any chance I could throw my 2 cents in and ask for netstandard 1.2 instead of 1.3? That is the one that closest aligns with Profile111 as it supports the Win8.1 platform. I know its a "dead" platform, but the store still accepts submissions, and it is a default project type when creating new Xamarin cross-platform solutions, mainly because 8.1 apps have full support in Windows 10, and is a more mature API than the UWP one.

Also, its much faster to develop/test on windows phone when creating xamarin applications because Monotouch and Monoandroid run like a dog and the debuggers are just terrible haha.

flagbug:
@jaymedavis Generally just using `netstandard1.x` is enough, since it includes `Profile111` and friends. Yes, the license dialog is annoying, but that's just how it is.

jaymedavis:
@flagbug Great, thanks.

@duckwaffle I don't see a problem with that. It supports 45 anyway - I'll see if everything aligns with that.

jaymedavis:
This is done - released 6.12.0. (@duckwaffle 1.2 was no problem)

The tests are at a point where I can at least search the output for failures and not deploy if any of them fail. These tests will probably hang around seeing as there are ~800 of them. Going forward, I think this warrants a look into XUnit, XBehave, etc... The goal is to keep these tests in the results and check for failures. New ones will be written towards another framework. 

Thank you everyone for the help. This was quite a battle. 👍

duckwaffle:
Alright, so here's what I've discovered, netcore/netstandard libraries, or at least, the way 6.12.0 is packaged, is not compatible with Xamarin PCL's with profile111.

I remember running into this issue previously when I was porting some of our libraries over from PCL to netstandard, essentially, Xamarin PCL's do not 100% support netstandard, for more info you can check out https://xamarinhelp.com/dot-net-standard-pcl-xamarin-forms/ .

You can test this out yourself by making a blank Xamarin.Forms app (Portable) and trying to add the latest Stripe.net package (or any netcore package, ie. autofac > 4.0.0) to the solution. It will have no problem adding the package to the "head" mono projects, but it will fail on the PCL.

In order to make the package consumable by a Xamarin PCL the following needs to be added to the project.json:

  "frameworks": {
    "netstandard1.2": {
      "imports": [
        "portable-net45+win8+wpa81"
      ]
    }
  }

nickrandolph:
Jay, further to @duckwaffle 's request, is there a reason why .netstandard1.2 and not .netstandard1.1 (which aligns with profile 111)?

duckwaffle:
Please ignore my previous comment, @nickrandolph is right, I just tried again by compiling the package locally as netstandard1.1 and I had no issues adding to any of my PCL's or head projects.

PR #590, opened 2016-09-09
async cancellation tokens on services

jaymedavis:


src/Stripe/Infrastructure/Requestor.cs
```diff
@@ -48,36 +51,62 @@ public static string PostStringBearer(string url, StripeRequestOptions requestOp
 
             return ExecuteRequest(wr);
         }
+        
+        private static string ExecuteRequest(HttpRequestMessage requestMessage)
+        {
+            var response = HttpClient.SendAsync(requestMessage).Result;
```
flagbug: @jaymedavis This looks like deadlock city, see http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

jaymedavis: Will dig into this later a bit more. This method is supposed to be synchronous, as ExecuteRequestAsync is the version used in Async calls. It still looks like there are some changes that can be made, though.

PR #596, opened 2016-09-12
Fixes #595 Changed ApplicationFee from int? to string

### LINKED ISSUE #595: application_fee property on the StripeTransfer object is the wrong type
https://stripe.com/docs/api#transfer_object
application_fee is `string` and should not be `int?`
This is because the transfer object returns the id of the application_fee object.

application_fee `int?` is correct for StripeTransferCreateOptions however.
https://stripe.com/docs/api#create_transfer

The stripe API appears to use the field with the same name for different purposes.

dplarina:


review-ninja:
<a href="https://app.review.ninja/jaymedavis/stripe.net/pull/596" target="_blank">ReviewNinja

jaymedavis:
Sorry about the Ninja / I wasn't ready for that to go out yet. The tests not running is something I need to figure out with AppVeyor. I'll leave some comments on the code.

src/Stripe/Entities/StripeTransfer.cs
```diff
@@ -18,7 +18,7 @@ public class StripeTransfer : StripeObject
         public int Amount { get; set; }
 
         [JsonProperty("application_fee")]
-        public int? ApplicationFee { get; set; }
+        public string ApplicationFee { get; set; }
```
jaymedavis: You can change this to an expandable property - just look in the codebase for expandable and you'll see how to handle it. This should be renamed ApplicationFeeId so it doesn't confuse anyone - like me.

xt0rted:
@jaymedavis the issue is with the environment variable being secure.

https://www.appveyor.com/docs/build-configuration/#environment-variables

>**However, secure variables are not decoded during Pull Request builds which prevents someone from submitting PR with malicious build script displaying those variables.** In more controlled environment through with a trusted team and private GitHub repositories there is an option on General tab of project settings to allow secure variables for PRs.

Maybe @FeodorFitsner could provide some guidance as to how best to set this up.

jaymedavis:
@xt0rted yeah, I figured that out. Right now, I don't really care and I could open it up, but that will change when my NuGet key is in there. I can always roll my stripe api key (test anyway), so I don't care. However, being able to add a reaction, or a comment that just says "build" would be a nice feature from AppVeyor.

src/Stripe/Entities/StripeTransfer.cs
dplarina: Added, however it doesn't appear to be expandable according to the documentation
https://stripe.com/docs/api#transfer_object

jaymedavis:
Looks perfect. I'll write a test to verify this works. My guess is it is expandable, because it is everywhere else, it's most likely a Stripe documentation issue.

PR #624, opened 2016-10-10
Project Clean up, for better compatibility of DOTNET Core

SperoSophia:
removing multiple projects in favour of single project with project.json configurations.

This will enable us to easily build for multiple targeted frameworks, (2.0, 3.5, 4.6, CLR)
without getting too many sharp projects in the solution.
using the project.json as the control layer.


SperoSophia:
Main project has been removed in favour of a core project that build for multiple targets, so appveyor might fail due to this reason, just have to reconfigure appveyor.

jaymedavis:
Thanks for doing this, but it is a duplicate of #581. @flagbug is waiting on me to catch up with some pull requests so he can refresh that PR to get things moved over. Out of curiousity, how did you go about doing this PR? I think choosing netstandard 1.1 makes a much better choice than 1.6, as the latest I've heard is to "go as low as possible". Let me know if I'm missing something!

SperoSophia:
just created a new core library and moved the project files, easy enough.
as for the "low as possible" rule, this would only be valid for stable Frameworks. 
netstandard has gone over some major changes over the past year. If we were talking about .NET Framework, i'd say go with .NET 2.0 when possible to comply with a wide compatibility since you don't target 4.6.1 for win XP. as for netstandard, it's too new to worry about compatibility of legacy machines.
Additionally, we are bundling the components in the core builds, so we are not referencing other machine related frameworks. basically the runtime is included in the build. (which you are aware of i'm sure).

jaymedavis:
Thanks for the discussion. I am aware it's all bundled, but I still thought netstandard1.6 only supported 4.6+, which would lose compatibility with 4.5. Perhaps that was what you are suggesting?

jaymedavis:
Re-read the docs. :)

So I think Stripe.net itself is fine running netstandard 1.6, it would be the consumers that would be a problem. Since we are trying to make *one assembly to rule them all*, anyone that implemented Stripe.net would have to have an equal to or greater version of netstandard on the consuming library. It is nice to know the list of APIs needed by Stripe.net follows 1.6, however. Thanks again for your time!

PR #657, opened 2016-10-27
added a raw json property to all entities, closes #378

### LINKED ISSUE #378: Add raw JSON property to entities
I like using stripe.net to create customers and make charges. I would also like to store the raw JSON response from these requests for logging purposes. It would be great if each service request would return the raw JSON as well as the parsed entity.

jaymedavis:


NeedHeelp6989:
Can you show me how to understand and make some money I need a favor I'm hoping you can do me a solid

jaymedavis:
trollolol

Issue #575, opened 2016-08-28
Organize the readme in alphabetical order


jaymedavis:
I want all of the examples on the readme in alphabetical order, like they are in the links at the top of the readme. PR's welcome if you want to help!

PR #673, opened 2016-11-10
Order readme alphabetically

t0mburton:
Every little helps! #575


jaymedavis:
Sure does!! Thank you!!

Issue #649, opened 2016-10-24
StripeCustomerService List returns invalid Source data


viralr:
Currently if I use StripeCustomerService().List(), it returns a list of customers as StripeCustomer. The issue is if the customer has an ACH account as a payment method, the account is added as part of StripeCustomer.SourceList which is a of type List < StripeCard > . I think the ACH accounts should be added to StripeCustomer.CustomerBankAccounts which is of type StripeList < CustomerBankAccount >

jaymedavis:
This actually needs to be of type StripeList < Source > -- which supports everything. Source allows the list to be of multiple types - card and bank account are supported now. This was changed in StripeCharge, it just needs to be updated in StripeCustomer too (or anywhere StripeCard and StripeBankAccount/CustomerBankAccount) are used.

PR #674, opened 2016-11-14
StripeCustomerService List returns invalid Source data #649

viralr:
StripeCustomerService List returns invalid Source data #649

### DIFF
--- a/src/Stripe.Portable/Stripe.Portable.csproj
+++ b/src/Stripe.Portable/Stripe.Portable.csproj
@@ -226,6 +226,9 @@
     <Compile Include="..\Stripe\Infrastructure\JsonConverters\SourceConverter.cs">
       <Link>Infrastructure\JsonConverters\SourceConverter.cs</Link>
     </Compile>
+    <Compile Include="..\Stripe\Infrastructure\JsonConverters\SourceListConverter.cs">
+      <Link>Infrastructure\JsonConverters\SourceListConverter.cs</Link>
+    </Compile>
     <Compile Include="..\Stripe\Infrastructure\JsonConverters\StripeDateTimeConverter.cs">
       <Link>Infrastructure\JsonConverters\StripeDateTimeConverter.cs</Link>
     </Compile>
--- a/src/Stripe.Tests/card_behaviors.cs
+++ b/src/Stripe.Tests/card_behaviors.cs
@@ -25,40 +25,40 @@ namespace Stripe.Tests
             StripeCustomer.LiveMode.ShouldEqual(false);
 
         It should_have_an_id = () =>
-            StripeCustomer.SourceList.Data.First().Id.ShouldNotBeNull();
+            StripeCustomer.SourceList.Data.First().Card.Id.ShouldNotBeNull();
 
         It should_have_the_correct_source_address_line_1 = () =>
-            StripeCustomer.SourceList.Data.First().AddressLine1.ShouldEqual(StripeCard.AddressLine1);
+            StripeCustomer.SourceList.Data.First().Card.AddressLine1.ShouldEqual(StripeCard.AddressLine1);
 
         It should_have_the_correct_source_address_line_2 = () =>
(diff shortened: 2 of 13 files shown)


jaymedavis:
Thanks for this - much needed. Have you seen the [SourceListConverter](https://github.com/jaymedavis/stripe.net/blob/master/src/Stripe/Infrastructure/JsonConverters/SourceConverter.cs)? I'm wondering if there's some way to leverage this without creating a list converter. Thoughts?

viralr:
How about I change SourceConverter to return a list if json object is of type list, else return a single object. That should cover both implementation and I can remove SourceListConverter

viralr:
`using System;
using Newtonsoft.Json;
using Newtonsoft.Json.Linq;

namespace Stripe.Infrastructure
{
    internal class SourceConverter : JsonConverter
    {
        public override bool CanConvert(Type objectType)
        {
            throw new NotImplementedException();
        }

        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
        {
            var incoming = JObject.FromObject(value);

            incoming.WriteTo(writer);
        }

        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
        {

            if (incoming.SelectToken("object").ToString() == "list")
            {
                var list = incoming.ToObject<StripeList<dynamic>>();

                var result = new StripeList<Source> { Data = new List<Source>() };

                foreach (var item in list.Data)
                {
                    result.Data.Add(item.SelectToken("object").ToString() == "bank_account"
                                        ? new Source() { BankAccount = item.ToObject<StripeBankAccount>(), Type = SourceType.BankAccount }
                                        : new Source() { Card = item.ToObject<StripeCard>(), Type = SourceType.Card });
                }

                return result;
            }
            
            var source = new Source
            {
                Id = incoming.SelectToken("id").ToString()
            };

            if (incoming.SelectToken("object").ToString() == "bank_account")
            {
                source.Type = SourceType.BankAccount;
                source.BankAccount = Mapper<StripeBankAccount>.MapFromJson(incoming.ToString());
            }

            if (incoming.SelectToken("object").ToString() == "card")
            {
                source.Type = SourceType.Card;
                source.Card = Mapper<StripeCard>.MapFromJson(incoming.ToString());
            }

            return source;
        }
    }
}

`

jaymedavis:
Actually, I was hoping for more of a 'merge' between the two - like using the code from the existing one with the list. After looking at this separate, I think the original way you had it is best. That way it's a little smaller class and they handle a different concern. Curious - why use item.ToObject over the built in mapper? Does that automatically populate the response with the json attributes?

jaymedavis:
If I'm going to introduce this breaking change, I'd like to ahead and get Source and SourceList everywhere in the app that requires it. If you have time, that's awesome. If not, I can get this work from you started on a branch (once I finish up some other items), and we can collaborate on that if you'd like.

viralr:
Ahh yes. I will substitute item.ToObject to use the in built mapper. I can get to changing it everywhere in the app may be by next week. Is that ok? Also do we want to replace usages of Stripe Card or StripeBankAccount with Source in all the entities? Like I am unsure about whether to change StripeToken or not. Is it possible to get a list of entities that requires this change?

And was just curious about the test failures on my previous merge. Is that due the api string not being set?

jaymedavis:
Test failures aren't on you - AppVeyor doesn't run them on pull requests because someone could submit malicious code and find my private keys. I will fix this to only fire on my branches. Once I push this to a branch, the tests will run.

Regarding Source - yes, mostly just everywhere there is a StripeCard and that endpoint allows a card or a bank account. Tokens shouldn't be part of that. Ideally, i would like to leave the existing StripeCards in place and just add another property for the Sources. That way, we can just mark the StripeCard or StripeBankAccount stuff as all obsolete. I can work on getting that list together for us. Really appreciate you helping whenever time permits - no worries or rush. Thanks!

jaymedavis:
Thanks - this was great. The only thing missing was the converter needed to populate the base properties of the list (Url, HasMore, etc). Released in 7.0.0

Issue #683, opened 2016-11-27
Setting StripeConfiguration.HttpMessageHandler fails on Android project


duckwaffle:
Hello,

This is going back to an issue from a couple of months ago, but I've only just gotten around to upgrading to your newest version, when you added support to use custom HttpMessageHandlers I'm now getting a System.MethodAccessException.

I did a bit of poking around, and it seems that its happening because you are setting the HttpClient in the static constructor of the Internal Requestor class. Unfortunately, its not possible to do it this way in the PCL, because (for some reason) it can't handle the public/internal interaction.

I was able to fix this problem by setting the HttpClient in the Getter of the property in the Requestor class, see below:

        private static HttpClient httpClient;

        internal static HttpClient HttpClient
        {
            get
            {
                if (httpClient == null)
                {
                    if (StripeConfiguration.HttpMessageHandler == null)
                    {
                        httpClient = new HttpClient();
                    }
                    else
                    {
                        httpClient = new HttpClient(StripeConfiguration.HttpMessageHandler);
                    }
                }

                return httpClient;
            }
        }

Yes, I know this doesn't make a lot sense, such is the joy of Android development :(

duckwaffle:
Decided to just go ahead and make a PR for it

jaymedavis:
have you had a chance to try 6.12.0? let me know if that still has the issue -

duckwaffle:
Not yet, although it's the first thing I'm gonna do when I get into the office, Sydney time makes this a bit asynchronous :)

duckwaffle:
Sorry it's taken me so long to reply, its been a crazy day of releases!

Essentially, this is not fixed by 6.12.0 because 6.12.0 is no longer compatible with my project because of the migration to .netstandard/core (I will add more info to the relevant issue).

Additionally, I just realised that the original problem that I was having was not due to it being an android project, but because of a lot of dependency injection weirdness that is going on (the project is rather large, and inherited from another company who lets just say didn't have the best code practises).

Long story short, it looks like in order to continue using this package for this project, I will have to roll my own versions from now on :(

duckwaffle:
As I said in my last comment in #581 ignore my last comment.

When I tested using 6.12.0 modified to be netstandard1.1, this issue no longer occurred, so it looks like the migration to netcore did fix the original problem.

duckwaffle:
[...]

PR #684, opened 2016-11-28
Moving contruction of HttpClient from static constructor to property getter

duckwaffle:
PR for issue #683


jaymedavis:
Why?

jaymedavis:
Let me know if 6.12.0 has this same issue. netstandard might treat this a bit differently.

Issue #730, opened 2017-01-30
When creating a managed account: Invalid URI: The format of the URI could not be determined.


KennyKaplanHP:
Getting the error noted when trying to create a managed account. Just migrated from 6.4 to 7.0.3 and this is one of our key uses. Creating a customer works just fine.

Here's how we're calling it:
`
                var newAccount = new StripeAccountCreateOptions
                {
                    Email = user.Email,
                    Managed = true,
                    Country = "US",
                    DefaultCurrency = "USD",
                    TosAcceptanceDate = DateTime.Now,
                    TosAcceptanceIp = ip,
                    TosAcceptanceUserAgent = agent,
                    LegalEntity = new StripeAccountLegalEntityOptions
                    {
                        Type = "individual"
                    },
                    TransferScheduleInterval = "daily",
                    TransferScheduleDelayDays = "5"
                };
                return await _stripeAccountService.CreateAsync(newAccount);
`

Stacktrace:
   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
   at System.Uri..ctor(String uriString)
   at Stripe.Infrastructure.Requestor.BuildRequest(HttpMethod method, String url)
   at Stripe.Infrastructure.Requestor.GetRequestMessage(String url, HttpMethod method, StripeRequestOptions requestOptions, Boolean useBearer)
   at Stripe.StripeAccountService.<CreateAsync>d__6.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Service.Stripe.StripeService.<CreateAccount>d__14.MoveNext() in C:\_Dev\Service\Stripe\StripeService.cs:line 97

fscahill:
same problem. Awaiting response....

jaymedavis:
If someone could build a PR or branch with a test reproducing this, that would help me out. Alternatively, just add a project reference to the library directly, and let me know the values being passed at the time of the exception. There's a visual studio plugin to break on any CLR exceptions that might come in handy.

* To setup a test

1) Clone the library
[...]

PR #738, opened 2017-02-10
Added List and ListAsync methods to StripeAccountService

onemancat:
Added List and ListAsync methods to StripeAccountService, along with unit tests and readme update. Also fixed a bug that prevented StripeAccountService.CreateAsync from working.

This allows the listing of all connected accounts, useful in my case for the maintenance of stored credit cards. I want to delete all saved credit cards from all connected accounts within a certain time period, and needed a way to iterate through all connected accounts on a nightly basis.

I was thinking about adding some kind of list iterator to Stripe.net to automatically iterate through all connected accounts (100 at a time, as 100 is the max list limit in Stripe API), which could also work with other returned lists, but I didn't want to make architectural presumptions on how you would like to structure that in this project (this being my first pull request to it), so for now I will do that in my client code.

### DIFF
--- a/src/Stripe.net/Services/Account/StripeAccountService.cs
+++ b/src/Stripe.net/Services/Account/StripeAccountService.cs
@@ -1,6 +1,7 @@
 ﻿using System.Threading;
 using System.Threading.Tasks;
 using Stripe.Infrastructure;
+using System.Collections.Generic;
 
 namespace Stripe
 {
@@ -15,6 +16,16 @@ namespace Stripe
         {
             return Mapper<StripeAccount>.MapFromJson(
                 Requestor.PostString(this.ApplyAllParameters(createOptions, $"{Urls.BaseUrl}/accounts", false),
+                SetupRequestOptions(requestOptions))
+            );
+        }
+
+        public virtual IEnumerable<StripeAccount> List(StripeListOptions listOptions = null, StripeRequestOptions requestOptions = null)
+        {
+            var path = $"{Urls.BaseUrl}/accounts";
+
+            return Mapper<StripeAccount>.MapCollectionFromJson(
+                Requestor.GetString(this.ApplyAllParameters(listOptions, path, false),
                 SetupRequestOptions(requestOptions))
             );
         }
@@ -57,6 +68,17 @@ namespace Stripe
         {
             return Mapper<StripeAccount>.MapFromJson(
                 await Requestor.PostStringAsync(this.ApplyAllParameters(createOptions, "{Urls.BaseUrl}/accounts", false),
+                SetupRequestOptions(requestOptions),
+                cancellationToken)
+            );
+        }
+
+        public virtual async Task<IEnumerable<StripeAccount>> ListAsync(StripeListOptions listOptions = null, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var path = $"{Urls.BaseUrl}/accounts";
+
+            return Mapper<StripeAccount>.MapCollectionFromJson(
+                await Requestor.GetStringAsync(this.ApplyAllParameters(listOptions, path, false),
                 SetupRequestOptions(requestOptions),
                 cancellationToken)
             );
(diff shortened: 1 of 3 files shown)


src/Stripe.net/Services/Account/StripeAccountService.cs
```diff
@@ -56,7 +67,18 @@ public virtual StripeDeleted Delete(string accountId, StripeRequestOptions reque
         public virtual async Task<StripeAccount> CreateAsync(StripeAccountCreateOptions createOptions, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
             return Mapper<StripeAccount>.MapFromJson(
-                await Requestor.PostStringAsync(this.ApplyAllParameters(createOptions, "{Urls.BaseUrl}/accounts", false),
+                await Requestor.PostStringAsync(this.ApplyAllParameters(createOptions, $"{Urls.BaseUrl}/accounts", false),
```
jaymedavis: nice catch. this must be the root cause of #730

jaymedavis:
great work on this!

typically speaking, I'm keeping stripe.net strictly a tight wrapper to their api. I have considered a pro-version though (similar tooling as www.checkoutnet.com). would like to hear your thoughts and opinions on implementing

onemancat:
Thanks, Jayme. You should be proud of the Stripe.net tool, it is very useful and a nicely crafted bit of code as well.

I agree that for something like Stripe.net, the tight wrapper approach is best because it maximizes initial comprehension, minimizes learning curve, and basically provides loads of useful documentation for free (Stripe's own docs). But I do feel that list iteration might be a useful exception, because there are so many more use cases where you want the entire list than where you want some subset--in fact, at this moment I'm at a bit of loss to describe a single use case where you would *not* want full list iteration! If I'm right about this, then most/all clients who call a List method are then having to write the exact same client code to iterate through the list fetches--so many C# "while" statements worldwide! If I recall correctly, even otherwise strict AWS wrappers implemented these pretty early on. 

I haven't searched for this deeply, but when using Stripe.net to get a List of customers, for example, is the has_more exposed anywhere there? If not, how would client users know whether to make another List call? If this is not being exposed, then I would argue that the List methods are probably not in wide use at all, and that a List iterator would be a crucial upgrade.

As far as implementation goes, there are 20 current List methods in the different services, each with a distinct method signature, which is why you cleverly used reflection when applying parameters to the url. All List methods use either StripeListOptions or a derived class as the first argument. Having a full listing capability would render the StripeListOptions (start, end, limit) moot, as those would instead be under the control of the iterator, however any properties from derived classes still need to be respected. So to maintain backward compatibility, I think the most elegant approach would be to add an additional property to StripeListOptions:

        [JsonIgnore] // Do not send this property along to Stripe API
        public bool IterateFullList { get; set; }

This property would be false by default, of course, so would be fully backward compatible. When IterateFullList is true, the Limit, StartingAfter, and EndingBefore have any client-set values ignored and instead fall under the control of the implementing method.

In each List method, where there is now something like this:

            return Mapper<StripeCustomer>.MapCollectionFromJson(
                Requestor.GetString(this.ApplyAllParameters(listOptions, Urls.Customers, true),
                SetupRequestOptions(requestOptions))
            );

In each of the 20 List methods I would do something like this:

if (listOptions != null && listOptions.IterateFullList)
{
	return IterateList<StripeCustomer>(listOptions, requestOptions);
}
else
{
	// Call the existing method
        return Mapper<StripeCustomer>.MapCollectionFromJson(
                Requestor.GetString(this.ApplyAllParameters(listOptions, Urls.Customers, true),
                SetupRequestOptions(requestOptions))
        );
}

IterateList would be a base class method in StripeService (or could be an extension method) which did the following:

1) Set listOptions properties for StartingAfter and EndingBefore to null, Limit to 100 (the Stripe API maximum), and IterateList to false, retaining all other list options. Reflection is not needed, as these are properties of the StripeListOptions base class.

2) Call the existing List method with the new list options, and examine the result. 

3) If Stripe API sends back "has_more": false, then map the result to a collection, append to the full result collection and return. If "has_more": true, then set StartingAfter to the ID of the last object received, and go to step 2.

In terms of modifying other existing Stripe.net code, it looks like we may need to add a JSON mappings to get "has_more" (I see HasMore in StripeList, but those StripeLists seem to only be used for child lists), and then something to get the last ID that was returned.

Then there would be some unit tests for getting a long list (although this would *seriously* clutter up the tester's Stripe sandbox!), and then minor documentation adds for the IterateList optional parameter.

I have perhaps missed the boat of some of this implementation idea, as I've only spent a small amount of time with the code and I'm sure there's a lot I haven't noticed, but this is how I think about it given what little I've seen so far.

Issue #626, opened 2016-10-11
Card expiration month & year are returned as strings instead of integers


xt0rted:
When getting a card's information the expiration month & year are returned as strings ([StripeCard.cs#L16-L20](https://github.com/jaymedavis/stripe.net/blob/master/src/Stripe/Entities/StripeCard.cs#L16-L20)), while in the API docs it shows them [as integers](https://stripe.com/docs/api#card_object-exp_month). Is there a reason for them being strings or can those properties be updated?

jaymedavis:
Was this a Stripe recent update? I don't recall them ever using int's. I will research this more - thanks for reporting.

xt0rted:
I'm not sure. I noticed they were strings when checking if there was a property indicating if the card was expired or not.

jaymedavis:
Well, talking to Stripe about it in the IRC room, it seems they have always been ints. I'll ask around a little more tomorrow to verify, but it's highly possible I just got it wrong and it's always worked as is. However - you are right. If they are integers on Stripe, they should be integers here. 👍

jaymedavis:
better late then never 👍

PR #744, opened 2017-02-20
Product support

aschaible:


jaymedavis:
first off, this is a serious PR, thank you very much. 👍 

All the entity stuff, services, tests, look good. It's that middleware we'll have to figure out. This uses the old 'parameter builder' stuff, where as now it just uses a series of plugins. I would love to find a way to merge this in as is. Does this work for you locally along with all of the other services?

aschaible:
@jaymedavis thanks! @daniel-cohen did the vast majority of the work.  I am using several of the services - cardService, customerService, balanceService, and a few others.  I've ran the unit tests the best I can (many fail due to it saying invalid api key) and everything seems to be working.  Do you have the ability to run the unit tests on your side to verify?  I'm pretty sure everything is working but I don't have much experience with the codebase.

jaymedavis:
You can set the `stripe_test_secret_key` environment variable to your stripe test key, that will take care of that. Sorry it's lowercase... I know that's not typical of environment variables.

lostmsu:
Hi @aschaible , any progress on this? If you're not planning to work on this, I might be willing to take over the PR completion.

aschaible:
@lostmsu The only thing that needs to be done is run the tests and make sure everything is working.  I'm not too familiar with XUnit but I think I just need to set a key in the environment variables.  Many of the tests passed for me and the only failures thus far were key invalid exceptions.  I'll merge in master and see if I can get all of the tests to pass.  I'm using this branch in production right now with good results.

aschaible:
I will be wrapping this up today - almost finished.  There are a few failing tests, but they are failing against master too.

aschaible:
This is ready to go.  The tests fail when run in parallel due to concurrency issues and it looks like the appveyor runner is running them in parallel.  It would be possible to correct this by using multiple "cached" entities but to some degree that defeats the purpose of the test Cache object - at least how I have inferred the usage of it.  It might be better long-term to statically initialize everything in Cache, or just get rid of the concept of sharing these Stripe results.

jaymedavis:
this looks great, from a quick glance. I'm not sure I want to introduce the breaking changes to expiration month/year just yet, as that would force me to bump to 8.0. I think the fixture route is the best bet (the way you implemented it), because that will be implemented as a static automagically. Otherwise, the constructor fires with xunit on every ```[Fact]```. I'll have to look deeper to see why cache was implemented as well. Thanks so much for all of your work!

jaymedavis:
I pulled this down, merged in master (just the version was the only difference) and ran the tests. Most of them failed for me, so I pushed them remote to see if it's something with my environment - you can follow that here: https://github.com/jaymedavis/stripe.net/tree/pr/744. I will probably work from this branch, just to verify everything and fixup any remaining items - things that need to be expandable, etc.

aschaible:
Hello - yes the tests mostly failed for me too if I let the test runner run all tests in the project.  If I limited them to a single test class/fixture at a time, only the tests in the 'account' section failed and the error was in regards to my Stripe test account not having access to those features.  When the tests outside of the account tests failed, it was an exception about the dictionary already containing that key.   The methods within cache create an Items dictionary and add to it.  The first line in the method is to check to see if something exists in the cache and if so return it.  If not, it generates the item, adds it to cache, and returns it.  There's a race condition here if two different fixtures request the item simultaneously, one of them is likely to fail.  Here's an [example of the check](https://github.com/aschaible/stripe.net/blob/b335bdee5861c3d44715780a6ef1ddc210c6f9b7/src/Stripe.Tests.XUnit/orders/_cache.cs#L20).  I could convert to ConcurrentDictionary or restructure the tests to avoid this.

Happy to remove the fix for #626 if you want - just let me know.

jaymedavis:
Thanks for clearing that up. I am going to approach this by cherry picking the services. As is, this is so large, it's very hard to verify. I am going to create a branch for 'Orders', and just do that service, those entities, those tests, and then merge that in. This is great work as is, but it would take a long time to see this hit stripe.net as is, just because it's so large.

If you want to help me break this into smaller pieces that are easier to release, that would be appreciated. I am happy to do the work to break this up into the individual services though.

aschaible:
Sure thing, no problem.  I'll work through these over the next couple days.

jaymedavis:
I've already started orders, so I'll ping you on that branch when it's up. Some small items I've run into (no big deal, with a lot of work it's bound to happen!):

1) Every entity should inherit from StripeEntity or StripeEntityWithId - this way, StripeResponse will be populated from the service call (read about this towards the top of the readme)
2) In the comment documentation, I am removing the "Gets or sets " ... just because the entities should never be set, they are only responses from Stripe. I'm sticking with their exact text on stripe.com/api/docs. I haven't documented the infrastructure of stripe.net well enough, but the flow goes -> Populate ```xOptions``` -> Call a ```Service``` -> Get an ```Entity```. Small detail, but I don't want anyone thinking they can set the entities values from there.
3) You're awesome. :)

I'll just leave this PR open until all it's goodness is copy/pasted and included in stripe.net. I'll reference it from every sub-PR created. Working on orders right now 👍

aschaible:
@jaymedavis awesome!  Lots of the work was @daniel-cohen -- I added some tests and fixed a small bug.  Good info on the structure of the project, thanks much.

jaymedavis:
The more I get through this, the more I'm realizing how dependent these are on each other. I got through copying most of the entities for orders, and realized I have to have a SKU to test the orders. I have to have a product to test those SKUs... I am starting to see why it was implemented altogether in the first place. Most of the entities are re-done (different comments and a few small changes), but it looks like a base level of product and sku will have to be created just to test orders. So tomorrow, I'm moving on to products. I'll check in what I have so far, as that will at least demo how I am doing testing now

aschaible:
@jaymedavis OK I think I see how the testing is done - basically getting rid of the cache and generating a new entity every time.  Do you think it would be better for me to go back to the original PR and update the entities to extend the appropriate objects, update the tests with this approach, and remove the fix for #626 or should I work off of your 744-orders branch?

jaymedavis:
Going back to this PR might make the most sense, seeing as these services rely on each other. Your approach seems good - 744-orders can just be merged back into this. Removing 626 would be helpful too.

jaymedavis:
Regarding testing, I only use the cache when an entity needs to be used over and over (think TestHelper). If I just need it once, I use the fixture. I don't move it to cache unless I have too

aschaible:
I made some progress here.  I don't quite understand what is wrong with the subscription tests, but I made the updates to the other tests and they are running in parallel successfully.  I also updated the entities to extend StripeEntity or StripeEntityWithId and removed the fix for #626.  Can you take a look @jaymedavis ?

jaymedavis:
Thanks. I am not sure what is going on with the tests - I wouldn't think subscriptions or the tests would have anything to do with the new bits, but something could be missing. One thing I noticed as I started looking at SKUs, was the options entity inherits from the data entity that is designated as a return object, which I want to avoid. It may cause some double typing here and there, but so far the options have been completely separated from the entities. Reason being if a data point is added to Stripe (which happens a lot), it's expected to just add that value to the entity. In this case, it would also add another option when you are creating the request, which could bust.

Is INestedNamedOptions necessary anymore? Was it changed to inherit from INestedOptions just as  a quick fix? It's fine if it was, that means it's easy to remove. :)

It looks like the `Products` might be able to be implemented independent of anything else right now. The only missing property when it's returned would be "SKUs" - but I think it stands alone besides that. Maybe the best approach is to comment out everything except that, get everything working without being tied to the return entities or any additional infrastructure (unless it's needed). Then we could start slowly adding things back - I don't have any problem with the code being commented out... make sense?

Not sure if this is the best way to handle this, though - I'll keep digging a bit.

jaymedavis:
Btw - I like the strategy of using different fixtures in the constructors for testing. That somehow did not go through my brain when I started writing the cache stuff - if this works out, it's a much better approach 👍

jaymedavis:
Ok, I've had a chance to go through all of the files and comment everything out. The build is on the 744-commented-out branch. There is still a test or two failing with an error about using the same token for a customer charge. I'm not sure yet why this doesn't exist on master.

Here's some notes/questions:

1) It appears you ported all of the tests from the mspec way to the xunit way. I think we are safe to remove the mspec tests now? -@aschaible
2) The ParameterBuilder.cs was a very old version, and I just restored the version from master. I imagine it was to use the INestedNamedOptions, which I also removed. I believe the functionality that was intended to solve has already been solved and we can just mark things with INestedOptions.
3) A lot of properties have changed since this was first implemented, but it's still a hell of a good head start!

I think this is workable from here to start adding things back in as they are verified from Stripe's docs. I just need to figure out this test issue.

Issue #718, opened 2017-01-16
Allow setting `additional_owners` to null


remi-stripe:
In Europe, Stripe always requires that you provide information about additional owners of the company and this is documented [here|https://stripe.com/docs/connect/updating-accounts#additional-owners]. You can have up to 4 additional owners as they need anyone owning 25% or more but you don't always have 4.

You need to explicitly tell to Stripe's API whether you have more (or any). This is done by explicitly indicating this. For example in PHP, you'd update the Stripe account this way:
```
$account = \Stripe\Account::retrieve({CONNECTED_STRIPE_ACCOUNT_ID});
// Indicate that there are no additional owners
$account->legal_entity->additional_owners = null;
$account->save();
```

Behind the scenes, our bindings are explicitly setting `legal_entity[additional_owners]: ""` to tell us to expect none. In the logs for the update (in the connected account's dashboard) you'd see this as a parameter:
```
{
  "legal_entity": {
    "additional_owners": ""
  }
}
```

A similar logic is taken on account creation though you don't pass null and instead explicitly pass the `legal_entity[additional_owners]: ""`. In PHP the code would look like this:

```
$account = \Stripe\Account::create(
  array(
    "country" => "FR",
    "managed" => true,
    "legal_entity" => array(
        "type"               => "company",
        "additional_owners"  => ""
    )
  )
);
```

Stripe.net needs to support something like this to allow users to explicily tell the API that they won't have any additional owners on creation or update.

jaymedavis:
Thanks for the detail! 

A good exercise might be recreating [this test](https://github.com/jaymedavis/stripe.net/blob/ff333215d5e0cec1f0a32ef11129e5ae8769b7eb/src/Stripe.net.Tests.xUnit/accounts/adding_additional_owners.cs) but omitting the values in the AdditionalOwners - something like this:

```
var newAccount = new StripeAccountService(_stripe_api_key).Create(
                new StripeAccountCreateOptions
                {
                    Email = $"koop@ajah.com",
                    Managed = true,
                    LegalEntity = new StripeAccountLegalEntityOptions
                    {
                        AdditionalOwners = new List<StripeAccountAdditionalOwner>
                        {
                            new StripeAccountAdditionalOwner { }
                        }
                    }
                }
            );
```

[...]

PR #746, opened 2017-02-20
Adds support for specifying no additional owners

agentcoops:
This PR addresses #718 and allows for setting the additional_parameters field to null on a managed account. The proposed syntax is:
`AdditionalOwners = new List<StripeAccountAdditionalOwner> { }`

### DIFF
--- a/src/Stripe.net.Tests.XUnit/accounts/when_no_additional_owners.cs
+++ b/src/Stripe.net.Tests.XUnit/accounts/when_no_additional_owners.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using Stripe.Infrastructure;
+using FluentAssertions;
+using Xunit;
+
+namespace Stripe.Tests.xUnit
+{
+    public class when_no_additional_owners : test
+    {
+        private StripeAccountCreateOptions AccountOptions { get; }
+        private String AccountJSON { get; }
+
+        public when_no_additional_owners()
+        {
+            AccountOptions =
+                new StripeAccountCreateOptions
+                {
+                    Email = $"merry@christmas.com",
+                    Managed = true,
+                    LegalEntity = new StripeAccountLegalEntityOptions
+                    {
+                        AdditionalOwners = new List<StripeAccountAdditionalOwner> { }
+                    }
+                };
+        }
+
+        [Fact]
+        public void has_legal_entity()
+        {
+            var accountString = new StripeAccountService(_stripe_api_key).ApplyAllParameters(AccountOptions, $"{Urls.BaseUrl}/accounts", false);
+            accountString.Should().Contain("legal_entity[additional_owners]");
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


src/Stripe.net/Infrastructure/Middleware/ParserPlugin/AdditionalOwnerPlugin.cs
```diff
@@ -12,8 +12,15 @@ public bool Parse(ref string requestString, JsonPropertyAttribute attribute, Pro
...
 
+            if (owners == null) return true;
+            if (owners.Count == 0)
+            {
+                RequestStringBuilder.ApplyParameterToRequestString(ref requestString,
+                    attribute.PropertyName,
+                    "");
+                return true;
+            }
             var ownerIndex = 0;
```
jaymedavis: owners could still be null here (causing this to fail) - see above

src/Stripe.net/Infrastructure/Middleware/ParserPlugin/AdditionalOwnerPlugin.cs
```diff
@@ -12,8 +12,15 @@ public bool Parse(ref string requestString, JsonPropertyAttribute attribute, Pro
             if (attribute.PropertyName != "legal_entity[additional_owners]") return false;
 
             var owners = ((List<StripeAccountAdditionalOwner>) property.GetValue(propertyParent, null));
-            if (owners == null) return true;
```
jaymedavis: instead of this (or maybe in addition?), why not just check for null, apply the parameter and return true? ...

I realize this would change the test a bit. You would just set it to null instead of a new empty list. 

Alternatively, your check could just be owners.Count == 0 || owners != null. I think this would allow both ways to work.

src/Stripe.net.Tests.XUnit/accounts/when_no_additional_owners.cs
```diff
@@ -0,0 +1,36 @@
...
+using Stripe.Infrastructure;
+using FluentAssertions;
+using Xunit;
+
+namespace Stripe.Tests.xUnit
+{
+    public class when_no_additional_owners : test
+    {
+        private StripeAccountCreateOptions AccountOptions { get; }
+        private String AccountJSON { get; }
```
jaymedavis: this doesn't appear used anywhere

src/Stripe.net.Tests.XUnit/accounts/when_no_additional_owners.cs
```diff
@@ -0,0 +1,36 @@
...
+        public when_no_additional_owners()
+        {
+            AccountOptions =
+                new StripeAccountCreateOptions
+                {
+                    Email = $"merry@christmas.com",
+                    Managed = true,
+                    LegalEntity = new StripeAccountLegalEntityOptions
+                    {
+                        AdditionalOwners = new List<StripeAccountAdditionalOwner> { }
```
jaymedavis: this is what my notes below refer to

src/Stripe.net/Properties/InternalsVisibleTo.cs
```diff
@@ -1,3 +1,4 @@
 ﻿using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Stripe.net.Tests")]
+[assembly: InternalsVisibleTo("Stripe.net.Tests.xUnit")]
```
jaymedavis: i imagine this was an artifact that worked with AccountJSON. This is fine by me, just want to confirm it isn't needed anymore.

PR #770, opened 2017-03-06
First version of serviceitems, with syncronous

TVNielsen:
Changes:
- The StripePlan in a StripeSubscription can now be null if there are multiple SubscriptionItems included 
- StripeSubscription now supports SubscriptionItems, a collection of individual plans in a subscription instead of having only one.
- StripeSubscriptonItem is added, to encompass a StripePlan with a quantity on a subscription
-  Infrastructure.Urls now contain SubscriptionItem base URL
- The SubscriptionItemService is added to the Subscription namespace


src/Stripe.net/Entities/StripeSubscriptionItem.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Newtonsoft.Json;
```
jaymedavis: this file is already in the library

src/Stripe.net/Infrastructure/Urls.cs
```diff
@@ -36,6 +36,8 @@ internal static class Urls
 
         public static string Subscriptions => BaseUrl + "/subscriptions";
 
+        public static string SubscriptionItems => BaseUrl + "/subscription_items";
+
```
jaymedavis: I am working towards eliminating this file. The urls are going directly in the service - see the StripeSubscriptionItemService

jaymedavis:
Sorry if this was confusing the other day - I have already implemented this feature. The only thing I seem to be missing from here is the link to subscription items on StripeSubscription

src/Stripe.net/Infrastructure/Urls.cs
TVNielsen: Ok, yea great idea, one place less to maintain :)

PR #789, opened 2017-03-18
Migrates to VS 2017 (moves from project.json and xproj to cproj files)

dusda:
I couldn't get this to build in vs2017, since .net core no longer supports xproj or project.json files. Additionally, using 'dotnet migrate', a tool MS made to move those project types back to csproj, breaks if you do not specify a version number on your frameworkAssemblies section of project.json. frameworkAssemblies is for dlls you want to load from the GAC, so I just looked up all of them using 'gacutil -l Name' and changed all of the wildcard versioning to specific versions. Then the migration worked and I could build the project.


jaymedavis:
thank you - I tried to do this with no luck before. I'll try your steps.

PR #820, opened 2017-04-10
Invoice subscription items

kspearrin:
This PR adds support for the missing `subscription_items` parameters for https://stripe.com/docs/api#invoice_lines and https://stripe.com/docs/api#upcoming_invoice .

You can see it's implementation in action here: https://github.com/bitwarden/core/commit/fa565f46c61a9464b51384a2397d4672afb60531#diff-6e37b3164a3aea7da3eb22f74e3b4f9dR400

Note that I changed the `subscription_items*` keywords to `subscription_items_array*` in the middleware plugins as to not be confused with properties that are identified in the actual API spec as `subscription_items`. If you prefer some other approach here let me know.

Also, note that I did not add any tests for this new functionality. Feel free to fill in any gaps you would like covered for test cases. I did, however, test this with my implementation linked above and it works.

### DIFF
--- a/src/Stripe.net/Infrastructure/Middleware/ParserPlugin/InvoiceSubscriptionItemPlugin.cs
+++ b/src/Stripe.net/Infrastructure/Middleware/ParserPlugin/InvoiceSubscriptionItemPlugin.cs
@@ -5,13 +5,13 @@ using Newtonsoft.Json;
 
 namespace Stripe.Infrastructure.Middleware
 {
-    internal class SubscriptionItemPlugin : IParserPlugin
+    internal class InvoiceSubscriptionItemPlugin : IParserPlugin
     {
         public bool Parse(ref string requestString, JsonPropertyAttribute attribute, PropertyInfo property, object propertyValue, object propertyParent)
         {
-            if (attribute.PropertyName != "subscription_items_array") return false;
+            if (attribute.PropertyName != "subscription_items_array_invoice") return false;
 
-            var items = ((List<StripeSubscriptionItemOption>) property.GetValue(propertyParent, null));
+            var items = ((List<StripeInvoiceSubscriptionItemOptions>) property.GetValue(propertyParent, null));
 
             var itemIndex = 0;
             foreach (var item in items)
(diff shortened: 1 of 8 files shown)


kspearrin:
P.S. Any luck on getting a release for this any time soon? I'm stuck with local builds for now since I need this functionality.

src/Stripe.net/Infrastructure/Middleware/ParserPlugin/InvoiceSubscriptionItemPlugin.cs
```diff
@@ -0,0 +1,40 @@
...
+                {
+                    var value = prop.GetValue(item, null);
+                    if (value == null) continue;
+
+                    // it must have a json attribute matching stripe's key, and only one
+                    var attr = prop.GetCustomAttributes<JsonPropertyAttribute>().SingleOrDefault();
+                    if (attr == null) continue;
+
+                    RequestStringBuilder.ApplyParameterToRequestString(ref requestString,
+                        $"subscription_items[{itemIndex}][{attr.PropertyName}]", value.ToString());
```
jaymedavis: was this little blurb of text the only difference in this plugin? I need to start thinking about merging these. This makes three that don't differ much if so.

src/Stripe.net/Infrastructure/Middleware/ParserPlugin/InvoiceSubscriptionItemPlugin.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Stripe.Infrastructure.Middleware
+{
+    internal class InvoiceSubscriptionItemPlugin : IParserPlugin
+    {
+        public bool Parse(ref string requestString, JsonPropertyAttribute attribute, PropertyInfo property, object propertyValue, object propertyParent)
+        {
+            if (attribute.PropertyName != "subscription_items_array_invoice") return false;
+
+            var items = ((List<StripeInvoiceSubscriptionItemOptions>) property.GetValue(propertyParent, null));
```
kspearrin: This type here as well.

kspearrin:
Hah, I started with a "plugin" suffix!

jaymedavis:
will go out with the next bump - thanks, good work on this one. I need to put more thought into how to solve the problem. Right now, it's probably a base class for the item plugins - cast to that where you changed the type in the plugin, I think the reflection would still work. If not, could always take a <T>. No reason to hold this up though 👍

PR #823, opened 2017-04-12
payouts and transfers

jaymedavis:


davidxcheng:
Nice, looked for this just two days ago and now it's already added - good job 👍

PR #839, opened 2017-04-25
converted the StripeBalanceService List function to return a StripeList

jaymedavis:


nicodenzl:
It seems to work ok, except that StripeBalanceTransactionListOptions should still have the IncludeTotalCount added:
[JsonProperty("include_total_count")]
        public bool IncludeTotalCount { get; set; } = true;

jaymedavis:
Thanks!

jaymedavis:
I just tried adding this, but it didn't work. I don't think TotalCount or HasMore works unless you are in the context of pagination. It wasn't listed on the stripe docs either :(

nicodenzl:
What failed? I literally copy-pasted that property in from StripeChargeListOptions.

My calling code is:
 public void ListAllBalances()
        {
            string lastObjectId = null;
            StripeList<StripeBalanceTransaction> response = null;
            do
            {
                response = GetBalances(lastObjectId);
                Console.WriteLine("Total: {0}, Count: {1}", response.TotalCount, response.Data.Count);
                lastObjectId = response.Data.LastOrDefault()?.Id;
            }
            while (response.HasMore);


        }

        private StripeList<StripeBalanceTransaction> GetBalances(string startingAfter)
        {
            var listOptions = new StripeBalanceTransactionListOptions
            {
                Limit = 100,
                Created = new StripeDateFilter { GreaterThanOrEqual = DateTime.UtcNow.Date },
                StartingAfter = startingAfter
            };
            StripeList<StripeBalanceTransaction> response = balanceService.List(listOptions);
            return response;
        }

jaymedavis:
https://github.com/jaymedavis/stripe.net/compare/add_total_count_to_balance_list?expand=1

My guess is it's just not supported there and it is on charges - because they are more popular by far. You can see my changes and test here.

jaymedavis:
Ok - followed up with Stripe, and it is supported. My problem in the test is I wasn't passing the options. Even though it defaulted to true, I still had to pass an empty instance of the options.

With that said, be aware that the default being true is going away. This appears to be a performance concern on stripes end, so they want it explicitly set to true. 

I'll get this pushed out today.

nicodenzl:
Awesome, thank you very much! Yeah, having it set to false by default probably makes sense.

PR #869, opened 2017-05-08
Managed accounts japan

carlosboeing:
Hey @jaymedavis,

I'm using Stripe Connect for Japanese accounts and need some missing properties in StripeLegalEntity which are the kana and kanji representation of first name, last name, business name, address and personal address. I've added those properties to it and I've also added Town to StripeAddress which is used for kana and kanji addresses. For that, I've renamed the existing property CityOrTown to City and added a new property Town which is used only for Japan. It's a breaking change, but I didn't want to end up with a property called CityOrTown and another one called Town because that was confusing. Alternatively, I could create a new class called StripeAddressJapan which contains Town and it's only used by kana and kanji addresses. That way I don't have to touch StripeAddress and would keep backwards compatibility. Let me know what you prefer.

Regards,
Carlos


jaymedavis:
There are some other places that need to be checked in regard to addresses as well, however. I copied CityOrTown from the Stripe official go bindings.

carlosboeing:
Yeah, town is only used for Japanese addresses in address_kana or address_kanji (https://stripe.com/docs/api/curl#account_object-legal_entity-address_kana-town). These are required verification fields for Japanese connected accounts. I'm happy to add the [Obsolete] attribute if you want to go ahead with this solution.

carlosboeing:
Here's an example of Japanese address from stripe:
```javascript
{
  "country": "JP",
  "legal_entity": {
    "address_kana": {
      "country": "JP", // 2-letter country code
      "postal_code": "1500001", // Zip/Postal Code
      "state": "ﾄｳｷﾖｳﾄ", // Prefecture
      "city": "ｼﾌﾞﾔ", // City/Ward
      "town": "ｼﾞﾝｸﾞｳﾏｴ 3-", // Town/cho-me
      "line1": "27-15", // Block/Building number
      "line2": "FLAG 3A", // Building details
    },
    "address_kanji": {
      "country": "JP", // 2-letter country code
      "postal_code": "１５００００１", // Zip/Postal Code
      "state": "東京都", // Prefecture
      "city": "渋谷区", // City/Ward
      "town": "神宮前　３丁目", // Town/cho-me
      "line1": "２７－１５", // Block/Building number
      "line2": "ＦＬＡＧ　３Ａ", // Building details
    }
  }
}
```

jaymedavis:
If that's all you found, then no worries - just leave it as is. I need to kill of recipients soon anyway, so the breaking change is no big deal. :)

jaymedavis:
Thanks @carlosboeing!

PR #876, opened 2017-05-11
Added expandable properties to StripePayout

carlosboeing:
Hey @jaymedavis,

As part of the transfer payout split (from April 6, 2017), some properties on StripePayout and StripeBalanceTransactionListOptions needed to be updated/renamed. I've also added 3 expandable properties to StripePayout. For the expandable property Destination (bank account or debit card the payout was sent to), I've created a new class called StripeExternalAccount which is exactly like Source, but I had to add the custom JsonConverter at the class level otherwise `ExpandableProperty.Map` wouldn't work. I did that because I didn't want to touch Source. By the way, this is an existing bug with expandable properties and custom JsonConverter. For instance, `StripeCustomer.DefaultSource` never gets populated because the attribute `[JsonConverter(typeof(SourceConverter))]` is set on the property DefaultSource and not on the class Source. This would be fixed if we move the attribute `[JsonConverter(typeof(SourceConverter))]` to the class level, but I haven't checked the side-effects of that. Let me know if you would like to do that and also if you want to merge StripeExternalAccount and Source. I saw your comment on the StripeExternalAccountService which is unfinished, so let me know what you have in mind for that.

Cheers,
Carlos


src/Stripe.net/Entities/StripePayout.cs
```diff
@@ -18,13 +18,40 @@ public class StripePayout : StripeEntityWithId
         [JsonConverter(typeof(StripeDateTimeConverter))]
         public DateTime ArrivalDate { get; set; }
```
jaymedavis: thanks, looks perfect

src/Stripe.net/Services/Balance/StripeBalanceTransactionListOptions.cs
```diff
@@ -13,12 +13,12 @@ public class StripeBalanceTransactionListOptions : StripeListOptions
         [JsonProperty("currency")]
         public string Currency { get; set; }
```
jaymedavis: wasn't completely clear on this one - thanks!

src/Stripe.net/Services/Payouts/StripePayoutService.cs
```diff
@@ -9,6 +9,10 @@ public class StripePayoutService : StripeBasicService<StripePayout>
     {
```
jaymedavis: I've had about 5 new ideas for how to handle expandables so there isn't as much cookie cutter. My fear is this is an easy way for the developer to use it, though.

jaymedavis:
note: read my comments on the files changed tab, it wont make much sense on the conversation tab :)

carlosboeing:
Hey @jaymedavis, I also prefer to use Source everywhere to be honest, it's a it confusing to have ExternalAccounts when they actually are the same thing as Source. Check out my last commit (907f36f8c3dae0f414e9bca909611adb5845713c). By moving the `[JsonConverter(typeof(SourceConverter))]` to the class `Source` and removing it from the properties, it will make sure it's always called. This would solve our problem with expandable and it will also work with `StripeList<Source>` so there won't be the need for `SourceListConverter` anymore. I've tested all the places where Source is used and it works fine now. The classes that have properties with Source are: StripeCharge, StripeAccount, StripeCustomer and StripePayout. Let me know if you are happy with this solution so we can go ahead and I can do a bit of clean up (remove SourceListConverter, StripeExternalAccountConverter and StripeExternalAccount).

src/Stripe.net/Infrastructure/JsonConverters/StripeExternalAccountConverter.cs
```diff
@@ -21,20 +21,26 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
```
carlosboeing: Yeah, this wasn't being used that's why I refactored, but if we go with Source everywhere, we don't need that either.

src/Stripe.net/Services/Payouts/StripePayoutService.cs
carlosboeing: I like the way we use expandables right now, I don't think that needs to be changed. The only issue is the expandable property with a JsonConverter (e.g: SourceConverter) because that would never be called, since the deserialization happens on the internal property and the object is set in the Map method of the ExpandableProperty class.

For instance, with StripeCustomer (below), the property DefaultSource was always null because we deserialize the Source like json into the InternalDefaultSource which receives the value of "default_source" from the payload, then we handle other 2 properties (DefaultSourceId and DefaultSource) inside the mapper ExpandableProperty<Source>.Map by setting them individually after deserializing the object into T. If we keep the JsonConverter attribute in the property, that won't be called when settings the properties in the mapper. The way to fix that is by moving the JsonConverter to the class level of Source (my last commit: 907f36f8c3dae0f414e9bca909611adb5845713c)

```
public string DefaultSourceId { get; set; }

[JsonIgnore]
[JsonConverter(typeof(SourceConverter))]
public Source DefaultSource { get; set; }

[JsonProperty("default_source")]
internal object InternalDefaultSource
{
	set
	{
		ExpandableProperty<Source>.Map(value, s => DefaultSourceId = s, o => DefaultSource = o);
	}
}
```

jaymedavis:
I checked out the `when_creating_a_customer_with_a_defaultcard_expanded` test, and noticed the id was populated and the type of card was populated, but surely enough - no card object. My test facts weren't good enough. Nice catch.

jaymedavis:
This looks perfect! Not only does it fix it, it's better, imo 👍

carlosboeing:
Cool. So let me do a bit of clean up before we merge.

src/Stripe.net/Services/Payouts/StripePayoutService.cs
jaymedavis: it's really awesome that this handles the StripeList<Source> too and lets us kill a class entirely.

jaymedavis:
looks great - i'll merge this in tomorow 👍

jaymedavis:
For some reason, my tests will run (up to 333 of them), then will just bail out with 

```
An exception occurred while invoking executor 'executor://machine.vstestadapter/': Could not load file or assembly 'System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. The system cannot find the file specified.
========== Run test finished: 780 run (0:03:44.4956702) ==========
```

Did the test finish for you, or did you run them all? Switching back to master works fine. This makes no sense with the file diff on this PR, naturally :)

jaymedavis:
Update:

`when_serializing_a_stripe_charge` is the issue. WriteJson on the converter is throwing an exception on the ```JObject.FromObject(value);```. I generally tell people to avoid serializing the classes on their own, since they wont deserialize properly (due to things like EpochTimeConverters, etc), but it is something people do, so I'd like to fix it if possible. 

Besides this, it seems everything else is running ok.

jaymedavis:
See the [commit message and description] here. It was a circular reference for trying to serialize the same thing over and over. Needed little of nothing to fix. 

Nice use of half a day, at least it should be fixed. Waiting on build server 👍

jaymedavis:
Hey @carlosboeing - closing this in replacement of #878 -- the small fixes with the testing lib.

Thanks again for your great work. 👍

carlosboeing:
Hey @jaymedavis, thanks for the fixes. Nice finding with the circular reference, that was a tricky one. 👍

PR #896, opened 2017-05-26
Make build settings work on Mac with the dotnet command line tool

fred-stripe:
This involves changing the target to netcoreapp1.1 & including an implicit NetStandard
version in each test project's build settings.

cc @andrewnelder

### DIFF
--- a/src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
+++ b/src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
@@ -8,7 +8,7 @@
     <RuntimeIdentifiers>win10-x64;win81-x64;osx.10.12-x64</RuntimeIdentifiers>
     <RuntimeFrameworkVersion>1.1.1</RuntimeFrameworkVersion>
     <PackageTargetFallback>$(PackageTargetFallback);dotnet5.6;portable-net45+win8</PackageTargetFallback>
-    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
+    <NetStandardImplicitPackageVersion>1.2</NetStandardImplicitPackageVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Stripe.net.Tests/Stripe.net.Tests.csproj
+++ b/src/Stripe.net.Tests/Stripe.net.Tests.csproj
@@ -2,11 +2,12 @@
 
   <PropertyGroup>
     <VersionPrefix>0.0.0</VersionPrefix>
-    <TargetFramework>net46</TargetFramework>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
     <AssemblyName>Stripe.net.Tests</AssemblyName>
     <PackageId>Stripe.net.Tests</PackageId>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <RuntimeIdentifiers>win</RuntimeIdentifiers>
+    <NetStandardImplicitPackageVersion>1.2</NetStandardImplicitPackageVersion>
   </PropertyGroup>
 
   <ItemGroup>
@@ -20,7 +21,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.*" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net46' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.1' ">
     <EmbeddedResource Include="fileuploads\test_data\logo.png" />
     <EmbeddedResource Include="infrastructure\test_data\card.json" />
     <EmbeddedResource Include="infrastructure\test_data\charge.json" />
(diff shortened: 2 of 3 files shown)


fred-stripe:
Hm, not sure why that was a build success when it clearly didn't have an API key. Odd.

jaymedavis:
Yep, this clearly should have failed. Probably because it's powershell and we're being punished.

jaymedavis:
will address in #897 - I don't think there's anything specific to this repo that _requires_ the powershell script. hoping to use native appveyor.xml alone 👍

src/Stripe.net.Tests/Stripe.net.Tests.csproj
```diff
@@ -20,7 +21,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.*" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net46' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.1' ">
```
anelder-stripe: Why is this even necessary if we've changed the target framework on line 5?

src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
```diff
@@ -8,7 +8,7 @@
     <RuntimeIdentifiers>win10-x64;win81-x64;osx.10.12-x64</RuntimeIdentifiers>
     <RuntimeFrameworkVersion>1.1.1</RuntimeFrameworkVersion>
     <PackageTargetFallback>$(PackageTargetFallback);dotnet5.6;portable-net45+win8</PackageTargetFallback>
-    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
+    <NetStandardImplicitPackageVersion>1.2</NetStandardImplicitPackageVersion>
```
anelder-stripe: Should this be 1.1 based on https://docs.microsoft.com/en-us/dotnet/standard/library?

src/Stripe.net.Tests/Stripe.net.Tests.csproj
```diff
@@ -2,11 +2,12 @@
 
   <PropertyGroup>
     <VersionPrefix>0.0.0</VersionPrefix>
-    <TargetFramework>net46</TargetFramework>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
     <AssemblyName>Stripe.net.Tests</AssemblyName>
     <PackageId>Stripe.net.Tests</PackageId>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <RuntimeIdentifiers>win</RuntimeIdentifiers>
+    <NetStandardImplicitPackageVersion>1.2</NetStandardImplicitPackageVersion>
```
anelder-stripe: Same as above.  1.1?

https://docs.microsoft.com/en-us/dotnet/standard/library

src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
jaymedavis: The cipher suites available may be an issue? I believe TLS 1.2 is supported all around here, but it has to be specified like [this](https://stackoverflow.com/questions/39560249/servicepointmanager-in-asp-net-core) or  the `ServicePointManager` that is being used now. Worth more investigation.

src/Stripe.net.Tests/Stripe.net.Tests.csproj
jaymedavis: Same answer as above - I bring this up only because I'm assuming you're thinking stripe.net should target the same standard.

src/Stripe.net.Tests/Stripe.net.Tests.csproj
jaymedavis: Good catch. It was multiple frameworks before and two different sections (the other removed) 👍

anelder-stripe:
Let's hold off on merging this into master until I get a bit more feedback from some community developers.  I should have some better answers by Wednesday of next week.  Until then, we can just fork @fred-stripe's branch.

jaymedavis:
It seems that the netstandard1.1 will be fine for all consumers. The problem with using this on the previous platform (by means of `Profile111` and the like), was it took away some newer language features. With netstandard though, it's just a dependency you pull in to your project - so it doesn't limit you on language features. I imagine there are still some things you cannot do, but I don't think they apply to stripe.net or they can be changed.

I tested by creating an old school net45 forms app. I noticed that when the dll was imported to the project directly, it didn't have support for pulling down the dependency. I'm hoping this will be resolved via nuget, as I just grabbed the dll from the bin. I investigated other repos (such as newtonsoft), and all the ones I found still support the net45 moniker. I'm not entirely sure why? Legacy maybe?

I don't see why there couldn't be a single netstandard1.1 build that's used for net45 as well, assuming we can resolve the dependency with nuget still.

I would like to phone a couple friends - @xt0rted and @flagbug for opinions - tl;dr;  why can't the library just remove everything net45 specific and only be a netstandard1.1 library? Thanks in advance for anything you can offer!!

xt0rted:
I actually haven't ventured down the .NET Core path due to tooling, the project system, NuGet, entity framework, and lots of dependencies not fully supporting it yet (most are Microsoft's too 😕).

All of the projects I've looked at contain at least `net45,netstandard1.1` or some combination of the two types. This is most likely due to missing APIs in .NET Core and using only `netstandard*` may work across platforms just fine depending on the project.

flagbug:
@jaymedavis There's absolutely nothing stopping us from throwing out `NET45`, except maybe the one line of code that I saw somewhere in the codebase, that reads the API key from the `ConfigurationManager` - but consumers of this library can just do this themselves and then pass the API key into the individual service constructors

jaymedavis:
Any new developments on this? I believe all of the net45 stuff can go away - the only caveat will be advising users how to read their key if they previously used web/app.config. There will need to be a little work in the Client.cs for user agent data, but we can fix that. 👍 

@andrewnelder @fred-stripe

andrewnelder:
@jaymedavis - Yes, I've found some conflicts that actually prevent us from targeting `netstandard1.1`.  Specifically, the cryptography include.  Which means, we may need to target `netstandard1.2` or `netstandard1.3`.  Since that's potentially breaking for some users, I'd like to continue to support the `net45` integration and just clean up the configuration files and the test runners we're deploying to CI.  I'll submit a PR soon with my changes.  I'm very sorry it's taken so long.

andrewnelder:
@jaymedavis - Okay.  So the situation seems to be this.  We can set a target framework of `netstandard1.1` in the core project; however, if we attempt to test against `net45` it'll break.  I'm working around that meow.

jaymedavis:
Awesome. As long as the test projects target netstandard, shouldn't have any issues with the existing items targeting net45, sounds good to me.

img_1229

andrewnelder:
@jaymedavis - I've submitted a second build.  We can still use it to test on mac.  Essentially, when running tests in MacOSX, you would do the following:

```
dotnet test /p:TargetFramework=netcoreapp1.1
```

andrewnelder:
@jaymedavis - I don't think this is relevant anymore, correct?  Can we close this?

PR #898, opened 2017-05-27
[WIP] Support API version 2017-05-25

remi-stripe:
Stripe's latest API version, [2017-05-25](https://stripe.com/docs/upgrades#2017-05-25) introduced some breaking changes that this PR aims to support:

- [x] replace `managed` by `type` on the Account resource.
- [x] replace reference to managed accounts by Custom accounts and standalone accounts by Standard accounts.
- [x] `user_id` property on the Event object is now called `account`.
- [x] `request` property on the Event object is now a hash with `request[id]` and `request[idempotency_key]`.
- [ ] confirm that the change is for `previous_attributes` on the Event object is handled gracefully.

### DIFF
--- a/src/Stripe.net/Services/Account/StripeAccountCreateOptions.cs
+++ b/src/Stripe.net/Services/Account/StripeAccountCreateOptions.cs
@@ -7,7 +7,7 @@ namespace Stripe
         [JsonProperty("country")]
         public string Country { get; set; }
 
-        [JsonProperty("managed")]
-        public bool? Managed { get; set; }
+        [JsonProperty("type")]
+        public string Type { get; set; }
     }
 }
--- a/readme.md
+++ b/readme.md
@@ -125,15 +125,15 @@ Accounts
 --------
 ### Creating an account
 
-When creating an account, you can create a standalone or managed account. Standalone accounts are managed by Stripe and the account owner directly. Managed accounts are handled by your platform. See the Stripe documentation for more information.
+When creating an account, you can create a Standard or Custom account. Standard accounts are managed by Stripe and the account owner directly. Custom accounts are handled by your platform. See the Stripe documentation for more information.
 
 Since Stripe returns `ExternalAccounts` as a single array (contains StripeCard's and/or StripeBankAccount's), that type is a dynamic StripeList. These are split up as `ExternalCards` and `ExternalBankAccounts` for your convenience.
 
 ```csharp
 	var account = new StripeAccountCreateOptions();
-	account.Email = "jayme@yoyoyo.com"  // this is required if it is not a managed account. the user is emailed on standalone accounts,
-	                                    // it's only used for reference on managed accounts
-	account.Managed = false;            // set this to true if you want a managed account (email is not required if this is set to true)
+	account.Email = "jayme@yoyoyo.com"  // this is required for a Standard account as the owner gets emails from Stripe.
+	                                    // it's only used for reference on Custom accounts
+	account.Type = "custom";            // this can be "custom" or "standard". Express accounts are not created via the API.
 
 	// a few optional settings
 	account.Country = "US"                                 // defaults to your country
--- a/src/Stripe.Tests.XUnit/accounts/_cache.cs
+++ b/src/Stripe.Tests.XUnit/accounts/_cache.cs
@@ -20,7 +20,7 @@ namespace Stripe.Tests.Xunit
                 DeclineChargeOnAvsFailure = false,
                 DeclineChargeOnCvcFailure = true,
                 DefaultCurrency = "usd",
-                Managed = true,
+                Type = "custom",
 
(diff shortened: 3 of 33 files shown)


remi-stripe:
@jaymedavis I was wondering if `StripeEventRequest` should inherit from `StripeEntityWithId` since it has an `Id` property too. I don't think it's a good id because it's not the resource's id and just the id of the request (which is the upper hash) but let me know what you think!

src/Stripe.net/Services/Account/StripeAccountCreateOptions.cs
```diff
@@ -7,7 +7,7 @@ public class StripeAccountCreateOptions : StripeAccountSharedOptions
         [JsonProperty("country")]
         public string Country { get; set; }
 
-        [JsonProperty("managed")]
-        public bool? Managed { get; set; }
+        [JsonProperty("type")]
+        public string Type { get; set; }
```
anelder-stripe: I'm a bit hesitant about the use of the attribute `Type`.  To avoid future namespace issues, it might make sense to change this to something like `AccountType`.  I'm pretty soft on this feedback though, as I don't know if it could have any weird namespace conflicts.

remi-stripe: @anelder-stripe I thought the same thing when we named the property internally but Type is used elsewhere for example on [Balance Transaction](https://github.com/jaymedavis/stripe.net/blob/master/src/Stripe.net/Entities/StripeBalanceTransaction.cs#L40) so I think we're good on this one.

jaymedavis: No worries with conflicts here (.net uses `GetType` and `typeof` as well) . You'll see various places i've had the same reservations for properties like this, Id, and Name. Since this is in the context of an AccountCreateOptions, I think it's intuitive that the type is for Account. 👍

jaymedavis:
couple small nits: https://github.com/jaymedavis/stripe.net/tree/pr/898 fix on the top two commits.

1.  Instead of passing "custom", I created a ```StripeAccountType``` class, then referenced this class in the built-in documentation for setting type. This will allow users to set it to ```.Custom``` or ```.Standard``` without worry for a string being spelled incorrectly or an invalid type.

2. For the tests, most of them use the Should syntax, vs Assert. Changed this just for consistency.

carlosboeing:
Hey guys,

Last Friday we upgraded one of our environments to use the latest Stripe API and I borrowed a few of the changes submitted by @remi-stripe in this PR. Overall, everything seems to be working well so far, the only thing I've noticed is a problem with the deserialization of the payload for old webhooks due to the changes in the "Request" property of "StripeEvent".

Webhooks (prior to May 23, 2017) will have the "request" property as a simple string which contains the id of the request, whereas new webhooks (after May 23, 2017) will have the "request" property as a hash containing the "id" and "idempotency_key". So, when we deserialize old webhooks it, will throw an exception as it can't convert string into StripeEventRequest.

To make it backwards compatible, we can either write a custom JsonConverter for StripeEventRequest or we can use something similar to what we have for expandable properties:

```
public string RequestId { get; set; }

[JsonIgnore]
public StripeEventRequest Request { get; set; }

[JsonProperty("request")]
internal object InternalRequest
{
	set
	{
		ExpandableProperty<StripeEventRequest>.Map(value, s => RequestId = s, o => Request = o);
	}
}
```

Let me know your thoughts.

Cheers,
Carlos

remi-stripe:
@jaymedavis what do you think of adding Carlos' changes? I'm not sure how you try and handle backwards compatibility for events created on older versions of the API.

jaymedavis:
Sorry about that - this question got away from me!

Perhaps a better name for `ExpandableProperty`  is `StringOrObject`. It's a little more generic and offers a wider purpose.  With that said, I think this solution works fine. If we add a custom converter, we are just going to steal the code from that class anyway.

Clever solution @carlosboeing!

remi-stripe:
@jaymedavis not sure I follow what you mean exactly. Can you maybe add this specifically? We're getting a few emails from users with that problem so it'd be great to merge the PR soon :)

jaymedavis:
@remi-stripe yes, I will work on this today. 

I meant that the name ExpandableProperty doesn't fit here, but it still works. I will fix and add to my other changes on pr/898.

Regarding your last todo item on previous_attributes, this is a dynamic object so pretty much anything will work. It's not the most developer friendly solution, but I believe various types and values can be part of this property? Let me know if there's a specific case you want me to checkout here.

jaymedavis:
this is on master now - going to release after fixing a couple of other items (10.0)

* ExpandableProperty was renamed to StringOrObject
* Instead of passing the strings as "custom", it's now ```StripeAccountType.Custom``` and the xml documentation nudges the user towards using this class so they can't make a mistake with an invalid type
* Went ahead and bumped the api up to `2017-06-05`

PR #905, opened 2017-05-31
Fixed bug: NullReferenceException in SourceConverter.ReadJson

carlosboeing:
It happens when Source has been deleted (e.g: payout with deleted bank account) as there's no "object" attribute present.

For instance, in a payout with a bank account the destination object should be something like this (this could also be a card, therefore we use Source):

```
"destination": {
    "id": "ba_xxxxxx",
    "object": "bank_account",
    "account": "acct_xxxxxxxx",
    "account_holder_name": null,
    "account_holder_type": null,
    "bank_name": "BANK NAME",
    "country": "US",
    "currency": "usd",
    "default_for_currency": true,
    "fingerprint": "xxxxxxxxxx",
    "last4": "6789",
    "metadata": {},
    "routing_number": "110000000",
    "status": "new"
  }
```

Now the issue is that Stripe is returning a completely different payload when destination (bank account or card) associated to the payout has been deleted:

```
"destination": {
    "deleted": true,
    "id": "ba_xxxxxx",
    "currency": "usd"
  }
```

This was throwing a null reference exception during deserialization in the SourceConverter.

### DIFF
--- a/src/Stripe.net/Infrastructure/JsonConverters/SourceConverter.cs
+++ b/src/Stripe.net/Infrastructure/JsonConverters/SourceConverter.cs
@@ -29,13 +29,13 @@ namespace Stripe.Infrastructure
                 Id = incoming.SelectToken("id").ToString()
             };
 
-            if (incoming.SelectToken("object").ToString() == "bank_account")
+            if (incoming.SelectToken("object")?.ToString() == "bank_account")
             {
                 source.Type = SourceType.BankAccount;
                 source.BankAccount = Mapper<StripeBankAccount>.MapFromJson(incoming.ToString());
             }
 
-            if (incoming.SelectToken("object").ToString() == "card")
+            if (incoming.SelectToken("object")?.ToString() == "card")
             {
                 source.Type = SourceType.Card;
                 source.Card = Mapper<StripeCard>.MapFromJson(incoming.ToString());


jaymedavis:
nice catch. I wonder if it might make sense to add a StripeDeleted reference to Source, and then check to see if deleted is present and populate that object. Would this be useful in your implementation, or is it easy enough to just say it's null and carry on?

Just a thought - this doesn't need to block the PR.

carlosboeing:
I actually thought about making Source inherit from StripeDeleted, but then I didn't want to change too much because Source is used in many places. At the moment I'm just checking for null, if Source has Id but no bank account or card details I know it's been deleted. Not the most beautiful solution but it works.

If you want I can make a change to include StripeDeleted. 

There's also another issue when deserializing a deleted external account into Source: we have no "object" information, so we don't know if it's a card or bank account and the Source.Type defaults to Card. Maybe I should make Type property nullable and set it to null when that happens.

Any thoughts?

jaymedavis:
You could just add a new value in the enum SourceType (as the first entry) called `Deleted`. That would take care of it defaulting to card. Also, add ```public StripeDeleted Deleted { get; set; }``` to Source, then modify the source converter to check for ```SelectToken("deleted")``` and ```source.Deleted = Mapper<StripeDeleted>.MapFromJson(incoming.ToString())``` .... I think this would work for all cases?

jaymedavis:
I pulled down your branch and made a few changes to reflect above - https://github.com/jaymedavis/stripe.net/compare/pr/905?expand=1

The idea is to merge this in tomorrow and create 10.0 along with the API changes. Review appreciated.

carlosboeing:
It looks good to me. Just about the SourceType enum, you probably don't
need to make Deleted as the first (default) value of the enum since you
always set within the JsonConverter when "deleted" attribute is present.


On 6 Jun. 2017 03:53, "Jayme Davis" <notifications@github.com> wrote:

I pulled down your branch and made a few changes to reflect above -
https://github.com/jaymedavis/stripe.net/compare/pr/905?expand=1

The idea is to merge this in tomorrow and create 10.0 along with the API
changes. Review appreciated.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/jaymedavis/stripe.net/pull/905#issuecomment-306359131>,
or mute the thread

PR #919, opened 2017-06-08
Migrate support to netstandard1.1

andrewnelder:
The goal of this build is to target net45+ and netcoreapp1.*.
Unfortunately, while the project is building in netstandard1.1, I cannot
target net45 in the tests without mspec breaking.

### DIFF
--- a/src/Stripe.net.Tests/Stripe.net.Tests.csproj
+++ b/src/Stripe.net.Tests/Stripe.net.Tests.csproj
@@ -2,25 +2,45 @@
 
   <PropertyGroup>
     <VersionPrefix>0.0.0</VersionPrefix>
-    <TargetFramework>net46</TargetFramework>
+
+    <TargetFrameworks>netcoreapp1.1</TargetFrameworks>
     <AssemblyName>Stripe.net.Tests</AssemblyName>
     <PackageId>Stripe.net.Tests</PackageId>
-    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
-    <RuntimeIdentifiers>win</RuntimeIdentifiers>
+    <DebugType>portable</DebugType>
+    <OutputType>Library</OutputType>
+    <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
+
+    <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
+    <GenerateAssemblyDescriptionAttribute>false</GenerateAssemblyDescriptionAttribute>
+    <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
+    <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
+    <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
+    <GenerateAssemblyCopyrightAttribute>false</GenerateAssemblyCopyrightAttribute>
+    <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
+    <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
+    
   </PropertyGroup>
 
+  <!-- Include the Stripe.net project -->
   <ItemGroup>
     <ProjectReference Include="..\Stripe.net\Stripe.net.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="Machine.Specifications" Version="0.11.0" />
-    <PackageReference Include="Machine.Specifications.Runner.VisualStudio" Version="2.2.3" />
-    <PackageReference Include="Machine.Specifications.Should" Version="0.11.0" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.*" />
+  <!-- Include latest NetCore -->
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.0' ">
+    <PackageReference Include="Microsoft.NetCore.App" Version="1.*" type="platform" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net46' ">
+  <!-- Include test dependencies -->
+  <ItemGroup>
+    <PackageReference Include="Machine.Specifications" Version="0.11.0" />
+    <PackageReference Include="Machine.Specifications.Should" Version="0.11.0" />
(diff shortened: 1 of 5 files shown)


andrewnelder:
@jaymedavis - I'm aware this is breaking.  I'm unable to successfully run tests from either mspec or xunit against net45.  Would you be able to take a look?  net46 works fine in both, but that's a bit weird for me.  I suspect there is some kind of weird dependency problem in the mspec tests for net45, but I was unable to figure it out.  Guidance would be awesome.  ^.^

src/Stripe.net.Tests/Stripe.net.Tests.csproj
```diff
@@ -31,22 +52,4 @@
     <EmbeddedResource Include="infrastructure\test_data\subscription.json" />
   </ItemGroup>
```
jaymedavis: My gut reaction is this is the issue. If it's not targetting netstandard, it will need to pull in the assemblies from the .net framework it needs to run. I'll check this out - I don't think we need all of these anyway.

jaymedavis:
So here is what I've dug up so far:

First, make sure you are restoring the project to the right target framework:  ```dotnet restore /p:TargetFramework="net45"```. Next, you have to install (via nuget) the [old mspec adapter](https://github.com/machine/dotnet-test-mspec). However, I haven't figured out how you'd override dotnet test to use that adapter without ruining harmony with the other. It looks like the solution is best handled by another project file. 

With that said, I don't think this route is best. The only things stopping net45 from going away entirely is setting the SecurityProtocol and a small config issue. The first is attainable by offering a line of code to the user in the event TLS 1.2 doesn't negotiate, or possibly there is another netstandard solution.

src/Stripe.net.Tests/Stripe.net.Tests.csproj
jaymedavis: shouldn't trust your gut 👎

jaymedavis: technically it was half right - it needs to pull in the assemblies (dotnet restore) :)

anelder-stripe:
@jaymedavis - I think we realized the same thing.   ^.^  Good.  I'm glad I'm not crazy.  I think my take on this is:

While I would like `net45` to disappear.  We're not currently able to accurately represent which of your users is on what .NET Framework in the bindings.  My goal for the bindings would then be to continue to support `net45` as a `netstandard1.1` library, then collect usage information using a script (that microsoft provides) to identify the .NET Framework version for people using it.  Once we know who doesn't even need the `net45` dll, we can then decide to destroy it.  If there are a number of users dependent on it, I'd like to continue to target it for the time being; until we can convince our users to upgrade to a version that microsoft officially supports (ie. `net46+`).

jaymedavis:
> collect usage information using a script (that microsoft provides)

This would be great. From memory, I believe most of the data showed 4.0.0.x, which is the last strong named version (i think?). I spent awhile trying various tactics to get more information, but hit a wall.

> Once we know who doesn't even need the net45 dll, we can then decide to destroy it.

The thing is, nobody needs it. :) ... there are two parts in the codebase I mentioned above that the net45 library differs on from netstandard. This is just a couple bullet points in the release notes to make changes. We can also add these to the readme for new users. One of these might even be a non-issue and netstandard defaults to supporting the highest security.

I recognize the user agent data will show more net45 users, but all of them are compatible with netstandard1.1 (with tweaks listed above). You don't need to convince the users to upgrade, as it was designed with the purpose to be backwards compatible - doesn't matter if they are on 4.5 with 0 patches.

With that said, I still agree it's a good idea to get more information first (a script and/or fixing up Client.cs to get more data). The library as is hasn't tested net45 (for the last year or so) because it's covered under netstandard. To date, I haven't had any issues come up because something didn't work in net45 that worked in netstandard. 👍

anelder-stripe:
@jaymedavis - I hear ya.  If we don't think we can get `net45` tested, then I'm going to remove all of the reqs in `net45` in this revision and only set `netstandard1.1` as the target and we'll test against that (if we can).   I think there are similar compatibility issues with `netstandard1.1` as the target in tests.

jaymedavis:
Sounds good. I'll take a look at that build warning message that pops up too (is this the compatibility issue you're referring to?). It hasn't made a difference with any of the shipped libs, but it's annoying and needs to go away. 😿

anelder-stripe:
@jaymedavis - I'd bet money that build warning relates to the cryptography library used for event signatures.

jaymedavis:
@anelder-stripe ohhh.. that would make sense. It is new. Thanks.

xt0rted:
I just updated the [sendgrid .net client](https://github.com/sendgrid/sendgrid-csharp) to use the new project system and while poking around in there I tried to remove the `net452` target they had and only target `netstandard1.3`.

I then ran `dotnet pack` and added that to a .net 4.6 console app. Adding the package worked but it pulled in some 50+ NuGet packages. If i target both frameworks then it only pulls in json.net with it.

What benefit is there to removing the `net45` target? Personally I've tried to avoid .net standard only packages like this because of all the extra references since it gets really annoying trying to manage them all.

anelder-stripe:
@xt0rted - We're trying to supply a number of options to Stripe users.  We're not going to immediately remove `net45` (as I mentioned earlier).  We're testing to see which of our users are actually dependent on it before we potentially do that (as a long term goal).

What we're discussing now is actually making our tests (mspec and xunit) run against `net45` which seems to be a problem.

jaymedavis:
Thanks @xt0rted. Today, I have been battling the following warning error in the build:

```
warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed.
```

This is related to using `System.Security.Cryptography.Algorithms` for secure webhooks, so it's required. It compiles fine, appears everything as fine. This library is defined in the overall project as a reference, but it seems like net45 is missing something and it brings in 4.2, instead of 4.3? So if you know any kung-fu in this department, it's appreciated. I've been down many rabbit holes today. 😄

andrewnelder:
@jaymedavis - I made all of the changes, but now I'm getting weird errors.  Do you mind taking a look?  It looks like some weird three_d_secure OR json related issue.

jaymedavis:
I think upgrading the project to netstandard1.3 actually make sense now, since users are just going to target net45 behind that moniker (which we are now keeping). Seeing as that's the case, I don't think doing this matters much anymore. Are there any specifics of this PR you want to keep besides that?

The test from this PR has already been fixed from another PR.

andrewnelder:
@jaymedavis - I agree.  Thanks!  I think everything else I want to do is just cleanup, so we can do that after the fact.

PR #922, opened 2017-06-10
SourceType on StripeTransferCreateOptions  

ryn0:
This field is required so that a bank transfer can occur. Without this, the "source_type" defaults to "card", adding this allows setting the source type to "bank_account". There is no way to send bank transfers without this to my knowledge. See the Transfers API: https://stripe.com/docs/api#transfers


jaymedavis:
I see this on the response object, but it's not listed in the parameters for creating a transfer on the api docs. @remi-stripe - is this just missing from the docs?

remi-stripe:
@jaymedavis yes not sure why it's missing but it's definitely supported (and needed if you don't just charge cards)

PR #924, opened 2017-06-11
Adding support for the login_link resource for Express accounts

remi-stripe:
[Express accounts](https://stripe.com/docs/connect/express-accounts) are created via OAuth and not the API. A platform can let their user log into their dashboard by creating a login link via the [API](https://stripe.com/docs/api#create_login_link). This PR adds support for it.


remi-stripe:
Added async's methods, use a real id for the main account for Stripe.net. This should now be ready to review.

r? @jaymedavis

PR #930, opened 2017-06-16
build moved to powershell

jaymedavis:


jaymedavis:
this is good to go

cc @andrewnelder @fred-stripe

PR #931, opened 2017-06-16
removed some properties from invoice update that don't exist

jaymedavis:


jaymedavis:
These may have accidentally made it to the invoice update options - I don't see any reference to them on the documentation.

can you confirm these didn't exist? ready to go @stripe-fred @andrewnelder

brandur-stripe:
I wish I'd seen this before doing version 11 this morning :) This makes perfect sense though. Thanks!

PR #937, opened 2017-06-19
determine net45 version from registry

jaymedavis:
example output:
`net45.platform: Microsoft Windows NT 10.0.14393.0, OS: Windows, framework: 4.6 RC or later`


anelder-stripe:
:+1:

PR #940, opened 2017-06-21
concept (do not merge): reducing code in the services

jaymedavis:


jaymedavis:
There are two mutually exclusive concepts in this PR for improving the services. Both offer benefits over the existing way:

* The first way (just calling Async from the sync) reduces code duplication and would have no breaking changes for the user.
 
* The second way promotes using async (the code doesn't lock while the request is being made), which I think most would agree is ideal when calling a web api. However, it would require all existing user implementations with sync add ".Result" to the end of all their service calls. The benefit is it removes having to have two implementations of sync/async in every service. Another possibility of moving this direction is measuring sync/async usage via user agent.

jaymedavis:
I added a new class `CustomerService2` here that should probably replace `CustomerService`. This would mean existing users that use sync would need to add `.Result` to the same function calls they have now. Stripe.net just does this for them behind the scenes. This also shows off how to refactor to the StripeBasicService, cutting out lots of cookie cutter code.

src/Stripe.net/Services/Customers/StripeCustomerService.cs
```diff
@@ -17,10 +17,16 @@ public class StripeCustomerService : StripeService
...
-                Requestor.PostString(this.ApplyAllParameters(createOptions, Urls.Customers, false),
-                SetupRequestOptions(requestOptions))
-            );
+            // since the infrastructure uses .Result anyway, this is processsed the same way behind the scenes.
+            // we could simplify the services by removing the duplication of Create/CreateAsync by calling the async method and adding .Result for the sync method
+
+            return CreateAsync(createOptions, requestOptions, CancellationToken.None).Result;
+
+            // alternatively, we could default to all async by renaming CreateAsync below to Create. Then the documentation for sync changes are below
+            // from:
```
jaymedavis: added `CustomerService2` to show what this would look like

fred-stripe:
Is the `GetThing` / `GetThingAsync` a common pattern in C# libraries? It feels like this concept could be pushed down much lower in the stack if it's shared across all of the services.

jaymedavis:
using `Async` at the end of the method signature is a common pattern (for async methods). 

hmm... something like this may be possible to determine from the generic:
``` var service = new GetService<Customer>.Create() ```

you still have the one off's where things have Cancel() or Upcoming() on invoices for example to deal with. 

is that what you mean?

ob-stripe:
Closing this PR. As of v20.0.0, it's a lot less necessary: the request code is now concentrated in `Service.cs` ([here](https://github.com/stripe/stripe-dotnet/blob/28ac21d8ee0d97687f75ec3e216e1824680b2bfa/src/Stripe.net/Services/Service.cs#L76-L125)).

PR #957, opened 2017-07-11
Update readme.md

johnbindel:
Fixed return type of countrySpecService.List() in readme

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -576,7 +576,7 @@ Each account's country can have different rules for required fields, payment met
 
 ```csharp
 	var countrySpecService = new CountrySpecService();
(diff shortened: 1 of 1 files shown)


anelder-stripe:
@johnbindel - The readme is actually soon to be replaced with an entirely new document and the return type of all of the `List()` requests is now `StripeList<T>`.  I am really sorry for the delay.  Please don't let this discourage you from committing!  If you're so inclined, you're welcome to update the PR w/ the correct return-type, but yeah-- it's not going to be around for much longer.

ob-stripe:
Closing the PR as it's no longer applicable.

PR #966, opened 2017-07-19
Remove raw PANs from the Test suite

remi-stripe:
Using raw PANs server-side comes with a higher [PCI compliance](https://stripe.com/docs/security#validating-pci-compliance) burden. Most users of the library should not be using those and instead tokenize client-side.

This PR updates the test suite to remove usage of raw PANs from as many parts as possible:
* Updated the readme to stop referencing raw PAN support
* Updated almost all tests to use pre-built tokens
* Removed some tests making assumptions on Card data as pre-build tokens don't have this information set.
* Added clear comments on the few PANs left as to why whey are still here.


src/Stripe.Tests.XUnit/orders/_fixture.cs
```diff
@@ -27,7 +27,7 @@ public orders_fixture()
                     {
                         Amount = 1,
                         Description = "some random items in a furry sack",
-                        Parent = "sku_stripe_net_test", // i hardcoded this because skus isn't written yet. created from the dashboard
+                        Parent = "sku_B3QG08Tr9vICLA", // i hardcoded this because skus isn't written yet. created from the dashboard
```
anelder-stripe: Please remove this mod, as the `sku_stripe_net_test` actually exists on the test account.

src/Stripe.net.Tests/charges/charge_behaviors.cs
```diff
@@ -40,33 +40,6 @@ public class charge_behaviors
...
-
-        It should_have_the_correct_source_expiration_month = () =>
-            StripeCharge.Source.Card.ExpirationMonth.ShouldEqual(StripeCard.ExpirationMonth);
-
-        It should_have_the_correct_source_expiration_year = () =>
-            StripeCharge.Source.Card.ExpirationYear.ShouldEqual(StripeCard.ExpirationYear);
-
-        It should_have_the_correct_source_name = () =>
-            StripeCharge.Source.Card.Name.ShouldEqual(StripeCard.Name);
-
```
anelder-stripe: Can we modify to check that these values are not null, rather than deleting the test?  I think knowing whether or not they're being populated is important, since this is on such a major object type.

src/Stripe.net.Tests/charges/test_data/stripe_charge_create_options.cs
```diff
@@ -6,26 +6,11 @@ public static class stripe_charge_create_options
...
-            };
-
             return new StripeChargeCreateOptions()
             {
-                SourceCard = cardOptions,
-                Description = "Joe Meatball Charge",
-                StatementDescriptor = "Joe Meatball Sub",
+                SourceTokenOrExistingSourceId = "tok_bypassPending",
+                Description = "my charge description",
+                StatementDescriptor = "my statement",
```
anelder-stripe: 💯

readme.md
```diff
@@ -340,31 +338,6 @@ With a token:
...
-		AddressState = "NC",                  // optional
-		AddressZip = "27617",                 // optional
-		Name = "Joe Meatballs",               // optional
-		Cvc = "1223"                          // optional
-	};
-
-	var cardService = new StripeCardService();
-	StripeCard stripeCard = cardService.Create(*customerId*, myCard); // optional isRecipient
-```
-
```
anelder-stripe: I'm not big on removing this example.  I think just framing it in the context of what you should and shouldn't do.  .NET isn't just used for Web Applications, it's also used for mobile applications.  I could be convinced otherwise, but yeah-- this seems harmless to continue to include provided there is a large warning attached.  (cc: @brandur-stripe - thoughts?)

src/Stripe.net.Tests/charges/charge_behaviors.cs
remi-stripe: We can't though since they are explicitly null with that token as far as I know

src/Stripe.Tests.XUnit/orders/_fixture.cs
remi-stripe: damn you git. will fix.

src/Stripe.net.Tests/charges/charge_behaviors.cs
anelder-stripe: Hrmm.  Okay, if that's true I guess this is peaceful.  It worries me that we can't detect if the deserializer is broken though.

readme.md
remi-stripe: It's still incorrect to say that. PCI compliance and mobile has implications too if you don't use our official tools and the Stripe.net SDK is not part of our PCI reviewed code so we can't pretend it is.

janoserdelyi:
This feels like nanny'ism. Let people know the burden/compliance issues.
We're adults and professionals

remi-stripe:
>This feels like nanny'ism. Let people know the burden/compliance issues.
> We're adults and professionals

@janoserdelyi You might know of PCI compliance but that's not the case for all developers. We have helped a lot Stripe.net users (and Stripe users in general) navigate this issue after they had built their integration through raw PANs because they didn't know it would be a security risk and extra PCI burden to do this. Anyone who wants to use this feature can still do so and the code stays in the library and is not going away.

readme.md
anelder-stripe: I was more using it as an example of GitHub markdown's capability to hide details unless clicked.  Sort-of like spoilers.  Meaning, we can include it but not put it front and center for people to find.  I don't really care what the wording is.

anelder-stripe: It's also worth noting that this is going to be removed in favor of official docs at some point very soon.

readme.md
```diff
@@ -327,9 +327,7 @@ Cards
 
 ### Creating a card
 
-When creating a card you can use either a card or a token
-
-With a token:
+When creating a card you can use either a card or a token. For [PCI compliance](https://stripe.com/docs/security#validating-pci-compliance) though we highly recommend that you pass a token generated client-side.
```
brandur-stripe: Maybe follow this up with "... so that raw credit card numbers are never directly handled by your servers".

remi-stripe:
I just fixed the readme to make it a bit clearer

PR #967, opened 2017-07-19
Remove the TotalCount property of lists

fred-stripe:
Fetching the total count of a list has already been removed from other Stripe libraries and is no longer recommended.


brandur-stripe:
Mandatory comment that this is a breaking change, but other than that, I'm totally on board with seeing more references to this disappear.

anelder-stripe:
I think I inadvertently broke your TotalCount removal by adding tests that depended on TotalCount.  🤕  Sorry!  Will fix.

brandur-stripe:
@fred-stripe Mind rebasing this given that it's pretty stale (want to make sure the tests are still happy because I'm still not confident in my skills to fix them, hah)? I'm going to try and build up some release cadence here, so we'll get this in after you do. Thanks!

ob-stripe:
I rebased the PR so that we can include it in the next major version bump.

PR #970, opened 2017-07-20
Added a small fix that should address the iteration issue

### LINKED ISSUE #968: IEnumerable -> StripeList
I don't think we should be using IEnumerable in 99% of cases where pagination is a possibility, because you lose the `has_more`-attribute.

Affected `List()`-methods should have their return types modified to `StripeList<T>` and their mapper call modified like this:

`Mapper<StripeList<T>>.MapCollectionFromJson()` -> `Mapper<StripeList<T>>.MapFromJson()`

andrewnelder:
Extended the `StripeList` object to implement the `IEnumerable<T>` class.  It then routes the `GetEnumerator()`-method to that of the `Data`-attribute.  This should allow us to then modify the implementation to have it be "correct" without actually breaking any integrations.

cc: @fred-stripe + @remi-stripe - What do you think of this fix?  If you like it, we can just extend it to all of the `List()` methods.

Fixes #968

### DIFF
--- a/src/Stripe.net/Entities/StripeList.cs
+++ b/src/Stripe.net/Entities/StripeList.cs
@@ -1,9 +1,12 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
 using Newtonsoft.Json;
 
 namespace Stripe
 {
-    public class StripeList<T> : StripeEntity
+    [JsonObject]
+    public class StripeList<T> : StripeEntity, IEnumerable<T>
     {
         [JsonProperty("object")]
         public string Object { get; set; }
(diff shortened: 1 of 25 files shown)


src/Stripe.net/Entities/StripeList.cs
```diff
@@ -1,9 +1,12 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
```
remi-stripe: why is it better to do 3 `using` instead of the specific one that loads just `Generic`?

anelder-stripe: Those were auto imports.  They're not all necessary, I think.  :)

fred-stripe:
This approach is great! Does the test suite need to be updated to use this? Seeing a lot of failures from appconveyor.

andrewnelder:
Hrmm.  This is passing locally, but seems to be failing on appveyor with an "invalid key" error, which is neat-- because I've not changed the key OR the appveyor file.  Soooooo...

PR #973, opened 2017-07-25
Added a (small?) fix that should address the iteration issue

### LINKED ISSUE #968: IEnumerable -> StripeList

anelder-stripe:
Extended the `StripeList` object to implement the `IEnumerable<T>` class.  It then routes the `GetEnumerator()`-method to that of the `Data`-attribute.  This should allow us to then modify the implementation to have it be "correct" without actually breaking any integrations.

This is a recommit of a previously closed PR that doesn't screw with AppVeyor.  This recommit also includes a fridge ton of tests!  YAY!

r? @remi-stripe 
r? @fred-stripe 

Fixes #968


brandur-stripe:
Hey @anelder-stripe, sorry about the total fail on my part on the review front, and thank you for continuing to move these forward!

This seems very reasonable to me. Just a couple high level questions so that I can make sure that I actually understand what's going on:

1. Just so I have a rough understanding of the impetus here, do you know why these services in particular used an `IEnumerable<>` instead of `StripeList`? I notice that in some cases like `StripeChargeService`, `StripeList` was already in use, and I just want to make sure that I understand what the perceived distinction was.
2. Does Stripe.NET implement any kind of auto-pagination right now? I just wanted to call it out because if not, there might be a small chance that you want to wrap `StripeList` in a different object that provides this functionality, return that instead, and leave `StripeList` as a very basic JSON wrapper, which is what it seems to be right now.

fred-stripe:
@brandur-stripe re: autopagination, Stripe.net does not currently support that. However, since this changes moves everything to `StripeList` we could potentially add the capability to that object in a later improvement.

Re: IEnumerable vs StripeList is that this might be related to wanting to return the `total_count` on a list sometimes so you'd need a StripeList instead of a generic enumerable? `total_count` is a field I've removed, since we deprecated it in other libraries as well. It doesn't seem like the current used StripeList vs IEnumerable for any other reason.

My only FUD about these changes was related to changing the signature for existing codebases. Thinking through it, by making StripeList implement IEnumerable now that means any existing code that uses IEnumerables will continue to work (it shouldn't require an explicit cast), and anyone expecting a StripeList won't have to change things either.

Since my FUD is addressed, and this gets us closer to having autopagination for all resources, I'm very +1 on this whole change.

I'm also in favor of refactoring and reducing the redundancy of some of these methods and tests further down the road, but this seems like a sensible change in the interim.

brandur-stripe:
> @brandur-stripe re: autopagination, Stripe.net does not currently support that. However, since this changes moves everything to StripeList we could potentially add the capability to that object in a later improvement.

Ah, k. Yep, seems like a nice place to add autopagination at some point. Thanks!

PR #979, opened 2017-08-02
Added capture options to allow for all capturable fields to be utilized

hatdragon:
This update will give users the ability to pass a StripeChargeCaptureOptions object into a capture request.  This will allow for additional API properties to be utilized for items such as destination[amount] and  statement_descriptor.

### DIFF
--- a/src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
+++ b/src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
@@ -27,4 +27,8 @@
     <ProjectReference Include="..\Stripe.net\Stripe.net.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+  </ItemGroup>
+
 </Project>
--- a/src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
+++ b/src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
@@ -27,8 +27,4 @@
     <ProjectReference Include="..\Stripe.net\Stripe.net.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
-  </ItemGroup>
-
 </Project>
--- a/src/Stripe.net.Tests/Stripe.net.Tests.csproj
+++ b/src/Stripe.net.Tests/Stripe.net.Tests.csproj
@@ -49,8 +49,4 @@
     <Reference Include="System" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
-  </ItemGroup>
-
 </Project>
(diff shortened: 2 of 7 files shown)


src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
```diff
@@ -27,4 +27,8 @@
     <ProjectReference Include="..\Stripe.net\Stripe.net.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
anelder-stripe: Can we nix this from the PR?

src/Stripe.net.Tests/Stripe.net.Tests.csproj
```diff
@@ -49,4 +49,8 @@
     <Reference Include="System" />
   </ItemGroup>
 
+  <ItemGroup>
```
anelder-stripe: This too please!  :)

src/Stripe.net.Tests/charges/test_data/stripe_charge_capture_options.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+
+namespace Stripe.Tests.test_data
```
anelder-stripe: Would you be willing to either (a) move these tests to the XUnit project or (b) write some more tests for the XUnit project to include here?  We're going to be removing the mspec tests.  :)

hatdragon: Sure, I'll take a look tomorrow

src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
hatdragon: I can, but VS is just going to add it back in.   Per Microsoft, it's intended behavior.  https://connect.microsoft.com/VisualStudio/feedback/details/800245/vs2013rc-adds-to-vs2012-c-project-section-itemgroup  

>This behavior is intentional. 
>
>To support third-party test frameworks, like NUnit and XUnit, Visual Studio 2012 
>loaded Test Explorer on solution open, regardless of whether it contained test 
>projects. This added seconds of delay to startup and solution open scenarios for all 
>users, majority of whom don't use tests.
>
>In Visual Studio 2013, we changed it so that Test Explorer package is loaded only 
>when the solution contains one or more test projects. Test projects are identified in 
>two different ways. Projects created from one of the built-in unit test project 
>templates are identified using project type GUIDs. Other types of projects, such as 
>Class Library project with XUnit or NUnit tests, are identified by Test Explorer 
>during first test discovery and “tagged” with the <Service/> item. 
>
>You can avoid Visual Studio automatically adding the <Service/> item to your test 
>projects by creating them from any of the built-in unit test project templates. In 
>other words, select the "Unit Test Project" template instead of the "Class Library" 
>template in the New Project dialog.

hatdragon:
Added additional xunit tests and cleaned up a few small things that I missed with the options in my original testing.   Please take a look and let me know if there are any additional things we need to wrap this one up.

hatdragon:
@anelder-stripe?

remi-stripe:
Bumping this one to make sure it gets handled soon

remi-stripe:
@hatdragon I totally missed that you did this a few months ago and I did my own version in https://github.com/stripe/stripe-dotnet/pull/1019. It has been merged though so you should be all set.

Issue #985, opened 2017-08-08
StripeCustomerCreateOptions & StripeCustomerUpdateOptions missing Shipping property


APErebus:
As per #407, the StripeCustomerCreateOptions & StripeCustomerUpdateOptions are missing the optional Shipping property.

This is part of the create/update as per https://stripe.com/docs/api?lang=curl#create_customer-shipping & https://stripe.com/docs/api?lang=curl#update_customer-shipping

PR #986, opened 2017-08-09
Support shipping on Customer creation and update

remi-stripe:
The shipping hash is similar on Customer and Order and has more fields on the Charge resource. For this reason, we now have a common class called StripeShippingOptions and the Charge specific fields are added in a separate class called StripeChargeShippingOptions.

This also adds improved tests to ensure we test the extra fields for the `shipping` on Charge and that we test those fields properly on customer creation and update.

Fixes https://github.com/stripe/stripe-dotnet/issues/985

### DIFF
--- a/src/Stripe.Tests.XUnit/charges/_cache.cs
+++ b/src/Stripe.Tests.XUnit/charges/_cache.cs
@@ -25,13 +25,15 @@ namespace Stripe.Tests.Xunit
                 Amount = 1000,
                 Currency = "usd",
                 SourceTokenOrExistingSourceId = "tok_visa",
-                Shipping = new StripeShippingOptions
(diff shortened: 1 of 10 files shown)


APErebus:
damn that's quick :)

:+1:

remi-stripe:
Bumping this one to make sure it gets handled soon

brandur-stripe:
Sorry about the silence here Remi.

Would you mind rebasing? (Unfortunately, we seem to have produced a bit of a conflict since it was written.) I'll cut a release after that's done and we're green.

remi-stripe:
@brandur-stripe  rebased on master. PTAL!

brandur-stripe:
Released as 11.2.0.

PR #987, opened 2017-08-10
Update readme.md

mglukhovsky:
I've added the same _Documentation_ header that [stripe-node](https://github.com/stripe/stripe-node) uses to point users to our shiny new API docs.


remi-stripe:
LGTM

PR #997, opened 2017-08-30
Added support for ephemeral key creation

remi-stripe:
Ephemeral keys are used in the iOS SDK to edit details on a customer safely in the app without leaking the real Secret API key.

This PR adds support for those. Ephemeral Keys have to be created with a specific API version. The reason is that the raw JSON is sent to the mobile application where our mobile SDK will parse the JSON to be able to use the ephemeral key. Since our SDK will change/evolve over time, the SDK has to explicitly pass an API version so that they get the JSON in the format expected.

Our library only handles one API version though similarly to stripe-go. This needs to be handled so that an API version can be forced for ephemeral keys and nothing else.

For now, I did the quick/dirty version of passing `requestOptions.EphemeralKeyStripeVersion` to see what it would look like.

cc @grey-stripe as you built this for the other SDKs.

### DIFF
--- a/src/Stripe.net/Entities/StripeEphemeralkeyAssociatedObject.cs
+++ b/src/Stripe.net/Entities/StripeEphemeralkeyAssociatedObject.cs
@@ -2,7 +2,7 @@ using Newtonsoft.Json;
 
 namespace Stripe
 {
-    public class StripeEphemeralkeyAssociatedObject : StripeEntityWithId
+    public class StripeEphemeralKeyAssociatedObject : StripeEntityWithId
     {
         [JsonProperty("type")]
         public string Type { get; set; }
--- a/src/Stripe.net/Entities/StripeEphemeralkeyAssociatedObject.cs
+++ b/src/Stripe.net/Entities/StripeEphemeralkeyAssociatedObject.cs
@@ -1,0 +1,10 @@
+using Newtonsoft.Json;
+
+namespace Stripe
+{
+    public class StripeEphemeralkeyAssociatedObject : StripeEntityWithId
+    {
+        [JsonProperty("type")]
+        public string Type { get; set; }
+    }
+}
--- a/src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyCreateOptions.cs
+++ b/src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyCreateOptions.cs
@@ -1,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using Stripe.Infrastructure;
+
+namespace Stripe
+{
+    public class StripeEphemeralKeyCreateOptions
+    {
+        [JsonProperty("customer")]
+        public string CustomerId { get; set; }
+    }
+}
--- a/src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyCreateOptions.cs
+++ b/src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyCreateOptions.cs
@@ -10,6 +10,7 @@ namespace Stripe
         [JsonProperty("customer")]
         public string CustomerId { get; set; }
 
+        [JsonIgnore]
         public string StripeVersion { get; set; }
     }
 }
--- a/src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyCreateOptions.cs
+++ b/src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyCreateOptions.cs
@@ -9,5 +9,7 @@ namespace Stripe
     {
         [JsonProperty("customer")]
         public string CustomerId { get; set; }
+
+        public string StripeVersion { get; set; }
     }
 }
--- a/src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyService.cs
+++ b/src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyService.cs
@@ -1,0 +1,52 @@
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Stripe.Infrastructure;
+
+namespace Stripe
+{
+    public class StripeEphemeralKeyService : StripeService
+    {
+        public StripeEphemeralKeyService(string apiKey = null) : base(apiKey) { }
+
+
+
+        //Sync
+        public virtual StripeEphemeralKey Create(StripeEphemeralKeyCreateOptions createOptions, StripeRequestOptions requestOptions = null)
+        {
+            return Mapper<StripeEphemeralKey>.MapFromJson(
+                Requestor.PostString(this.ApplyAllParameters(createOptions, Urls.EphemeralKeys, false),
+                SetupRequestOptions(requestOptions))
+            );
+        }
+
+        public virtual StripeDeleted Delete(string EphemeralKeyId, StripeRequestOptions requestOptions = null)
+        {
+            return Mapper<StripeDeleted>.MapFromJson(
+                Requestor.Delete(this.ApplyAllParameters(null, $"{Urls.EphemeralKeys}/{EphemeralKeyId}", false),
+                SetupRequestOptions(requestOptions))
+            );
+        }
+
+
+
+        //Async
+        public virtual async Task<StripeEphemeralKey> CreateAsync(StripeEphemeralKeyCreateOptions createOptions, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return Mapper<StripeEphemeralKey>.MapFromJson(
+                await Requestor.PostStringAsync(this.ApplyAllParameters(createOptions, Urls.EphemeralKeys, false),
+                SetupRequestOptions(requestOptions),
+                cancellationToken)
+            );
+        }
+
+        public virtual async Task<StripeDeleted> DeleteAsync(string EphemeralKeyId, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return Mapper<StripeDeleted>.MapFromJson(
+                await Requestor.DeleteAsync(this.ApplyAllParameters(null, $"{Urls.EphemeralKeys}/{EphemeralKeyId}", false),
+                SetupRequestOptions(requestOptions),
+                cancellationToken)
+            );
+        }
+    }
+}
--- a/src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyService.cs
+++ b/src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyService.cs
@@ -5,48 +5,32 @@ using Stripe.Infrastructure;
 
 namespace Stripe
 {
-    public class StripeEphemeralKeyService : StripeService
+    public class StripeEphemeralKeyService : StripeBasicService<StripeEphemeralKey>
     {
         public StripeEphemeralKeyService(string apiKey = null) : base(apiKey) { }
 
-
-
-        //Sync
+        // Sync
         public virtual StripeEphemeralKey Create(StripeEphemeralKeyCreateOptions createOptions, StripeRequestOptions requestOptions = null)
         {
-            return Mapper<StripeEphemeralKey>.MapFromJson(
-                Requestor.PostString(this.ApplyAllParameters(createOptions, Urls.EphemeralKeys, false),
-                SetupRequestOptions(requestOptions))
-            );
+            return Post(Urls.EphemeralKeys, requestOptions, createOptions);
         }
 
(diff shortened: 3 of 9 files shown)


remi-stripe:
@brandur-stripe @grey-stripe what do you think of my approach in here? Every resource inherits from `StripeEntity` which has a `StripeResponse` property that itself has `ResponseJson`. Is this enough for our needs instead of implementing a custom deserializer to store the JSON at the top level?

src/Stripe.net/Services/StripeRequestOptions.cs
```diff
@@ -5,5 +5,8 @@ public class StripeRequestOptions
         public string ApiKey { get; set; }
         public string StripeConnectAccountId { get; set; }
         public string IdempotencyKey { get; set; }
+        // This is used specifically for Ephemeral Keys as they have to be created
+        // with a specific API version. Don't use it for anything else.
+        public string EphemeralKeyStripeVersion { get; set; }
```
brandur-stripe: I don't think this ephemeral key specific field should be in this broadly used class. Can't you have each service method add in a normal Stripe version when they're called into?

brandur-stripe: Or I guess there is no other `StripeVersion` option. IMO it wouldn't be too bad to just add a general purpose `StripeVersion` here to allow per-request overrides even if the library pegs itself to a single version normally.

brandur-stripe:
> @brandur-stripe @grey-stripe what do you think of my approach in here? Every resource inherits from StripeEntity which has a StripeResponse property that itself has ResponseJson. Is this enough for our needs instead of implementing a custom deserializer to store the JSON at the top level?

Yeah, that seems like it'd be fine.

src/Stripe.net/Entities/StripeEphemeralkeyAssociatedObject.cs
```diff
@@ -0,0 +1,10 @@
+using Newtonsoft.Json;
+
+namespace Stripe
+{
+    public class StripeEphemeralkeyAssociatedObject : StripeEntityWithId
```
brandur-stripe: I think the casing here should be the same as the other entity class (i.e. "EphemeralKey" instead of "Ephemeralkey").

brandur-stripe:
I still don't really know the library conventions very well here so I'm a little hesitant to commit to a full-on review, but this looks good to me. Thanks for taking a crack at it!

One thing that I find a little concerning compared to something like Go is that because the raw JSON is hidden away by a few layers of indirection, it may not be very transparent to users that they should be passing that back to clients instead of re-serializing the entity they got back (given no other information, I would probably do the later). I'm not exactly sure how to fix this though ... maybe even just additional documentation would be enough? Another possibility is a helper method for `RawJSON` that just reaches into the response for you and which is itself documented.

remi-stripe:
> given no other information, I would probably do the later

brandur-stripe:
Oops, that was supposed to read "latter" no "later".

> Agreed but it's on us to document this well. I'm happy to build something I'm just not really sure how to approach this cleanly in the library.

Yeah I'm not sure either. I just mention it because I've seen many an error that could have been prevented by reading documentation, but time is short and most of us are in the business of skimming documentation instead.

I guess the easier thing to do for now would be just to try and document it well and see how it goes.

src/Stripe.net/Entities/StripeEphemeralkeyAssociatedObject.cs
remi-stripe: ah nice catch done.

src/Stripe.net/Services/StripeRequestOptions.cs
remi-stripe: Yeah, I'm not a fan of my solution either, I tried to have the naming be specific enough. Not sure how to offer an extra option because that's the way IdempotencyKey and other features were built. Do you have an idea?

brandur-stripe: > Do you have an idea?

IMO just renaming it to `StripeVersion` and enabling per-request overrides if a user really wants to do them is the better option.

remi-stripe: @brandur-stripe I disagree with this approach. Doing this would "pretend" you can set a specific version when really you can't and almost all objects would break right? @grey-stripe handled this in a custom way in Go by catching errors on JSON deserialization specifically for the ephemeralkeys package but I'm not sure how to approach this in stripe-dotnet and whether it's the right approach (I don't think it is).
I guess the other solution is to take the Go approach and have a custom parameter on the options class that is not associated to a JSON key but that seems dirty too.

remi-stripe:
> I guess the easier thing to do for now would be just to try and document it well and see how it goes.

the other/cleaner solution would be to figure out how to handle a custom deserializer. I can try to work on this next week and see.

brandur-stripe:
> the other/cleaner solution would be to figure out how to handle a custom deserializer. I can try to work on this next week and see.

I don't think we really need that right? It's only in Go because there'd otherwise be no way to access the response's raw JSON data.

remi-stripe:
> I don't think we really need that right? It's only in Go because there'd otherwise be no way to access the response's raw JSON data.

We need this to be able to set a custom property on the object that is initialized to the raw JSON instead of having to dig deeper. Sorry I think we're mixing issues because this is not threaded in github properly 😭 

You seemed worried that someone would not know that they have to use `ephemeralKey.StripeResponse.ResponseJson` and would do the wrong thing. If we want to have something like `keyJson` on the object we need a custom deserializer to set this up right? I don't know enough of the internals of the library to be sure but that was my assumption a bit similar to what the Source deserializer does.

brandur-stripe:
> You seemed worried that someone would not know that they have to use ephemeralKey.StripeResponse.ResponseJson and would do the wrong thing. If we want to have something like keyJson on the object we need a custom deserializer to set this up right? I don't know enough of the internals of the library to be sure but that was my assumption a bit similar to what the Source deserializer does.

Oh I see! Sorry I wasn't very clear, but I was thinking something simple like this:

``` c#
// RawJSON is the raw JSON data of the response that was used to initialize this 
// ephemeral key. When working with mobile clients that might only understand 
// one version of the API you should prefer to send this value back to them so 
// that they'll be able to decode an object that's current according to their version.
public RawJSON {
    get { return this.StripeResponse.ResponseJson; }
}
```

Basically just a pass through to a property that we already have.

remi-stripe:
> Basically just a pass through to a property that we already have.

Damn did not think of adding a getter this way to the class. I'll try this out tomorrow and let you know if it works.

jaymedavis:
If you want to be a little defensive: 

this.StripeResponse?.ResponseJson

brandur-stripe:
> If you want to be a little defensive:

Nice! One of the fancy new features introduced since the last time I did any serious C#. +1.

remi-stripe:
Okay fixed the getter with @brandur-stripe and @jaymedavis advice.

src/Stripe.net/Services/StripeRequestOptions.cs
brandur-stripe: > I disagree with this approach. Doing this would "pretend" you can set a specific version when really you can't and almost all objects would break right?

In reality almost everything would still work, even if you passed an old API version. Not all that much changes version to version, and for the majority of the resources in the API, nothing change at all.

Now that I think about it a little more though, why don't we just leverage visibility here? I think we should still call it `StripeVersion` because having a field specific to ephemeral keys here is pretty ugly, but just mark it `internal` so that no one but us can use it.

remi-stripe: Gotcha. I agree that in most cases it would just work, which makes it more dangerous because it's quite hard to catch in tests. But it's now flagged as internal with a clear comment so hopefully no one plays with it incorrectly

src/Stripe.net/Services/StripeRequestOptions.cs
```diff
@@ -5,5 +5,8 @@ public class StripeRequestOptions
         public string ApiKey { get; set; }
         public string StripeConnectAccountId { get; set; }
         public string IdempotencyKey { get; set; }
+        // This is used specifically for Ephemeral Keys as they have to be created
+        // with a specific API version. Don't use it for anything else.
+        internal string StripeVersion { get; set; }
```
brandur-stripe: Thanks Remi. That looks good to me!

Belal-Shaat: @remi-stripe @brandur-stripe  So when this pull request will be available  , I can't stand to use it , my ios mobile developer need this feature to complete the job .

jakubfijalkowski: Shouldn't this be public? Right now, user is unable to specify the `StripeVersion` when requesting the ephemeral key as this property is `internal` to the library (and tests - `InternalsVisibleTo`).

remi-stripe: @jakubfijalkowski ha nice catch, I assumed the tests would have failed similarly to how a code using the library would have failed. I'll have to figure out a better solution than this to set the property then.
@brandur-stripe do you have an opinion here?

remi-stripe: Okay I think I found the right fix and pushed it in https://github.com/stripe/stripe-dotnet/pull/997/commits/8490d0f4e5b8a192c102a119680830f80c178474

src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyService.cs
```diff
@@ -12,6 +12,13 @@ public class StripeEphemeralKeyService : StripeBasicService<StripeEphemeralKey>
         // Sync
         public virtual StripeEphemeralKey Create(StripeEphemeralKeyCreateOptions createOptions, StripeRequestOptions requestOptions = null)
         {
+            // Creating an ephemeral key requires a specific API version to be set. This is handled as a parameter
+            // but has to be set on the StripeRequestOptions instead.
+            if (requestOptions == null)
```
jakubfijalkowski: `requestOptions = requestOptions ?? new StripeRequestOptions();`

src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyCreateOptions.cs
```diff
@@ -9,5 +9,7 @@ public class StripeEphemeralKeyCreateOptions
     {
         [JsonProperty("customer")]
         public string CustomerId { get; set; }
+
+        public string StripeVersion { get; set; }
```
jakubfijalkowski: `[JsonIgnore]`? I don't know how the serializer is configured, but otherwise it might include the field in the serialized JSON.

src/Stripe.net/Services/StripeRequestOptions.cs
jakubfijalkowski: @remi-stripe Great! I've left some other small comments, feel free to ignore them. ;)

src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyService.cs
remi-stripe: oh neat I had never heard of this one but I can see it's used elsewhere in the library so will fix.

src/Stripe.net/Services/EphemeralKeys/StripeEphemeralKeyCreateOptions.cs
remi-stripe: yeah it works without it but it does not hurt to be clearer

remi-stripe:
Rebased on latest master and now test passes

brandur-stripe:
Nice work Remi! This LGTM.

One minor note: I notice that in stripe-go and stripe-ruby we error out if you try to  create an ephemeral key without specifying a version (i.e. you don't get one handed to you implicitly). I believe this is meant as a forcing function so that people will use the model in the right way. Do you think we should do the same here?

remi-stripe:
@brandur-stripe I think it could be useful but I don't see an example pattern in the library today where we explicitly try to validate something upfront. Would you have an idea of the best way to approach this?

I could technically throw a `StripeException` but this is clearly tailored towards an API error.

brandur-stripe:
> I think it could be useful but I don't see an example pattern in the library today where we explicitly try to validate something upfront. Would you have an idea of the best way to approach this?

@remi-stripe Yeah, I think `StripeException` is the wrong way to go. In cases like this where a method notices that it's been passed a bad argument, it's pretty normal convention to throw an [`ArgumentException`](https://msdn.microsoft.com/en-us/library/system.argumentexception(v=vs.110).aspx). Even though the library doesn't do it elsewhere, I still think that it'd be fine to add it here. What do you think?

brandur-stripe:
LGTM.

brandur-stripe:
Released as 11.2.0.

Issue #1003, opened 2017-09-05
Accessing `RequestId` from `


ob-stripe:
At the moment, it doesn't appear to be possible to access the request ID from a `StripeException` instance.

The request ID is accessible on [`StripeResponse`](https://github.com/stripe/stripe-dotnet/blob/96cdbaffa5609ab6f8c4925342247c5a8be6f0ac/src/Stripe.net/Infrastructure/Public/StripeResponse.cs#L9) objects, but unless I'm mistaken this class is not referenced directly or indirectly from `StripeException`.

Other language libraries do have this feature, e.g. the Java `StripeException` class does store the request ID and lets users retrieve it via the [`getRequestId()`](https://github.com/stripe/stripe-java/blob/da0f8208e7505b635508d09e7bd38686ed9f9790/src/main/java/com/stripe/exception/StripeException.java#L22-L24) accessor.

PR #1004, opened 2017-09-06
Allow access to the StripeResponse on the StripeException resource

remi-stripe:
Adding this allows developers to see details about the request when getting and exception such as the Request Id (req_XXXXX).

This fixes https://github.com/stripe/stripe-dotnet/issues/1003

### DIFF
--- a/src/Stripe.net.Tests/charges/when_creating_a_charge_with_an_invalid_card.cs
+++ b/src/Stripe.net.Tests/charges/when_creating_a_charge_with_an_invalid_card.cs
@@ -21,6 +21,7 @@ namespace Stripe.Tests
             var exception = (StripeException) Catch.Exception(() => _stripeChargeService.Create(StripeChargeCreateOptions));
             exception.Message.ShouldNotBeNull();
             exception.StripeError.ChargeId.ShouldNotBeNull();
+            exception.StripeResponse.RequestId.ShouldNotBeNull();
         };
     }
 }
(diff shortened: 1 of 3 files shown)


brandur-stripe:
Ah man, I need to do some serious spin up on .NET, but this looks mostly fine to me.

One thing that you might want to consider is to initialize the new exception property with an object initializer instead of adding a new constructor parameter:

``` c#
        StudentName student2 = new StudentName
        {
            FirstName = "Craig",
            LastName = "Playstead",
        };
```

The nice part about that is that you don't change that constructor API. I'm not sure I'd technically included it as part of our public interface, but it may be getting called in people's test suites.

remi-stripe:
Okay fixed the getter with @brandur-stripe and @jaymedavis advice.

remi-stripe:
@brandur-stripe can you still do this and pass the default parameter `message` from the parent exception? I'm not fully grasping the benefits here.

remi-stripe:
okay so this works: 
```
            return new StripeException(stripeError.Message) {
                HttpStatusCode = statusCode,
                StripeError = stripeError,
                StripeResponse = response
            };
```

What's the benefit versus the previous version? To avoid changing the constructor every time we add a new property to the class?

brandur-stripe:
@remi-stripe I was thinking this:

``` c#
return new StripeException(statusCode, stripeError, stripeError.Message) {
    StripeResponse = response,
};
```

(If C# allows it anyway.)

The advantage would be that you get to keep the constructor's current signature so that people already using it could still use it.

I don't think it matters that much though given that although technically public, things like `StripeException`'s constructor should be mostly part of our "Stripe-only"  API.

remi-stripe:
Ohhhhh gotcha, totally misunderstood what you were trying to solve. Makes sense, will add that now

remi-stripe:
@brandur-stripe PTAL

brandur-stripe:
LGTM.

brandur-stripe:
Thanks again Remi!

Issue #1002, opened 2017-09-05
Support updating expiration dates on Card Sources.


fred-stripe:
This is discussed in the Stripe docs here:

https://stripe.com/docs/sources/cards#updating-a-card-source-expiration-date

Unfortunately, when updating a Source, you can only update the owner & metadata properties:

https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/Sources/StripeSourceUpdateOptions.cs

So we either need to subclass this for Card sources, or just make an optional field on the StripeSourceUpdateOptions.

ob-stripe:
Fixed in #1005 and released in v11.1.0.

PR #1005, opened 2017-09-06
Added support for updating a Card Source expiration date

remi-stripe:
You can now update a Card Source's expiration date via a new type. Also improved the Card Source tests.

### DIFF
--- a/src/Stripe.Tests.XUnit/sources/creating_and_updating_card_source.cs
+++ b/src/Stripe.Tests.XUnit/sources/creating_and_updating_card_source.cs
@@ -5,27 +5,52 @@ namespace Stripe.Tests.Xunit
 {
     public class creating_and_updating_card_source
     {
-        private StripeSource Source { get; set; }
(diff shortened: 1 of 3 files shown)


remi-stripe:
This fixes https://github.com/stripe/stripe-dotnet/issues/1002

brandur-stripe:
Thanks for the fix! It looks like the test suite failures are related to the change, so maybe take a look at that.

I'm a little out of my league here, but based on what I see by looking at a few other options classes, I think this is correct. LGTM.

remi-stripe:
@brandur-stripe damn, forgot to refresh `git gui` before the commit. Sorry about that. Pushed a fix to the test.

I think the other failures are unrelated. My understanding from @anelder-stripe is that the test suite does not run properly on a brand or "something something". Unsure how to make sure the code works fully

brandur-stripe:
Cool. I still don't think I have the whole story on CI, but code LGTM.

remi-stripe:
I ran the tests locally so hopefully it's enough. I have no idea what the next steps (merge + deploy) are though.

Issue #779, opened 2017-03-14
Registering domain with Stripe and ApplePay for Stripe Connect


Kristea:
I'm trying to set up apple pay using Stripe Connect. Does this package have any functionality for authenticating a request and registering the domain for apple pay and the connect stripe account id? In a similar way to how the Stripe api describe it.

 Or would this just best be done using the usual .net Web request and passing the necessary keys in the headers?

Thanks.

jaymedavis:
I'm not very familiar with this process. 

> Or would this just best be done using the usual .net Web request and passing the necessary keys in the headers?

Can you point me to where you read this information? If it's a web request and keys in the headers, it may be something that has been implemented, or could be implemented.

KoanLeeroy:
https://stripe.com/docs/apple-pay/web#connect

Just a simple curl request required: 

```
curl https://api.stripe.com/v1/apple_pay/domains \
 -u {PLATFORM_SECRET_KEY}: \
 -H "Stripe-Account: {CONNECTED_STRIPE_ACCOUNT_ID}" \
 -d domain_name="example.com"
```

Would be great to see this in the SDK

PR #1007, opened 2017-09-07
Added support for the ApplePayDomain resource

remi-stripe:
This fixes https://github.com/stripe/stripe-dotnet/issues/779


remi-stripe:
@brandur-stripe do you want to try and cut a release with that one? It's a small one and adds a new resource so it should be less risky than the other ones. If it works we can then do a bigger release with all the pending changes

brandur-stripe:
LGTM.

> do you want to try and cut a release with that one? It's a small one and adds a new resource so it should be less risky than the other ones. If it works we can then do a bigger release with all the pending changes

Great idea! I'll try and see how it goes.

PR #1008, opened 2017-09-07
Always use Bearer auth

ob-stripe:
cc @stripe/api-libraries @remi-stripe 

In the wake of https://github.com/stripe/stripe-node/pull/379 and https://github.com/stripe/stripe-go/pull/447, I figured I'd also add Bearer auth support to the .NET library.

It turns out the library already supports Bearer auth, it just doesn't use it by default. I have checked our internal logs and over the past week, 100% of requests sent by the library used Basic auth, so I think it's safe to say that ~no one is using this feature at the moment.

~This PR simply switches the default value of the `useBearer` flag to `true` to enable Bearer auth by default.~

This PR removes support for Basic auth entirely and leaves Bearer auth as the only auth method.

### DIFF
--- a/src/Stripe.net/Infrastructure/Requestor.cs
+++ b/src/Stripe.net/Infrastructure/Requestor.cs
@@ -50,13 +50,6 @@ namespace Stripe.Infrastructure
             return ExecuteRequest(wr);
         }
 
-        public static StripeResponse PostStringBearer(string url, StripeRequestOptions requestOptions)
-        {
-            var wr = GetRequestMessage(url, HttpMethod.Post, requestOptions, true);
-
-            return ExecuteRequest(wr);
-        }
-
         public static StripeResponse PostFile(string url, string fileName, Stream fileStream, string purpose, StripeRequestOptions requestOptions)
         {
             var wr = GetRequestMessage(url, HttpMethod.Post, requestOptions);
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Nice!

Is your plan to turn bearer auth on for now and then remove support for the old path in the future?

ob-stripe:
Good question. I figured since the ability to choose between different auth methods already existed, we might as well leave it in, though I cannot think of any case where Basic auth is preferable over Bearer.

It's a private internal method so it wouldn't be a breaking change, plus log analysis suggests that nobody knows about or uses this feature anyway. Maybe we ought to kill the Basic auth path right away?

brandur-stripe:
> It's a private internal method so it wouldn't be a breaking change, plus log analysis suggests that nobody knows about or uses this feature anyway. Maybe we ought to kill the Basic auth path right away?

I don't have too strong of a position on this, but as long as we're reasonably sure that the bearer auth path works, I'd be lightly in favor of killing it down to a single auth mechanic. This is basically how all the other libraries work now (thanks to you :), so we may as well keep stripe-dotnet doing the same thing.

ob-stripe:
@brandur-stripe I updated the PR to kill the Basic auth code path entirely.

It turns out the library did use Bearer auth internally for OAuth requests, which require Bearer auth. (Also, TIL the .NET library supports OAuth requests!) I didn't see any such requests in our logs, so maybe users are unaware of this (undocumented, AFAIK) feature, or maybe I just don't know how to look at logs.

In any case, it should be safe to kill off Basic auth.

brandur-stripe:
So much cleaner! LGTM.

(BTW, I'm not yet merging PRs like I do in other languages because as of yet I still haven't really cut a release yet, and am not really set up for it right now. Definitely feel free to go for it if you are!)

PR #1014, opened 2017-09-26
Add Id to StripeCustomerCreateOptions.

johnnyoshika:
Stripe allows us to specify a customer ID when creating customers. This option was missing from `StripeCustomerCreateOptions` prior to this PR.


remi-stripe:
@johnnyoshika  this option has been deprecated and undocumented for well over 3 years and we actively discourage using it. This is not something you should be using. Stripe creates an id automatically for you `cus_XXXX` and you should map this to your own id in your own database.

This is not a feature we would bring in the library unfortunately.

johnnyoshika:
@remi-stripe I didn't know about the deprecation. That's unfortunate, as defining our own ids didn't require us to map Stripe's on our end.  Oh well!

Issue #1017, opened 2017-09-29
Change destination amount on pre auth charge


crono009:
I'm currently implementing the two step payment flow and I'm wondering if there is a way to change the destination amount of the uncaptured charge. 

StripeChargeService.Capture() has a parameter for changing the captureAmount and applicationfee, but I can't seem to find a way to change the destination amount even though the stripe documents say you should be able to.  I was wondering if Stripe.net has this feature.

Also, does anyone know how to release a pre-auth charge if we no longer need it? I know if we don't capture it within 7 days it will release the funds, but I would like to do it earlier out of courtesy for the customer.

Thanks in advanced!

remi-stripe:
@crono009 The `Capture` [method](https://github.com/stripe/stripe-dotnet/blob/f005b1b52ca42df408eaea4ccdd41b84820c3e24/src/Stripe.net/Services/Charges/StripeChargeService.cs#L57) does not support this just yet.

I think we would need to move `Capture` to using something like `StripeChargeCaptureOptions` instead of direct parameters in the method.

If you're up for submitting a PR, that would be awesome! If not, we'll do our best to get to it in the coming weeks.

PR #1019, opened 2017-10-02
Support the ability to pass custom parameters when capturing a charge

remi-stripe:
This introduces a new resource, StripeChargeCaptureOptions. This also changes the signature of the Capture method. It moves away from a list of parameters and groups everything under a common resource as we do for other APIs such as the Pay Order one.

r? @brandur-stripe 

Fixes https://github.com/stripe/stripe-dotnet/issues/1017


brandur-stripe:
LGTM.

I guess let's just be cognizant of the fact that this is a breaking change.

remi-stripe:
Rebased on master, all tests pass

PR #1020, opened 2017-10-02
Fix most of the test failures

remi-stripe:
This PR tries to fix most of the tests failures in the library. All XUnit tests were fixed (at least with my API keys). Only fixed one of the other tests. Most tests complain about file uploads (I think) but I could not figure out what the issue is yet.

r? @brandur-stripe

### DIFF
--- a/src/Stripe.Tests.XUnit/accounts/_fixture.cs
+++ b/src/Stripe.Tests.XUnit/accounts/_fixture.cs
@@ -44,7 +44,8 @@ namespace Stripe.Tests.Xunit
                         new StripeAccountAdditionalOwner
                         {
                             FirstName = "Bumble", LastName = "B",
-                            BirthDay = 29, BirthMonth = 8, BirthYear = 2013,
+                            // Ensure the owner is older than 18 to avoid API issues.
+                            BirthDay = 29, BirthMonth = 8, BirthYear = 1980,
                             VerificationDocument = file.Id
                         },
                         new StripeAccountAdditionalOwner
--- a/src/Stripe.Tests.XUnit/plans/_cache.cs
+++ b/src/Stripe.Tests.XUnit/plans/_cache.cs
@@ -4,14 +4,21 @@ namespace Stripe.Tests.Xunit
 {
     public static partial class Cache
     {
+        private static Object cacheLock = new Object();
+
         public static StripePlan GetPlan(string planName = "plan")
         {
-            if (Items.ContainsKey(planName)) return (StripePlan) Items[planName];
+            // Ensures that creating a new plan only happens synchronously
+            // Avoids parallel calls to end up with the same options (and id) twice.
+            lock(cacheLock)
+            {
+                if (Items.ContainsKey(planName)) return (StripePlan) Items[planName];
 
-            var plan = new StripePlanService(ApiKey).Create(GetPlanCreateOptions(planName));
-            Items.Add(planName, plan);
+                var plan = new StripePlanService(ApiKey).Create(GetPlanCreateOptions(planName));
+                Items.Add(planName, plan);
 
-            return plan;
+                return plan;
+            }
         }
 
         public static StripePlanCreateOptions GetPlanCreateOptions(string planName = "plan")
--- a/src/Stripe.net.Tests/plans/when_deleting_a_plan.cs
+++ b/src/Stripe.net.Tests/plans/when_deleting_a_plan.cs
@@ -23,7 +23,7 @@ namespace Stripe.Tests
         It should_throw_exception_when_getting = () =>
         {
             var exception = Catch.Exception(() => _stripePlanService.Get(_createdStripePlanId));
-            exception.Message.ShouldContain("No such plan");
+            exception.Message.ShouldContain("No such ");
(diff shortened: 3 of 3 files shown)


src/Stripe.Tests.XUnit/accounts/_fixture.cs
```diff
@@ -44,7 +44,7 @@ public account_fixture()
                         new StripeAccountAdditionalOwner
                         {
                             FirstName = "Bumble", LastName = "B",
-                            BirthDay = 29, BirthMonth = 8, BirthYear = 2013,
+                            BirthDay = 29, BirthMonth = 8, BirthYear = 1980,
```
brandur-stripe: I guess this was broken by the age check that Security recently put in? Might be worth adding a comment to the effect that the `BirthYear` is somewhat important.

src/Stripe.Tests.XUnit/plans/_cache.cs
```diff
@@ -23,7 +23,7 @@ public static StripePlanCreateOptions GetPlanCreateOptions(string planName = "pl
                 Amount = 1000,
                 Currency = "usd",
                 Name = Guid.NewGuid().ToString(),
-                Id = Guid.NewGuid().ToString(),
+                Id = Guid.NewGuid().ToString() + DateTime.UtcNow.Ticks,
```
brandur-stripe: Any idea why this works? A GUID should already be universally unique (unless some crazy stubbing has been applied).

brandur-stripe:
A couple minor comments, but looking good. Thanks for taking the initiative here Remi!

src/Stripe.Tests.XUnit/plans/_cache.cs
remi-stripe: @brandur-stripe well I think I'm just unlucky enough? I keep getting "plan already exists" over and over when running the test suite and those conflict on the random ids. But I have ran the test suites hundreds of times so maybe it's due to that?

src/Stripe.Tests.XUnit/accounts/_fixture.cs
remi-stripe: I'm happy to put a comment though it feels self explanatory that the DOB should be a valid one and not for a 4 year old. Let me know if it's worth the fix though

src/Stripe.Tests.XUnit/plans/_cache.cs
brandur-stripe: Just for general interest, I looked it up and .NET uses a V4 UUID generator (random).

Here's a [good writeup from a Quora answer](https://www.quora.com/Has-there-ever-been-a-UUID-collision):

> UUIDs have 122 bits of entropy so the chance of two random UUIDs colliding is about 10^-37.
>
> If you generate 2^46 UUIDs (approximately 1 petabyte of entropy) the chance of getting a collision is 1 in 50 billion.[1]  You're 170 times more likely to win the jackpot in Powerball.[2]

Meaning to say that the chances of generating a duplicate GUID (modulo a bug in the underlying software) are ~zero. I guess it's not a big deal, but I think that the required addition of extra entropy to this string probably means that we have some other bug or feature that we don't completely understand somewhere.

src/Stripe.Tests.XUnit/accounts/_fixture.cs
brandur-stripe: Alright, fair enough.

src/Stripe.Tests.XUnit/plans/_cache.cs
remi-stripe: I think you're right though I'm not grasping why I experience this. My ids are definitely GUIDs and I can reproduce reliably. Unless I add some logs that is, when I do the error does not happen anymore.
I guess this is the wrong fix and we should understand what's causing this issue in the first place.

remi-stripe: Okay, I think the issue here is that we go into `GetPlan()` and the plan does not exist yet. But when we go into `GetPlanCreateOptions()` another concurrent call to `GetPlanCreateOptions()` already initialized the options in the `Items` array. This causes another part of the code to try and create a plan a second time with the same option.

I do not understand why adding a timestamp fix those. It's likely a side-effect and not a solution per se.

I guess the right solution is to take a lock in `GetPlan()` and release it only when leaving.

remi-stripe: removed this attempt for now as it's the wrong approach.

remi-stripe:
@brandur-stripe PTAL. I changed the fix for the subscriptions' tests failures.

brandur-stripe:
Way better!

LGTM.

remi-stripe:
@brandur-stripe @fred-stripe I wonder if we should merge this one? It's not a change to the library itself so it can be merged without needing a release right?
@gh-stripe is working on another modification and encountered all those tests errors too

brandur-stripe:
Yeah, I think we should. And BTW, after seeing Andrew's instructions, I'm fairly confident that we could ship a release without too much trouble, so we should try that at some point.

Issue #916, opened 2017-06-06
Stripe Sku Service


tjmcnaboe:
Could use the Sku service to go with Orders.

This is what I am using in my project to fill the void.  :

    public class StripeSkuService : StripeBasicService<StripeSku>
    {

        public string UrlsSkus = "https://api.stripe.com/v1/skus";

        public StripeSkuService(string apiKey = null) : base(apiKey) { }

        public virtual StripeSku Get(string skuId, StripeRequestOptions requestOptions = null)
        {
            return GetEntity($"{UrlsSkus}/{skuId}", requestOptions);
        }

        public virtual StripeSku Create(StripeSkuCreateOptions createOptions = null, StripeRequestOptions requestOptions = null)
        {
            return Post(UrlsSkus, requestOptions, createOptions);
        }

        public virtual StripeSku Update(string skuId, StripeSkuUpdateOptions updateOptions, StripeRequestOptions requestOptions = null)
        {
            return Post($"{UrlsSkus}/{skuId}", requestOptions, updateOptions);
        }

        public virtual IEnumerable<StripeSku> List(StripeSkuListOptions listOptions = null, StripeRequestOptions requestOptions = null)
        {
            return GetEntityList(UrlsSkus, requestOptions, listOptions);
        }

        public virtual StripeDeleted Delete(string skuId, StripeRequestOptions requestOptions = null)
        {
            return DeleteEntity($"{UrlsSkus}/{skuId}", requestOptions);
        }
    }

tangentmind:
thank you @tjmcnaboe ! @jaymedavis are there any plans for implementing the SKU service?

remi-stripe:
@tangentmind I've started a PR for this in https://github.com/stripe/stripe-dotnet/pull/1022. Hopefully we can merge this in the next few weeks.

PR #1022, opened 2017-10-09
Add support for SKU APIs

remi-stripe:
This PR adds support for SKU APIs and fixes https://github.com/stripe/stripe-dotnet/issues/916 along with an encoding issue for the Product resource.

* Before this fix, the `package_dimensions` parameters were not encoded properly. Instead, we sent a string set to `Stripe.StripePackageDimensionOptions` instead of being a sub-hash with the expected value. The first commit fixes the issue by creating a dedicated class that inherits from `INestedOptions` and sends each parameter as the right sub-hash by forcing the name in `JsonProperty`.
* The product tests now also clean up after themselves at the end. This avoids filling up accounts with data we never use.
* Add support for creating and updating SKU with a shared resource
* Add support for inventory update via `INestedOptions`. This is required for the class to be encoded properly when sent to the API
* Add custom handler for `attributes` in the `DictionaryPlugin.cs`. This was used for `fraud_details` and `metadata` before. Since `attributes` is a dictionary on SKU, we need a custom parser to encode each property.
* Add test suite to ensure complex parameters such as `package_dimensions` are passed to the API and well tested.
* Removes hardcoded sku id in the test suite and dynamically create one now. This avoids manual changes to the library when testing locally. The corresponding product/sku can't be deleted as they are used in the order.

r? @brandur-stripe

### DIFF
--- a/src/Stripe.Tests.XUnit/orders/_fixture.cs
+++ b/src/Stripe.Tests.XUnit/orders/_fixture.cs
@@ -18,6 +18,22 @@ namespace Stripe.Tests.Xunit
 
         public orders_fixture()
         {
+            var productService = new StripeProductService(Cache.ApiKey);
+            var product = productService.Create(new StripeProductCreateOptions {
+                Name = "T-shirt",
+            });
+
+            var skuService = new StripeSkuService(Cache.ApiKey);
+            var sku = skuService.Create(new StripeSkuCreateOptions
+            {
+                Currency = "usd",
+                Inventory = new StripeInventoryOptions {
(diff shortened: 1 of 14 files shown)


remi-stripe:
I considered doing the Product modification in a separate PR but since SKU has the same parameter it blocked adding tests and it felt easier this way. Let me know if I should split those entirely though.

brandur-stripe:
LGTM.

brandur-stripe:
(And thanks for the detailed change explanation too! This whole PR is a ton of work — you're on a roll.)

Issue #1024, opened 2017-10-10
StripeInvoiceLineItem doesn't contain the Discountable field


martijn-broekstra:
See title.

remi-stripe:
Thanks for the report! I've pushed a fix for this in https://github.com/stripe/stripe-dotnet/pull/1025 and we will likely ship this in the next few weeks!

PR #1025, opened 2017-10-10
Add support for discountable on line item

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/1024


brandur-stripe:
Killing it as usual Remi, thanks!

PR #1026, opened 2017-10-10
Experiment with Markdown changelog

brandur:
Here we experiment a bit with our changelog format by changing
`CHANGELOG` to `CHANGELOG.md` and refactoring.

The new format is largely based on keepachangelog.com [1], but I've
streamlined it a little bit (by removing subsubheadings) because I find
the default a little tedious to keep up-to-date and visually noisy
enough that it's hard to skim.

Advantages of the new format:

* Linkable headers
* Nice in-browser rendering (standard Markdown)
* Easy to copy and paste additions right into the releases page when
  cutting a new release
* Real links to all PRs that introduced changes
* ~~Text wrapping when in plaintext for readability that way too~~ Actually, it's a little strange but the releases page doesn't render Markdown like other places in GitHub and will not concatenate wrap lines. Given that, it's probably easiest to set no line wrapping for `CHANGELOG.md` so that it's still easy to copy and paste.

[Here's a link to the rendered version](https://github.com/stripe/stripe-dotnet/blob/brandur-md-changelog/CHANGELOG.md).

@remi-stripe What do you think? If we can find a format that works here, I'm
thinking we just move all our libraries over to look exactly the same.

[1] http://keepachangelog.com/en/0.3.0/


remi-stripe:
@brandur-stripe I wonder if it really is worth linking to the PRs? That sounds like a lot of manual work for very little benefit since it's easy to find the corresponding PR in most cases even a year later. What do you think? I'm okay with any approach as long as all libraries are moved to the format afterwards.

brandur-stripe:
> I wonder if it really is worth linking to the PRs? That sounds like a lot of manual work for very little benefit since it's easy to find the corresponding PR in most cases even a year later. What do you think?

Hm, I'm cautiously optimistic that this won't be too much extra work. In practice, what you can do is just copy an existing line and change the numbers around, which is relatively lightweight. I think it does add _some_ value at least too since a lot of these lines are quite terse and probably don't give you as much information as you'd ideally want when reading through it.

What do you think about trying it for a while and if it turns out to be a major pain we can just revert? We can do here initially, then move whatever we decide over to other libraries.

> I'm okay with any approach as long as all libraries are moved to the format afterwards.

Yeah, +1. Let's decide the one true way here, and then distribute everywhere.

brandur-stripe:
Remi, going to merge this one so I don't have to keep rebasing it.

Luckily changing the format of the changelog is not breaking, so we can tweak it in the future if we find it not to our liking.

brandur-stripe:
> Is it possible to ... like, generate these from Git history, or from the Github API? I mean, I know it's possible to generate a changelog from those sources, but would this work for us? I.e., having a script that the 'release maintainer' would run to (re)generate this in the appropriate format prior to tagging?

@jlomas-stripe Yeah that might be kind of cool. IMO, it would be best if a human still vetted the content because a lot of the time PR titles (which would presumably be the main source of the content) can be worded somewhat poorly and we should try to keep the quality of content in the changelog relatively high.

A system that used GitHub's API to examine merged PRs since the last release and then seeded the changelog file with a list of them before invoking `$EDITOR` for the releaser would be pretty cool.

However that said, I don't personally find this to be too much work, and I'm not sure that spending the time to automate the process would yield a very payoff in terms of time spent automated versus time saved. If you're interested though, feel free to take a look at our internal `bindings` repo (where all the current automation is happening) — that should serve as a good launchpad for the feature.

jlomas-stripe:
Totally makes sense - I tend to like to solve problems in ~overly complicated and often unnecessary~ interesting ways. :)

Also /++ on resurrecting my deleted comment. :joy:

PR #1028, opened 2017-10-11
Support for attaching and detaching sources to/from customers

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @stan-stripe @will-stripe 

Adds support for attaching and detaching source objects to/from customer objects.

Attaching a source looks like this:

```csharp
var SourceAttachOptions = new StripeSourceAttachOptions
{
  Source = "src_..."
};

var sourceService = new StripeSourceService();
StripeSource SourceAttached = sourceService.Attach("cus_...", SourceAttachOptions);
```

Detaching a source looks like this:

```csharp
var sourceService = new StripeSourceService();
StripeSource SourceDetached = sourceService.Detach("cus_...", "src_...");
```

### DIFF
--- a/src/Stripe.Tests.XUnit/sources/attaching_and_detaching_sources.cs
+++ b/src/Stripe.Tests.XUnit/sources/attaching_and_detaching_sources.cs
@@ -1,0 +1,64 @@
+using FluentAssertions;
+using Xunit;
+
+namespace Stripe.Tests.Xunit
+{
+    public class attaching_and_detaching_sources
+    {
+        public StripeCustomer Customer { get; }
+        public StripeSource SourceAttached { get; }
+        public StripeSource SourceDetached { get; }
+
+        public attaching_and_detaching_sources()
+        {
+            var SourceCardCreateOptions = new StripeSourceCreateOptions
+            {
+                Type = StripeSourceType.Card,
+                Token = "tok_visa"
+            };
+
+            var CustomerCreateOptions = new StripeCustomerCreateOptions
+            {
+                Email = "source-attach-detach@example.com",
+            };
+
+            var sourceService = new StripeSourceService(Cache.ApiKey);
+            var customerService = new StripeCustomerService(Cache.ApiKey);
+
+            var SourceCard = sourceService.Create(SourceCardCreateOptions);
+            Customer = customerService.Create(CustomerCreateOptions);
+
+            var SourceAttachOptions = new StripeSourceAttachOptions
+            {
+                Source = SourceCard.Id
+            };
+
+            SourceAttached = sourceService.Attach(Customer.Id, SourceAttachOptions);
+            SourceDetached = sourceService.Detach(Customer.Id, SourceAttached.Id);
+        }
+
+        [Fact]
+        public void source_attached_should_not_be_null()
+        {
+            SourceAttached.Should().NotBeNull();
+        }
+
+        [Fact]
+        public void source_attached_should_have_the_correct_customer_id()
+        {
+            SourceAttached.Customer.Should().Be(Customer.Id);
+        }
+
+        [Fact]
+        public void source_detached_should_not_be_null()
+        {
+            SourceDetached.Should().NotBeNull();
+        }
+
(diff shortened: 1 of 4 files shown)


remi-stripe:
@ob-stripe I wonder if we shouldn't approach this the opposite way here. The Source gets attached _to a customer_. Shouldn't we pass the Source id for attach and the Customer id as a parameter instead?

```
var SourceAttachOptions = new StripeSourceAttachOptions
{
  Customer = "cus_..."
};

var sourceService = new StripeSourceService();
StripeSource SourceAttached = sourceService.Attach("src_...", SourceAttachOptions);
```

ob-stripe:
From what little I've seen of the library, the convention is to pass the URI components as individual string parameters, and the parameters in a single `StripeObjectMethodOptions` object.

In this case, we're sending a POST request to '/v1/customers/{customerId}/sources` with a single POST parameter `source={sourceId}`, but the method might have more parameters in the future and we'd only have to add those to `StripeSourceAttachOptions`.

remi-stripe:
I see. It still feels quite backwards to me and unnatural. We're managing a Source, the main object should be the one whose id is passed first. The rest should be treated as parameters. That's what you did for `detach` on stripe-go [here](https://github.com/stripe/stripe-go/blob/master/source/client.go#L91).

The "right" flow in my eyes would be for `/v1/sources/src_XXXX` to support the `customer` parameter instead so that this makes more sense in all the libraries though I don't think this will happen.

I don't feel too strongly and you can ignore if you disagree. It felt important to raise because we're likely continuing to make Sources even more confusing over time with this approach.

ob-stripe:
I'd rather keep the current version to keep the code straightforward. Otherwise we'd have to extract the customer ID from the options object to form the URL, and add the source ID to the options object to form the POST parameters. It's also consistent with e.g. creating a card (even though I agree that "creating a card" and "attaching a source" are not 1:1 comparable operations).

For the detach method, I took a look at existing methods for nested resources (e.g. [`StripeApplicationFeeRefundService.Update`](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/ApplicationFeeRefunds/StripeApplicationFeeRefundService.cs#L23)) and it seems the convention is to have the URI components appear in the same order as they do in the URL (in this case, `/v1/customers/{customerId}/sources/{sourceId}`).

brandur-stripe:
No strong opinion here (I still don't know the C# conventions very well), but it seems like Olivier has done his homework. @remi-stripe Maybe add your thoughts on that last response when you get a chance ^

One thing I noticed is that in most libraries attaching a source is done by [what looks like creating a source for an existing customer](https://stripe.com/docs/sources/customers#attaching-a-source-to-an-existing-customer), and by introducing an "attach" verb for C#, we'd be deviating from this convention a little bit. I'm not sure if that's a problem or not.

remi-stripe:
I'm fine with the approach taken here. It's easy to undo in the future as most release, like Go, could be major anyway. I don't think this is likely to change in the next few months anyway.

ob-stripe:
> One thing I noticed is that in most libraries attaching a source is done by what looks like creating a source for an existing customer, and by introducing an "attach" verb for C#, we'd be deviating from this convention a little bit.

Yeah, but I couldn't reuse that here. Here's how you'd [create a card](https://stripe.com/docs/api/dotnet#create_card) for instance:

```csharp
var cardOptions = new StripeCardCreateOptions() {
  SourceToken = "tok_visa"
};

var cardService = new StripeCardService();
StripeCard card = cardService.Create("cus_...", cardOptions);
```

E.g. you'd call `StripeCardService.Create`, or `StripeBankAccountService.Create` for a bank account. However, I couldn't reuse that convention for sources because there is already a `StripeSourceService.Create` method for actually creating a source. Since for sources, creating a source and attaching it to a customer are two different operations, I had to use a different method.

Happy to revisit if anyone has a better idea though!

brandur-stripe:
Cool. Thanks for explaining the design process Olivier! That works for me.

Issue #1027, opened 2017-10-11
Coupon GET and DELETE... need to explicitly UrlEncode the coupon id


rmacfadyen:
Coupon delete (and get) appear to need to have the coupon code explicitly UrlEncode'd:

I have a coupon defined called "ABC123-1" and the code below is going to delete all coupons. It was failing with StripeException: No such coupon... and poking at it I realized the problem was lack of urlencoding:

    var s = new StripeCouponService(ConfigurationManager.AppSettings["StripePrivateKey"]);

    var l = s.List();
    foreach (var c in l)
    {
        // succeeds
        var n = s.Get(System.Net.WebUtility.UrlEncode(c.Id));

        // fails with StripeException: No such coupon
        s.Delete(c.Id);
    }

Googling for "stripe.net need to explicitly urlencode id values" didn't return anything useful.

Is it up to the dev to UrlEncode id values, in general, across the entire API?

remi-stripe:
@rmacfadyen I tried to reproduce and used the exact same coupon id and then some more complex and I was always able to retrieve or delete the coupon without having to URL encode the ID.

Do you have more details about your issue and potentially a simple test case that reproduces the exact issue end to end from coupon creation to deletion? Also, are you sure that you're not on an old version of the library?

rmacfadyen:
Ah hah!

The problem turned out to be that when creating the coupon my Id had spaces at the end (eg. "SZIMEK-01{space}{space}" (why the db uses char instead of varchar I refuse to answer on the grounds it may make me look foolish).

So a bit of trimming and things are fine.

However... something's "not right". Why should the API or the client care that there are spaces at the end? I haven't dug down to figure out if it's the API or the client... sorry.

Thanks for pointing me in the right direction :)

remi-stripe:
@rmacfadyen I think I found the issue. If you create a coupon in the dashboard with spaces at the end, those are simply stripped. On the other hand, if you do it in code, the spaces are kept.

When you try to retrieve the coupon or delete it though, the id is passed in the URL but it's not URL encoded. Because of this fetching the coupon fails.

For now I would recommend that you simply strip your coupon ids so that there are no trailing spaces at the end, which was likely a mistake in the first place. I'll look into fixing this in the library moving forward though it's an involved change so it might take some time.

rmacfadyen:
I agree 100%. Trims all around :)

Thanks for the assist!

remi-stripe:
[...]

PR #1029, opened 2017-10-11
Properly encode the resource id when sending those in the URL

remi-stripe:
Most of our APIs pass a given object's id in the URL. For example to retrieve a coupon you use `GET /v1/coupons/COUPON_ID` and to delete a plan you use `DELETE /v1/plans/PLAN_ID`. At the moment though, the library passes the raw string directly in the URL instead of properly encoding it as we do in other libraries such as stripe-go [here](https://github.com/stripe/stripe-go/blob/master/coupon/client.go#L57).

Since the Plan and Coupon resources' ids can be chosen by the merchant, those have to be properly URL-encoded before being used. This PR fixes the issue for the Coupon and Plan resources after the bug report in https://github.com/stripe/stripe-dotnet/issues/1027.

This also adds tests for Plan and Coupon to the XUnit test suite.

This is likely a **breaking change** as users might have already fixed this in their own code by URL-encoding the id.

### DIFF
--- a/src/Stripe.Tests.XUnit/coupons/_fixture.cs
+++ b/src/Stripe.Tests.XUnit/coupons/_fixture.cs
@@ -1,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+
+namespace Stripe.Tests.Xunit
+{
+    public class coupons_fixture : IDisposable
+    {
+        public StripeCouponCreateOptions CouponCreateOptions { get; set; }
+        public StripeCouponUpdateOptions CouponUpdateOptions { get; set; }
+
+        public StripeCoupon Coupon { get; set; }
+        public StripeCoupon CouponRetrieved { get; set; }
+        public StripeCoupon CouponUpdated { get; set; }
+        public StripeDeleted CouponDeleted { get; set; }
+        public StripeList<StripeCoupon> CouponsList { get; }
+
+        public coupons_fixture()
+        {
+            CouponCreateOptions = new StripeCouponCreateOptions() {
+              Id = "test-coupon-" + Guid.NewGuid().ToString(),
+              PercentOff = 25,
+              Duration = "repeating",
+              DurationInMonths = 3,
+            };
+
+            CouponUpdateOptions = new StripeCouponUpdateOptions {
+              Metadata = new Dictionary<string, string>{{"key_1", "value_1"}}
+            };
+
+            var service = new StripeCouponService(Cache.ApiKey);
+            Coupon = service.Create(CouponCreateOptions);
+            CouponRetrieved = service.Get(Coupon.Id);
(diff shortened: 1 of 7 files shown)


brandur-stripe:
Nice Remi! Thanks for adding those extra tests too.

> fixing each Service to properly URL encode the resource id as needed

I think the way you have it is the way to go. I'm not sure if you're with me here, but IMO the way we built coupons so that they could be referenced in URL with a user-supplied ID was probably a mistake — what we probably should have is have Stripe-generated IDs as the primary identifier and a user-generated one as a secondary field on each object. It's probably not worth changing now, but going forward I'm hoping that we'll avoid copying this pattern elsewhere and instead just use Stripe-generated (and by extension URL-safe) identifiers all over the place.

> Also, is WebUtility.UrlEncode(value) the correct approach here? That's what we use in the code here but I'm not sure it's the right method to call.

I'm not 100% sure, but yes I think so. I notice that there's also [`HttpServerUtility.UrlEncode`](https://msdn.microsoft.com/en-us/library/zttxte6w(v=vs.110).aspx) and [`HttpUtility.UrlEncode`](https://msdn.microsoft.com/en-us/library/4fkewx0t(v=vs.110).aspx), but they come from the `System.Web` assembly which has markedly less usage throughout this library. `WebUtility.UrlEncode` comes from `System.Net`, which is referenced in quite a few places already.

Anyway, I'd say this PR looks pretty good if you want to merge it.

remi-stripe:
@brandur I can't merge as is as there are other resources where the ID can be chosen by the user such as Plan. I'll make a pass on the whole list and what we do in stripe-go to see which ones are encoded now that we agreed on the right approach

brandur-stripe:
@remi-stripe Sounds good. Thanks!

jaymedavis:
Something else might be going on. If you trace that method down to the call made on the RequestStringBuilder, all parameters should be encoded. ApplyAllParamters ultimately ends up here, which does the same thing you are doing here iirc.

remi-stripe:
@jaymedavis The _parameters_ are encoded. The id _in the URL_ is not encoded. That's the issue since the id is chosen by the user.

jaymedavis:
👍 

ooc The original reason some values are in the service methods as arguments, is because they were required by Stripe. The _Options_ idea started because they were optional parameters. This starts to break down when lots of things are required though. hth

remi-stripe:
@jaymedavis I thought about it but we'd need to change the way we pass the resource id so that it's URL-encoded separately from the base URL. Also it's really specific to Coupon/Plan so a special-case is likely cleaner because it brings up the fact that it is an issue in the first place.

jaymedavis:
Makes sense! Great work @remi-stripe.

brandur-stripe:
Nice work Remi! (And good to see you still active on here Jayme :) This LGTM.

PR #1031, opened 2017-10-12
Adds support for subscription manual payments

### LINKED ISSUE #1030: How set billing property when create subscription ?
The model in StripeSubscriptionCreateOptions was not "billing" property. So how we can change "billing" ? 

        public DateTime? BillingCycleAnchor { get; set; }
        public bool BillingCycleAnchorNow { get; set; }
        public bool BillingCycleAnchorUnchanged { get; set; }
        [JsonProperty("customer")]
        public string CustomerId { get; set; }
        [JsonProperty("subscription_items_array")]
        public List<StripeSubscriptionItemOption> Items { get; set; }

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds support for the new attributes and parameters for [subscription manual payments](https://stripe.com/docs/subscriptions/ach-credit).

Fixes #1030.


src/Stripe.net/Services/Customers/StripeCustomerCreateOptions.cs
```diff
@@ -10,12 +10,18 @@ public class StripeCustomerCreateOptions
         [JsonProperty("account_balance")]
         public int? AccountBalance { get; set; }
 
+        [JsonProperty("billing")]
```
remi-stripe: This is a new feature. IMO we should not support this on Custom creation/update. We should only support this on the Subscription resource. That's what we do in stripe-go

src/Stripe.net/Constants/StripeBilling.cs
```diff
@@ -0,0 +1,8 @@
+namespace Stripe
+{
+    public static class StripeBilling
+    {
+        public const string ChargeAutomatically = "charge_automatically";
+        public const string SendInvoice = "send_invoice";
+    }
```
brandur-stripe: Should we make this an actual `enum` for better type safety? I see that there is some precedent for this in the form of `SourceType` in `src/Stripe.net/Entities/Source.cs`.

brandur-stripe:
Great work @ob-stripe! I left one comment above, and it looks like the test failures are related to the change. After those are addressed, I think we're good to ship. Thanks!

src/Stripe.net/Constants/StripeBilling.cs
jaymedavis: +1. To get this working, you would have to create a JsonConverter for each enum. Then decorate the entity with the converter. The converter would map the enum to the string it represents.

brandur-stripe: Thanks Jayme!

I was just Googling around out of curiosity, and came across [this Stack Overflow answer on a very related subject](https://stackoverflow.com/questions/26391632/how-to-serialize-enums-to-different-property-name-using-json-net). It looks like we may be able to get an event easier solution using something like `[EnumMember(Value = "charge_automatically")]`! (I didn't try it, but probably worth seeing what will happen.)

jaymedavis: Nice! As long as netstandard plays nice with this (I don’t see why it wouldn’t), you should be able to simply replace the class with a decorated enum, and change the entity property type to that enum. 

Good one!

ob-stripe: I think this is out of scope for this PR, but I'll try taking a stab at converting all the existing string constants used for parameters into enums in a separate PR.

(I think the existing `SourceType ` enum is used for a different purpose -- to distinguish between the different types of sources that can be attached to a customer.)

ob-stripe:
Apparently the API was rejecting the new parameters, but I cannot figure out why. That no longer seems to be the case -- I haven't changed anything but tests seem to be passing now 😕

src/Stripe.net/Constants/StripeBilling.cs
brandur-stripe: Fair enough. One more thought: I'm fairly certain this is probably the right way to go overall given that it forces people to use the type instead of just providing a custom string, so it might be okay to do just this one and then put a task in to convert everything over eventually too.

But cool, we'll need a breaking change for the others anyway, so up to you.

ob-stripe:
Okay, you convinced about using an enum instead :) `StripeBilling` is now an `Enum`.

The `[EnumMember(Value = "...")]` annotation makes it very easy to deserialize JSON into enums, but I had to add some custom code (`EnumPlugin.cs`) to serialize enums into URL-encoded values.

I tried a bunch of things but couldn't find a way to access the `EnumMemberAttribute`'s `Value` using reflection. Eventually I decided to just use the JSON converter for this since it's already capable of it. It returns the string wrapped in quotes so I trim those to remove them from the final string used to form the parameter. Not super elegant, but it works, and it will automatically apply to every `Enum` from now on so converting the other constants into enums should be straightforward.

brandur-stripe:
Haha, sorry Olivier, I'd forgotten about JSON decoding and didn't expect this to be that much of a yak shave :)

> I tried a bunch of things but couldn't find a way to access the EnumMemberAttribute's Value using reflection. Eventually I decided to just use the JSON converter for this since it's already capable of it. It returns the string wrapped in quotes so I trim those to remove them from the final string used to form the parameter. Not super elegant, but it works, and it will automatically apply to every Enum from now on so converting the other constants into enums should be straightforward.

Nice work! One minor request: do you think we might be able to put in a unit test for `EnumPlugin` just to do a very basic verification that it works (and also maybe to check what happens in the case that we receive an unrecognized value from the API that's not part of the enum we're trying to decode)?

ob-stripe:
To clarify: JSON decoding works fine and is all handled automatically by `Newtonsoft.Json`. The new plugin is for _encoding_ values when sending URL-encoded parameters to the API.

Nevertheless, I added some tests for both encoding and decoding in `encoding_and_decoding_enums.cs`.

src/Stripe.Tests.XUnit/_infrastructure/encoding_and_decoding_enums.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        [Fact]
+        public void encoding_should_convert_to_string_if_no_enummember_value()
+        {
+            var obj = new TestObject
+            {
+                Enum = TestEnum.TestTwo,
+            };
+            var url = new TestService().ApplyAllParameters(obj, "", false);
+            url.Should().Be("?enum=TestTwo");
```
brandur-stripe: Should we throw an error here? It seems that if we forgot an `EnumMember`, it was probably an accident that we should try to find out about before it starts causing problems.

brandur-stripe:
> To clarify: JSON decoding works fine and is all handled automatically by Newtonsoft.Json. The new plugin is for encoding values when sending URL-encoded parameters to the API.
>
> Nevertheless, I added some tests for both encoding and decoding in encoding_and_decoding_enums.cs.

Ah, I see! Thanks for the quick work. This is starting to look pretty nice.

src/Stripe.Tests.XUnit/_infrastructure/encoding_and_decoding_enums.cs
ob-stripe: That makes sense, but it's not possible with the current implementation as I'm actually delegating the encoding to `JsonConverter` ([here](https://github.com/stripe/stripe-dotnet/blob/3f78d8bf6ca49b3c9fb5c5ca71e91c57b9844a38/src/Stripe.net/Infrastructure/Middleware/ParserPlugin/EnumPlugin.cs#L20)) which will transparently fall back to using the constant name as a string if no `EnumMember` is present.

It would probably be better to use reflection to fetch the `EnumMember`'s `Value` ourselves, and that's what I tried to do initially but I couldn't manage it with my (basically inexistent) .NET skills.

Happy to revisit this, or delegate to someone with experience in dealing with reflection in .NET.

brandur-stripe: Cool. Alright, I don't think it's that big of a deal. Thanks for the explanation.

brandur-stripe:
Released as 11.4.0.

src/Stripe.Tests.XUnit/_infrastructure/encoding_and_decoding_enums.cs
jaymedavis: What you have works, but I think [this is what you are looking for](https://gist.github.com/jaymedavis/b1d773064d860dec0408fffe2ed30a5b)?

Issue #993, opened 2017-08-25
Follow secure coding practices 


bayotop:
There are two minor issues in the `StripeEventUtility.computeSignature()` method (https://github.com/stripe/stripe-dotnet/blob/ff283e7c50914e2847ecf9fd70b896950fe87ded/src/Stripe.net/Services/Events/StripeEventUtility.cs):

1. The `HMACSHA256` object should be explicitly disposed. It is important to dispose cryptographic objects ASAP given that these objects wipe internal buffers on disposal, which prevents sensitive data from sitting in memory indefinitely. Note that `HMACSHA256` has no finalizer so the GC may not ever trigger a disposal.
2. The second issue is the default behaviour of `Encoding.UTF8.GetBytes(string)`. Unicode codepoints without a valid UTF-8 representation will by default fallback and replace the failed mapping with "?" (See https://msdn.microsoft.com/en-us/library/system.text.encoding.utf8(v=vs.110).aspx). The problem can be seen in the following example: 
```csharp
string secret = "\ud801";
string secret2 = "\ud802";

if (!secret.Equals(secret2))
{
    Console.Out.WriteLine("The strings are not equal.");
}

computeSignature(secret, "timestamp", "payload");  // returns 1eba7aac5f10ee7aa985722256aa4125e8e59fe256386c8ab10295063d986e57
computeSignature(secret2, "timestamp", "payload"); // returns 1eba7aac5f10ee7aa985722256aa4125e8e59fe256386c8ab10295063d986e57
```

The example shows that the method calculated the same signature for two different keys. The reason is that both secrets became "?" because the `GetBytes()` method did silently fallback.

Both issues can be fixed quite easily. In the first case simply wrapping the `HMACSHA256` object into a using block will do the trick. In the latter case you should use a `UTF8Encoding` instance that throws an exception on failed mappings to prevent the situation that identical signatures are generated for different keys

brandur-stripe:
Thanks again for sticking through this @bayotop. We've released your fix as part of v11.6.0.

ob-stripe:
Closed by #1035.

PR #1035, opened 2017-10-16
follow secure coding BP, fix (#993)

bayotop:
I created a fresh PR for #993. I am quite certain the changes shouldn't affect other tests so I am not sure why they are still failing because of a missing API key. 

The tests related to constructing the event are passing. Please let me know if there is anything I can do. Thanks.

### DIFF
--- a/src/Stripe.Tests.XUnit/events/_fixture.cs
+++ b/src/Stripe.Tests.XUnit/events/_fixture.cs
@@ -1,0 +1,22 @@
+﻿using System.IO;
+using System.Reflection;
+using System.Text;
+
+namespace Stripe.Tests.Xunit
+{
+    public class events_fixture
+    {
+        // this was sent in the header Stripe-Signature along with the body of event.json in a test webhook
+        internal string StripeSignature => "t=1493329224,v1=dca14f723dfa3bf47a310c3d6c3aff8bdb2534263d051dd2613ece097b8bdea4,v0=63f3a72374a733066c4be69ed7f8e5ac85c22c9f0a6a612ab9a025a9e4ee7eef";
+        internal string StripeJson { get; set; }
+        internal string StripeSecret => "whsec_H68eTX02a4bCbiQOOoSAsIytOvuWZrQC";
+
+        public events_fixture()
+        {
+            StripeJson = new StreamReader(
+                typeof(when_constructing_an_event).GetTypeInfo().Assembly.GetManifestResourceStream("Stripe.Tests.XUnit.events.event.json"),
+                Encoding.UTF8
+            ).ReadToEnd();
+        }
+    }
+}
--- a/src/Stripe.Tests.XUnit/events/when_constructing_an_event.cs
+++ b/src/Stripe.Tests.XUnit/events/when_constructing_an_event.cs
@@ -1,46 +1,76 @@
-﻿using System.IO;
-using System.Reflection;
-using System.Text;
+﻿using System;
 using FluentAssertions;
 using Xunit;
 
 namespace Stripe.Tests.Xunit
 {
-    public class when_constructing_an_event
+    public class when_constructing_an_event : IClassFixture<events_fixture>
     {
-        // this was sent in the header Stripe-Signature along with the body of event.json in a test webhook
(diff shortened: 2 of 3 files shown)


ob-stripe:
Ah, I believe I understand what's happening. The `STRIPE_TEST_SK` environment variable is defined as a "secure variable" in our [`appveyor.yml`](https://github.com/stripe/stripe-dotnet/blob/a43bc26f5d91d94e049bfc9c307b017f4b1c3ded/appveyor.yml#L5-L6) configuration file.

From https://www.appveyor.com/docs/build-configuration/#secure-variables:
> However, secure variables are not decoded during Pull Request builds which prevents someone from submitting PR with malicious build script displaying those variables. In more controlled environment through with a trusted team and private GitHub repositories there is an option on General tab of project settings to allow secure variables for PRs.

We should probably just define the key as a regular variable -- the "secret" API key for the test account we use for our libraries is visible in plain text in many of our other language libraries (e.g. [here](https://github.com/stripe/stripe-python/blob/ef50aa5e96b05a47f4b277ac6eab76e336860b99/stripe/test/helper.py#L132-L133) in stripe-python).

src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
```diff
@@ -27,4 +27,8 @@
     <ProjectReference Include="..\Stripe.net\Stripe.net.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+  </ItemGroup>
```
brandur-stripe: Do you know why these ended up in here? Can we pull them out of the branch for hygiene?

src/Stripe.net.Tests/Stripe.net.Tests.csproj
```diff
@@ -49,4 +49,8 @@
     <Reference Include="System" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+  </ItemGroup>
```
brandur-stripe: Same with this one.

brandur-stripe:
Nice one @bayotop! Thanks again. I left a couple minor comments, but this looks great to me.

@ob-stripe Would you mind taking a review pass as well? You know this code a lot better than I do.

> Ah, I believe I understand what's happening. The STRIPE_TEST_SK environment variable is defined as a "secure variable" in our appveyor.yml configuration file.
>
> We should probably just define the key as a regular variable -- the "secret" API key for the test account we use for our libraries is visible in plain text in many of our other language libraries (e.g. here in stripe-python).

Ah yep, Travis does this too :/ I don't think there's a major problem putting a secret testmode key into here as plaintext — it does allow someone to potentially DOS our builds, but the advantages of having successful builds from third-party contributors outweights that by a fair bit.

In the meantime, we can just repush @bayotop's branch to our own repository just to get this unblocked.

src/Stripe.net.Tests/Stripe.net.Tests.csproj
bayotop: Good point. Those were added during the build. However, this isn't the case anymore and it seems they can be safely removed. I'll amend the commit.

bayotop:
I'll amended the commit according to you comments and aligned with master. Feel free to proceed with the repush. Thanks!

bayotop:
My event.json file had Windows line endings so produced a different hash then expected, thus the 2 failed tests on appveyor/branch. Should be fixed now.

brandur-stripe:
> My local event.json file had Windows line endings so produced a different hash then expected, thus the 2 failed tests on appveyor/branch. Should be fixed now.

Thanks for taking a look at that Martin! I've repushed to our origin again and will get this in as soon as the build passes.

src/Stripe.net.Tests/Stripe.net.Tests.csproj
xt0rted: These are added by VS to indicate the given project is a test project and that the test runner should be loaded for it.

PR #1037, opened 2017-10-17
Alphabetize fields

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This PR reorders the field names on all entities to appear in the same order as in the API reference (e.g. `object` followed by the fields in alphabetical order).

To keep the PR simple, I haven't modified anything besides changing the order of the fields in the source, but I've added a few "TODO" comments to note fields that are currently missing and should be added.

### DIFF
--- a/src/Stripe.net/Entities/StripeError.cs
+++ b/src/Stripe.net/Entities/StripeError.cs
@@ -4,28 +4,32 @@ namespace Stripe
 {
     public class StripeError : StripeEntity
     {
-        [JsonProperty("type")]
-        public string ErrorType { get; set; }
+        // For regular API errors:
 
-        [JsonProperty("message")]
-        public string Message { get; set; }
+        [JsonProperty("charge")]
+        public string ChargeId { get; set; }
 
         [JsonProperty("code")]
         public string Code { get; set; }
 
+        [JsonProperty("decline_code")]
+        public string DeclineCode { get; set; }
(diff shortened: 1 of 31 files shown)


src/Stripe.net/Entities/StripeError.cs
```diff
@@ -4,28 +4,32 @@ namespace Stripe
 {
     public class StripeError : StripeEntity
     {
+        // For regular API errors:
+
         [JsonProperty("type")]
         public string ErrorType { get; set; }
 
+        [JsonProperty("charge")]
+        public string ChargeId { get; set; }
```
brandur-stripe: Maybe put `ChargeId` above `ErrorType` and push `Message` down by a few parameters?

brandur-stripe:
One minor comment, but otherwise LGTM. I only reviewed the diff in this pull request which has relative changes but in some many cases left out the absolute context, so there's a chance that I didn't catch something. That's okay though, this is the kind of thing that's easy to fix later.

Looks like a ton of missing fields dropped out of this audit too! Nice one again Olivier!

src/Stripe.net/Entities/StripeError.cs
ob-stripe: Fixed!

brandur-stripe:
Thanks!

PR #1044, opened 2017-10-24
Improve parameter encoding

### LINKED ISSUE #1042: Metadata keys with spaces and special characters
It seems like the library is not correctly URL-encoding metadata keys when sending requests.

At first glance I think this is because the `argument` parameter is not passed to `UrlEncode` [here](https://github.com/stripe/stripe-dotnet/blob/cf95812d14c39efad38747e8ab3d2ecb4c8204d6/src/Stripe.net/Infrastructure/Middleware/RequestStringBuilder.cs#L54).

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fixes #1042.

This PR adds general improvements to parameter encoding:

- dictionary keys are now properly URL-encoded

- all dictionaries are now automatically correctly encoded (got rid of the special-casing based on parameter names in `DictionaryPlugin`)

- all arrays are now automatically correctly encoded (got rid of the special-casing based on prefixing the parameter names with `array:`)

- added some tests

### DIFF
--- a/src/Stripe.net/Infrastructure/Middleware/ParserPlugin/DictionaryPlugin.cs
+++ b/src/Stripe.net/Infrastructure/Middleware/ParserPlugin/DictionaryPlugin.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Reflection;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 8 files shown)


src/Stripe.net/Infrastructure/Middleware/ParserPlugin/DictionaryPlugin.cs
```diff
@@ -9,13 +10,25 @@ internal class DictionaryPlugin : IParserPlugin
     {
         public bool Parse(ref string requestString, JsonPropertyAttribute attribute, PropertyInfo property, object propertyValue, object propertyParent)
         {
-            if (!attribute.PropertyName.Contains("metadata") && !attribute.PropertyName.Contains("fraud_details") && !attribute.PropertyName.Contains("attributes")) return false;
+            // Check if the property is a Dictionary
+            var type = property.PropertyType;
+            if (!type.GetTypeInfo().IsGenericType) return false;
+            if (type.GetTypeInfo().GetGenericTypeDefinition() != typeof(Dictionary<,>)) return false;
+
+            // Ensure that key and value types are both string
```
ob-stripe: There's probably a way to be able to encode any `Dictionary<object, object>` by calling `ToString()` on keys and values, but I couldn't manage with my meager .NET reflection skills, so I'm just throwing an exception if the key and/or value type is not `string`.

It's not a big deal as all dictionaries in the library are `<string, string>` right now and the explicit exception should prevent contributors from mistakenly using other types.

brandur-stripe: Honestly, I think some explicitness is better here than more reflection magic :)

+1 on the `ArgumentException`s.

brandur-stripe:
Great work as usual OB :) That cleanup around arrays is top notch.

PR #1050, opened 2017-10-31
Make "exchange rate" singular

brandur:
We've recently tweaked the very new exchange rates API somewhat so that
a single "exchange rate" resource becomes singular instead of plural.
This is so that it falls into convention with all the other resources in
the API.

I jumped the gun a little bit of merging the .NET pull request, so this
one goes back through and amends a few places that were using a plural
exchange rate so that they're now singular.

Apparently we got a little lucky because 11.7.0 (the version I sent this
out under) apparently never made it to Nuget (possibly a build failure
which I'll have to look into separately). But anyway, it's safe to
release this without a breaking change because there is no official
package release with the other API.

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.Tests.XUnit/exchange_rates/when_getting_an_exchange_rate.cs
+++ b/src/Stripe.Tests.XUnit/exchange_rates/when_getting_an_exchange_rate.cs
@@ -7,13 +7,13 @@ using Xunit;
 
 namespace Stripe.Tests.XUnit
 {
-    public class when_getting_exchange_rates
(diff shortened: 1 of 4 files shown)


ob-stripe:
lgtm! Thanks for handling.

brandur-stripe:
Thanks Olivier.

Released as 11.7.1. I'm pretty sure I found the problem that was causing the build failure, so with any luck this version will make it over to Nuget.

PR #1054, opened 2017-11-14
Default to application/octet-stream

### LINKED ISSUE #1053: Error thrown when filename passed to FileUploadService doesn't match a known extension
This code doesn't check the return value from the MimeTypes.GetMimeType(filename)—which causes the .NET framework to throw an exception when a null mimeType is passed:

https://github.com/stripe/stripe-dotnet/blob/96cdbaffa5609ab6f8c4925342247c5a8be6f0ac/src/Stripe.net/Infrastructure/MimeTypes.cs

You can test this by passing an empty filename, or one without an exception—either should trigger the exception.

ob-stripe:
r? @fred-stripe 
cc @stripe/api-libraries 

Fixes #1053.

We could even remove `MimeTypes.cs` entirely and just use `application/octet-stream` always. This is what the Python library does ([here](https://github.com/stripe/stripe-python/blob/f948b8b95b6df5b57c7444a05d6c83c8c5e6a0ac/stripe/multipart_data_generator.py#L32)).

I've also set a default value for the filename, as the API does require one (cf. https://github.com/stripe/stripe-python/issues/227 for more context).

### DIFF
--- a/src/Stripe.net/Infrastructure/Requestor.cs
+++ b/src/Stripe.net/Infrastructure/Requestor.cs
@@ -179,6 +179,11 @@ namespace Stripe.Infrastructure
         {
             requestMessage.Headers.ExpectContinue = true;
 
+            if (string.IsNullOrEmpty(fileName))
+            {
+                fileName = "blob";
+            }
+
             var fileContent = new StreamContent(fileStream);
             fileContent.Headers.ContentDisposition = new ContentDispositionHeaderValue("form-data")
             {
--- a/src/Stripe.net/Infrastructure/MimeTypes.cs
+++ b/src/Stripe.net/Infrastructure/MimeTypes.cs
@@ -19,7 +19,7 @@ namespace Stripe.Infrastructure
                     return "image/png";
 
                 default:
-                    return null;
+                    return "application/octet-stream";
             }
         }
     }


remi-stripe:
@fred-stripe friendly bump on this one

src/Stripe.net/Infrastructure/Requestor.cs
```diff
@@ -179,6 +179,11 @@ private static void ApplyMultiPartFileToRequest(HttpRequestMessage requestMessag
         {
             requestMessage.Headers.ExpectContinue = true;
 
+            if (string.IsNullOrEmpty(fileName))
+            {
+                fileName = "blob";
```
remi-stripe: should we be more explicit to indicate it's a hardcoded value? blob has some meaning for me as a type (like in a database) and so it felt weird the first 2 times I read the PR.

No big deal obviously but wanted to flag

ob-stripe: "blob" is used with more or less the same meaning as in a database: it indicates that what follows is a binary large object. It's also the convention we already use in the [Node](https://github.com/stripe/stripe-node/blob/a7382226965ee399ce6b5b9fd162c5af74c8131d/lib/MultipartDataGenerator.js#L30) and [Python](https://github.com/stripe/stripe-python/blob/f948b8b95b6df5b57c7444a05d6c83c8c5e6a0ac/stripe/multipart_data_generator.py#L24) libraries.

remi-stripe: gotcha, works for me if we do this in the other libs!

remi-stripe:
LGTM

PR #1055, opened 2017-11-14
Properly encode DueDate

### LINKED ISSUE #1052: Invalid Integer Due Date
Hi,

I am trying to create an invoice and I populated the due date field with a date. 

Here is my code:

```csharp
            StripeInvoiceService service = new StripeInvoiceService(m_UserName);
            StripeInvoice invoice = null;

            try
            {
                invoice = service.Create(customerId, inv, null);
            }
            catch (WebException ex)
            {
                using (var stream = ex.Response.GetResponseStream())
                using (var reader = new StreamReader(stream))
                {
                    string errorJson = reader.ReadToEnd();
                }
            }
            catch(Exception ex)
            {

            }
```

I receive an error saying the due date is an invalid integer.

ob-stripe:
r? @fred-stripe 
cc @stripe/api-libraries 

Fixes #1052.

`DueDate` was not properly encoded as a UNIX timestamp when creating an invoice. This is now fixed.

(We ought to write an encoding plugin to handle `DateTime`s everywhere transparently, but this is a quick fix to unblock #1052.)

### DIFF
--- a/src/Stripe.Tests.XUnit/invoices/creating_and_updating_invoices_with_manual_invoicing.cs
+++ b/src/Stripe.Tests.XUnit/invoices/creating_and_updating_invoices_with_manual_invoicing.cs
@@ -1,3 +1,4 @@
+using System;
 using FluentAssertions;
 using Xunit;
 
@@ -31,7 +32,7 @@ namespace Stripe.Tests.Xunit
             var InvoiceCreateOptions = new StripeInvoiceCreateOptions
             {
                 Billing = StripeBilling.SendInvoice,
-                DaysUntilDue = 7,
+                DueDate = DateTime.UtcNow.AddDays(10),
             };
             InvoiceCreated = invoiceService.Create(Customer.Id, InvoiceCreateOptions);
 
--- a/src/Stripe.net/Services/Invoices/StripeInvoiceCreateOptions.cs
+++ b/src/Stripe.net/Services/Invoices/StripeInvoiceCreateOptions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using Newtonsoft.Json;
+using Stripe.Infrastructure;
 
 namespace Stripe
 {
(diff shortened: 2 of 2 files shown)


src/Stripe.Tests.XUnit/invoices/creating_and_updating_invoices_with_manual_invoicing.cs
```diff
@@ -31,7 +32,7 @@ public creating_and_updating_invoices_with_manual_invoicing()
             var InvoiceCreateOptions = new StripeInvoiceCreateOptions
             {
                 Billing = StripeBilling.SendInvoice,
-                DaysUntilDue = 7,
+                DueDate = DateTime.Today.AddDays(7),
```
remi-stripe: minor but elsewhere we use `DateTime.UtcNow.AddDays(10);` (for example for RedeemBy [here](https://github.com/stripe/stripe-dotnet/blob/96cdbaffa5609ab6f8c4925342247c5a8be6f0ac/src/Stripe.net.Tests/coupons/test_data/stripe_coupon_create_options.cs#L16)).

Might be worth sticking with conventions (makes grepping for examples and similar use-cases easier)

Similarly, if you look at how `RedeemBy` is implemented, the code uses a `long?` instead

```
        public DateTime? RedeemBy { get; set; }

        [JsonProperty("redeem_by")]
        internal long? RedeemByInternal
        {
            get
            {
                if (!RedeemBy.HasValue) return null;

                return EpochTime.ConvertDateTimeToEpoch(RedeemBy.Value);
            }
        }
```

`TrialEnd` uses a string because they can handle the `"now"` value though

ob-stripe: Good points. I updated the test to use the same delay, and more importantly I updated the `DueDateInternal` variable to be a `long?` instead of a `String`.

PR #1057, opened 2017-11-16
Support for passing arbitrary parameters

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

This PR adds support for passing arbitrary parameters when making API requests. If the `Options` class inherit from `StripeParamsBase`, users can call `AddExtraParam` to add arbitrary key/value pairs to the request.

### DIFF
--- a/src/Stripe.net/Infrastructure/ParameterBuilder.cs
+++ b/src/Stripe.net/Infrastructure/ParameterBuilder.cs
@@ -10,7 +10,7 @@ namespace Stripe.Infrastructure
 {
     internal static class ParameterBuilder
     {
-        public static string ApplyAllParameters(this StripeService service, object obj, string url, bool isListMethod = false)
+        public static string ApplyAllParameters(this StripeService service, StripeBaseOptions obj, string url, bool isListMethod = false)
         {
             // store the original url from the service call into requestString (e.g. https://api.stripe.com/v1/accounts/account_id)
             // before the stripe attributes get applied. all of the attributes that will get passed to stripe will be applied to this string,
@@ -32,6 +32,12 @@ namespace Stripe.Infrastructure
                         else
                             RequestStringBuilder.ProcessPlugins(ref requestString, attribute, property, value, obj);
                     }
+                }
+
+                foreach (KeyValuePair<string, string> pair in obj.ExtraParams)
+                {
+                    var key = WebUtility.UrlEncode(pair.Key);
+                    RequestStringBuilder.ApplyParameterToRequestString(ref requestString, key, pair.Value);
                 }
             }
 
--- a/src/Stripe.net/Services/StripeBaseOptions.cs
+++ b/src/Stripe.net/Services/StripeBaseOptions.cs
@@ -1,0 +1,13 @@
+using System.Collections.Generic;
+
+namespace Stripe
+{
+    public class StripeBaseOptions
+    {
+        public void AddExtraParam(string key, string value) {
+            ExtraParams.Add(key, value);
+        }
+
+        public Dictionary<string, string> ExtraParams = new Dictionary<string, string>();
+    }
+}
--- a/src/Stripe.Tests.XUnit/_infrastructure/adding_extra_parameters.cs
+++ b/src/Stripe.Tests.XUnit/_infrastructure/adding_extra_parameters.cs
@@ -1,0 +1,48 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using FluentAssertions;
+using Xunit;
+using Stripe.Infrastructure;
+
+namespace Stripe.Tests.Xunit
+{
+    public class adding_extra_parameters
+    {
+        public class TestOptions : StripeBaseOptions
+        {
+            [JsonProperty("a_string")]
+            public string AString { get; set; }
+        }
+
+        public class TestService : StripeService
+        {
+            public TestService() : base(null)
+            {
+            }
+        }
+
+        public TestService Service { get; }
+
+        public adding_extra_parameters()
+        {
+            Service = new TestService();
+        }
+
+        [Fact]
+        public void can_add_extra_parameters()
+        {
+            var obj = new TestOptions
+            {
+                AString = "foo",
+            };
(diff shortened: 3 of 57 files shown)


src/Stripe.net/Infrastructure/ParameterBuilder.cs
```diff
@@ -33,6 +33,16 @@ public static string ApplyAllParameters(this StripeService service, object obj,
...
                     }
                 }
+
+                var field = obj.GetType().GetRuntimeField("ExtraParams");
+                if (field != null)
+                {
+                    foreach (KeyValuePair<string, string> pair in (Dictionary<string, string>) field.GetValue(obj))
+                    {
+                        var key = WebUtility.UrlEncode(pair.Key);
+                        RequestStringBuilder.ApplyParameterToRequestString(ref requestString, key, pair.Value);
```
remi-stripe: Does this mean that you could pass `extra_params[sub_params][subsub_value]` and it would work?

src/Stripe.net/Services/StripeParamsBase.cs
```diff
@@ -0,0 +1,13 @@
+using System.Collections.Generic;
+
+namespace Stripe
+{
+    public class StripeParamsBase
```
remi-stripe: I like the new name but should it be `StripeBaseOptions` to match the rest of the class naming?

src/Stripe.net/Infrastructure/ParameterBuilder.cs
```diff
@@ -33,6 +33,16 @@ public static string ApplyAllParameters(this StripeService service, object obj,
                             RequestStringBuilder.ProcessPlugins(ref requestString, attribute, property, value, obj);
                     }
                 }
+
+                var field = obj.GetType().GetRuntimeField("ExtraParams");
+                if (field != null)
```
remi-stripe: This was added temporarily right? Should we assert if you option doesn't have that value so that you know you forgot to inherit from `StripeParamsBase`?

ob-stripe: Unless that breaks the reflection logic beyond saving, I think a better solution would be to change the `ApplyAllParameters` signature to receive a `StripeParamsBase` (or whatever name we choose for the parent class) instead of an `object`.

But to answer your question, yes, all params classes should inherit from `StripeParamsBase` and so there's no reason why `ExtraParams` would not be available.

src/Stripe.net/Infrastructure/ParameterBuilder.cs
ob-stripe: The brackets would be URL-encoded. I think the API doesn't mind and would still interpret the value as a nested object, but not entirely sure tbh. Maybe URL-encoding the key is not the right thing to do here, and instead we'd make it the caller's responsibility to pass a properly URL-encoded key?

src/Stripe.net/Services/StripeParamsBase.cs
ob-stripe: Yeah, that's probably a better name.

src/Stripe.net/Infrastructure/ParameterBuilder.cs
remi-stripe: So what I meant is that if we remove the if, then we can catch cases where someone creates a new Options class but forgets to inherit from the Base class.

src/Stripe.net/Infrastructure/ParameterBuilder.cs
remi-stripe: hmmm that seems frustrating for users to have to do this (url encoding themselves). Maybe not supporting sub-hashes for now is fine then. Also we could do a plugin that says "oh this is a dictionary, let's properly encode those"

remi-stripe:
@ob-stripe flagging that I just pushed this:

Move all top-level Options classes to StripeBaseOptions

* Ensure that RequestOptions does not inherit from it as it's used in a different way in the library (extra headers)
* Same for INestedOptions as it's specific to sub-resources and should not have extra parameters.

src/Stripe.net/Infrastructure/ParameterBuilder.cs
remi-stripe: @ob-stripe Now that all class derive from `StripeBaseOptions` should we remove the check? Or assert if `field == null` as it should not happen?

ob-stripe: Yeah, an assert seems reasonable 👍

src/Stripe.net/Infrastructure/ParameterBuilder.cs
ob-stripe: I'd rather keep the feature simple for now. If someone actually needs to use this method with nested objects we can think about it then.

src/Stripe.net/Infrastructure/ParameterBuilder.cs
```diff
@@ -10,7 +10,7 @@ namespace Stripe.Infrastructure
 {
     internal static class ParameterBuilder
     {
-        public static string ApplyAllParameters(this StripeService service, object obj, string url, bool isListMethod = false)
+        public static string ApplyAllParameters(this StripeService service, StripeBaseOptions obj, string url, bool isListMethod = false)
```
ob-stripe: So I actually went with what I mentioned before: the method now expects a `StripeBaseOptions` instance instead of just an `object`. That way if we forget to derive from `StripeBaseOptions` the error will be immediately visible at compile time rather than at runtime. (I actually caught a few classes that you'd forgotten this way ;)

Plus, there's no need to muck with reflection to access `ExtraParams` now, so the code is a bit cleaner.

ob-stripe:
ptal @remi-stripe

src/Stripe.net/Services/StripeBaseOptions.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Stripe
+{
+    public class StripeBaseOptions
+    {
+        public void AddExtraParam(string key, string value) {
+            ExtraParams.Add(key, value);
+        }
+
+        public Dictionary<string, string> ExtraParams = new Dictionary<string, string>();
```
remi-stripe: @ob-stripe did we want to figure out how to make that one not public before merging?

PR #1058, opened 2017-11-17
add missing paid filter to invoices list

ahm322:
add missing paid filter to invoices list


ob-stripe:
Hi @ahm322, thanks for the PR! Even though this filter is not officially documented, I think it's peaceful to add it to the library.

@dpetrovics-stripe (dev-platform run) Mind pulling this in and releasing a version?

ob-stripe:
(CI fails because the PR comes from a forked repo and AppVeyor will not decrypt the API key env variable for those. We should simply use a regular, non-encrypted variable for the API key to solve this.)

PR #1060, opened 2017-11-23
Make Paid optional

ob-stripe:
r? @anelder-stripe 
cc @stripe/api-libraries 

Makes `Paid` optional, otherwise it defaults to `paid=false` in every invoice listing request.


anelder-stripe:
👍

PR #1064, opened 2017-11-29
Add support for listing sources on customers

### LINKED ISSUE #1062: CardService.List does not load data from source type="card"
I have customer stored with 3 sources with type = "card", when calling StripeCardService List method i get back 3 instances of the object of StripeCard, but properties of that object are not populated. 

All of the properties of StripeCard has values set to null, except for Id, CustomerId and Object those 3 have correct values.

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Fixes #1062.

This PR adds support for listing sources on customers, with optional filtering by source type.

It should be noted that while source objects are top-level resources, they are not directly listable (they can only be listed when attached to a specific customer). I have named the method `List` so as to be consistent with `StripeCardService` / `StripeBankAccountService`, and I think the presence of `customerId` in the method's signature makes it clear that this method only applies to customers.

### DIFF
--- a/src/Stripe.Tests.XUnit/sources/listing_sources_on_customer.cs
+++ b/src/Stripe.Tests.XUnit/sources/listing_sources_on_customer.cs
@@ -1,0 +1,101 @@
+using FluentAssertions;
+using Xunit;
+
+namespace Stripe.Tests.Xunit
+{
+    public class listing_sources_on_customer
+    {
+        public StripeList<StripeSource> SourceListAll { get; }
+        public StripeList<StripeSource> SourceListCard { get; }
+        public StripeList<StripeSource> SourceListBitcoin { get; }
+
+        public listing_sources_on_customer()
+        {
+            var sourceService = new StripeSourceService(Cache.ApiKey);
+            var customerService = new StripeCustomerService(Cache.ApiKey);
+
+            // Create customer
+            var CustomerCreateOptions = new StripeCustomerCreateOptions
+            {
+                Email = "source-list@example.com",
+            };
+            var Customer = customerService.Create(CustomerCreateOptions);
+
+            // Create card source and attach it to customer
+            var SourceCardCreateOptions = new StripeSourceCreateOptions
+            {
+                Type = StripeSourceType.Card,
+                Token = "tok_visa"
+            };
+            var SourceCard = sourceService.Create(SourceCardCreateOptions);
+
+            var SourceAttachOptions = new StripeSourceAttachOptions
+            {
+                Source = SourceCard.Id
+            };
+            SourceCard = sourceService.Attach(Customer.Id, SourceAttachOptions);
+
+            // Create bitcoin source and attach it to customer
+            var SourceBitcoinCreateOptions = new StripeSourceCreateOptions
+            {
+                Type = StripeSourceType.Bitcoin,
+                Amount = 1000,
+                Currency = "usd",
+                Owner = new StripeSourceOwner
+                {
+                    Email = "jenny.rosen+fill_now@example.com",
+                },
+            };
+            var SourceBitcoin = sourceService.Create(SourceBitcoinCreateOptions);
+
+            SourceAttachOptions.Source = SourceBitcoin.Id;
+            SourceBitcoin = sourceService.Attach(Customer.Id, SourceAttachOptions);
+
+            // List sources on customer
+            SourceListAll = sourceService.List(Customer.Id);
+
+            var SourceListOptions = new StripeSourceListOptions
+            {
+                Type = StripeSourceType.Card
+            };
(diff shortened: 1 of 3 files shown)


src/Stripe.net/Services/Sources/StripeSourceService.cs
```diff
@@ -40,6 +40,19 @@ public virtual StripeSource Detach(string customerId, string sourceId, StripeReq
             );
         }
 
+        public virtual StripeList<StripeSource> List(string customerId, StripeSourceListOptions listOptions = null, StripeRequestOptions requestOptions = null)
+        {
+            var url = string.Format(Urls.CustomerSources, customerId);
+
+            if (listOptions == null) {
+                listOptions = new StripeSourceListOptions();
+            }
```
remi-stripe: Why do you need to do this? Is it to ensure that `object` is passed as `source` even if nothing is sent? Asking because we don't do this in other List methods such as [customer](https://github.com/stripe/stripe-dotnet/blob/f157cede05b8e3ceb2dcac658fa5190078f117fc/src/Stripe.net/Services/Customers/StripeCustomerService.cs#L50-L56)

src/Stripe.net/Services/Sources/StripeSourceService.cs
```diff
@@ -40,6 +40,19 @@ public virtual StripeSource Detach(string customerId, string sourceId, StripeReq
             );
         }
 
+        public virtual StripeList<StripeSource> List(string customerId, StripeSourceListOptions listOptions = null, StripeRequestOptions requestOptions = null)
+        {
+            var url = string.Format(Urls.CustomerSources, customerId);
```
remi-stripe: Is it worth using a method to share the code? Not sure but that's what we do in other services (though the URL is used more than twice)

src/Stripe.Tests.XUnit/sources/listing_sources_on_customer.cs
```diff
@@ -0,0 +1,101 @@
...
+                Amount = 1000,
+                Currency = "usd",
+                Owner = new StripeSourceOwner
+                {
+                    Email = "jenny.rosen+fill_now@example.com",
+                },
+            };
+            var SourceBitcoin = sourceService.Create(SourceBitcoinCreateOptions);
+
+            SourceAttachOptions.Source = SourceBitcoin.Id;
```
remi-stripe: Minor but I like having separate options instances. Jayme used to compare the type on the object to the info on the options object itself but here since you override you can't do this.

It's definite more verbose my way but I found it cleaner when I started to work on the library. Feel free to ignore

src/Stripe.Tests.XUnit/sources/listing_sources_on_customer.cs
```diff
@@ -0,0 +1,101 @@
...
+        public void list_all_should_have_both_sources()
+        {
+            SourceListAll.Data.Count.Should().Be(2);
+
+            var SourceCard = SourceListAll.Data[0];
+            SourceCard.Type.Should().Be(StripeSourceType.Card);
+            SourceCard.Card.Brand.Should().Be("Visa");
+
+            var SourceBitcoin = SourceListAll.Data[1];
+            SourceBitcoin.Type.Should().Be(StripeSourceType.Bitcoin);
```
remi-stripe: Shouldn't bitcoin be first? Aren't sources returned with most recent first? Or is it due to the default source always being first even if added before and this being undocumented?

src/Stripe.net/Services/Sources/StripeSourceService.cs
ob-stripe: Yes, exactly. The `StripeSourceListOptions` object will ensure that `object=source` is added as a query parameter in the request. This seemed less hacky than hardcoding the parameter in the URL, but it does force us to instantiate an object if none is provided.

src/Stripe.Tests.XUnit/sources/listing_sources_on_customer.cs
ob-stripe: Oops, not sure why I missed this comment earlier.

Yeah, if you need the options object in your test it's better to instantiate it separately. In this case the options object are only instantiated within the test constructor though, and only the resulting list objects are persisted and used in test assertions.

src/Stripe.Tests.XUnit/sources/listing_sources_on_customer.cs
ob-stripe: Yes, the card source is attached first and becomes the default source.

It's definitely not great practice to rely on that behavior in a test, and I should probably not have hardcoded the indices, but it's probably not worth putting in too much effort in those tests since we plan to 🔥  them and use stripe-mock Soon™️.

PR #1071, opened 2017-12-14
Add support for account debits

remi-stripe:
* A Charge can be created with an Account id as the source.
* Deserialization of the Account object will work even though the API only returns an id and an object.
* Account debit as a transfer already works as `destination` is a string and can not be expanded.

r? @ob-stripe

### DIFF
--- a/src/Stripe.net.Tests/charges/when_creating_a_charge_with_a_stripe_account.cs
+++ b/src/Stripe.net.Tests/charges/when_creating_a_charge_with_a_stripe_account.cs
@@ -39,5 +39,8 @@ namespace Stripe.Tests
 
         It should_have_the_right_id = () =>
             _stripeCharge.Source.Id.ShouldEqual(_stripeAccount.Id);
+
+        It should_deserialize_the_account_properly = () =>
+            _stripeCharge.Source.Account.Id.ShouldEqual(_stripeAccount.Id);
     }
 }
--- a/src/Stripe.net.Tests/charges/when_creating_a_charge_with_a_stripe_account.cs
+++ b/src/Stripe.net.Tests/charges/when_creating_a_charge_with_a_stripe_account.cs
@@ -1,0 +1,43 @@
+using Machine.Specifications;
+
+namespace Stripe.Tests
+{
+    public class when_creating_a_charge_with_a_stripe_account
+    {
+        private static StripeAccount _stripeAccount;
+        private static StripeCharge _stripeCharge;
+
+        Establish context = () =>
+        {
+            // create a new custom account
+            _stripeAccount = new StripeAccountService().Create(new StripeAccountCreateOptions()
+            {
+                Country = "US",
+                Type = StripeAccountType.Custom
+            });
+        };
+
+        Because of = () =>
+        {
+            // charge the customer something
+            _stripeCharge = new StripeChargeService().Create(new StripeChargeCreateOptions()
+            {
+                Amount = 100,
+                Currency = "usd",
+                SourceTokenOrExistingSourceId = _stripeAccount.Id
+            });
+        };
+
+        It should_have_the_right_source = () =>
+            _stripeCharge.Source.Account.ShouldNotBeNull();
+
+        It should_not_have_the_wrong_source = () =>
+            _stripeCharge.Source.Card.ShouldBeNull();
+
+        It should_have_the_right_source_type = () =>
+            _stripeCharge.Source.Type.ShouldEqual(SourceType.Account);
+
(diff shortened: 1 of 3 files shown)


src/Stripe.net.Tests/charges/when_creating_a_charge_with_a_stripe_account.cs
```diff
@@ -0,0 +1,43 @@
+using Machine.Specifications;
```
remi-stripe: I went with the `Stripe.net.Tests` because that's where all the others are for that specific deserialization use-case.

src/Stripe.net/Entities/Source.cs
```diff
@@ -15,6 +16,7 @@ public class Source : StripeEntityWithId
     {
         public SourceType Type { get; set; }
 
+        public StripeAccount Account { get; set; }
```
remi-stripe: It seems easier to use `StripeAccount` here even though the API only returns this:
```
{
  id: "acct_XXXXX",
  object: "account"
}
```

src/Stripe.net.Tests/charges/when_creating_a_charge_with_a_stripe_account.cs
```diff
@@ -0,0 +1,43 @@
...
+        };
+
+        Because of = () =>
+        {
+            // charge the customer something
+            _stripeCharge = new StripeChargeService().Create(new StripeChargeCreateOptions()
+            {
+                Amount = 100,
+                Currency = "usd",
+                SourceTokenOrExistingSourceId = _stripeAccount.Id
```
remi-stripe: That parameter name is really bad. I wonder if we should go back to just `Source` once and for all to match the API. Likely out of scope for this change

src/Stripe.net/Infrastructure/JsonConverters/SourceConverter.cs
```diff
@@ -29,6 +29,12 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
                 Id = incoming.SelectToken("id").ToString()
             };
 
+            if (incoming.SelectToken("object")?.ToString() == "account")
+            {
+                source.Type = SourceType.Account;
+                source.Account = Mapper<StripeAccount>.MapFromJson(incoming.ToString());
+            }
+
```
remi-stripe: So, I wanted to do an `if/else if/ else if/ ... / else` because it's faster but seems there's only one in the entire lib so wasn't sure if there was a reason.

remi-stripe:
cc @dawn-stripe for awareness

src/Stripe.net.Tests/charges/when_creating_a_charge_with_a_stripe_account.cs
ob-stripe: Ack. Not ideal, but understandable :)

src/Stripe.net.Tests/charges/when_creating_a_charge_with_a_stripe_account.cs
ob-stripe: ==

src/Stripe.net/Entities/Source.cs
ob-stripe: 👍

src/Stripe.net/Infrastructure/JsonConverters/SourceConverter.cs
ob-stripe: I don't think there's any specific reason -- these cases are all mutually exclusive so using `else if` would be fine.

src/Stripe.net.Tests/charges/when_creating_a_charge_with_a_stripe_account.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        It should_not_have_the_wrong_source = () =>
+            _stripeCharge.Source.Card.ShouldBeNull();
+
+        It should_have_the_right_source_type = () =>
+            _stripeCharge.Source.Type.ShouldEqual(SourceType.Account);
+
+        It should_have_the_right_id = () =>
+            _stripeCharge.Source.Id.ShouldEqual(_stripeAccount.Id);
+    }
```
ob-stripe: Do you mind adding a test for `_stripeCharge.Source.Account.Id`, just to make sure the partial account object gets deserialized properly?

ob-stripe:
lgtm!

remi-stripe:
Deferring to you to merge/release

PR #1074, opened 2017-12-18
Add support for expanding application on charge objects

### LINKED ISSUE #1073: Application name for a charge
How can I see the application name that is associated with a charge. I have a web store connected to Stripe, and on my charges I have the Application (Id) as "ca_XXXXX". Is there a call to determine what the name of the connected application is for ca_XXXXX?

Thanks!

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add support for expanding `application` attribute on charge objects

Fixes #1073.

### DIFF
--- a/src/Stripe.Tests.XUnit/connect/when_charging_from_managed_account.cs
+++ b/src/Stripe.Tests.XUnit/connect/when_charging_from_managed_account.cs
@@ -4,48 +4,27 @@ using Xunit;
 
 namespace Stripe.Tests.Xunit
 {
-    public class charging_from_another_account
+    public class charging_from_another_account : IClassFixture<connect_fixture>
     {
-        private StripeApplicationFee _appFee;
+        private readonly connect_fixture fixture;
 
-        public charging_from_another_account()
+        public charging_from_another_account(connect_fixture fixture)
         {
-            var anotherAccount = new StripeAccountService(Cache.ApiKey).Create(new StripeAccountCreateOptions
-                {
-                    DefaultCurrency = "usd",
-                    Email = "cu_xxxxxx@gmail.com",
-                    Type = StripeAccountType.Custom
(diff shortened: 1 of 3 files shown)


src/Stripe.Tests.XUnit/connect/when_charging_from_managed_account.cs
```diff
@@ -38,7 +40,14 @@ public charging_from_another_account()
...
 
-            _appFee = appFeeService.Get(charge.ApplicationFeeId);
+            _appFee = appFeeService.Get(_charge.ApplicationFeeId);
+        }
+
+        [Fact]
+        public void it_should_have_connected_app_on_charge()
+        {
+            _charge.ApplicationId.Should().StartWith("ca_");
+            _charge.Application.Should().NotBeNull();
```
fred-stripe: Is test order execution guaranteed in XUnit? I'd expect to see the charge initialization in a setup method or a new charge explicitly created for this test.

remi-stripe: agreed, usually this is done in a fixture first no?

ob-stripe: I'm not entirely sure what you two meant, or if you're even talking about the same thing, but:
- no, XUnit does not guarantee test order by default
- using a fixture allows for context sharing between tests (cf. https://xunit.github.io/docs/shared-context.html). I updated the test to use a fixture to avoid creating multiple resources needlessly and make the tests a bit faster

fred-stripe:
Ah, makes sense @ob-stripe, I was thrown because the class and constructor name were `snake_case` and was wondering how `connect_fixture()` would get called.

lgtm

PR #1077, opened 2017-12-23
Ensures that parameters are added to the URL only once

### LINKED ISSUE #1076: Creating subscriptions sends duplicate parameters
Using:
* .Net Core 2
* Windows 10 Pro 
* Visual Studio 2017

When creating a subscription, the parameters are sent twice in the request body. Here's an example HTTP Request:
```
POST https://api.stripe.com/v1/subscriptions HTTP/1.1

coupon=discount-10p&
plan=stripe-daily-4&
source=tok_1Bc5AUCkr3AsGInl9NHtHxpD&
tax_percent=0&
customer=cus_BynumueEYyVFKv&
coupon=discount-10p&
plan=stripe-daily-4&
source=tok_1Bc5AUCkr3AsGInl9NHtHxpD&
tax_percent=0
```
Notice that every parameter (except for customer) is duplicated in the request body (line breaks added for clarity).

I believe the problem is because `this. ApplyAllParameter` is being called twice from here:  https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/Subscriptions/StripeSubscriptionService.cs#L38-L49

After the first time `this. ApplyAllParameter` is called, the `requestString` that's returned from `this. ApplyAllParameter` looks like this (see the watch window in the screenshot):
image

The second time `this. ApplyAllParameter` is called, the `requestString` that's returned from `this. ApplyAllParameter` looks like this:
image

Notice that all of the parameters are doubled the second time.

remi-stripe:
Fixes #1076.

r? @ob-stripe

### DIFF
--- a/src/Stripe.net/Services/Subscriptions/StripeSubscriptionService.cs
+++ b/src/Stripe.net/Services/Subscriptions/StripeSubscriptionService.cs
@@ -42,7 +42,7 @@ namespace Stripe
 
             return Mapper<StripeSubscription>.MapFromJson(
                 Requestor.PostString(
-                    this.ApplyAllParameters(createOptions, url),
+                    url,
                     SetupRequestOptions(requestOptions)
                 )
             );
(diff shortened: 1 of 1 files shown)


ob-stripe:
lgtm

PR #1083, opened 2018-01-03
Add specialized ListOptions for each resource

### LINKED ISSUE #974: StripeListOptions
Explicitly declare a specialized ListOptions for each individual service or remove the object all together.  Maybe just use the generic ListOptions everywhere and add a bunch of optional items (I think this is less nice)?

Explicitly support TotalCount everywhere or remove it all together.  (I'm personally in favor of removing it, because it isn't possible to retrieve that and have it make sense when you have paginated results.)

EDIT: TotalCount is already fixed in a PR @fred-stripe made-- https://github.com/stripe/stripe-dotnet/pull/967

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries @anelder-stripe 

Fixes #974.

This PR adds specialized ListOptions classes for every resource.

### DIFF
--- a/src/Stripe.Tests.XUnit/transfer_reversals/_fixture.cs
+++ b/src/Stripe.Tests.XUnit/transfer_reversals/_fixture.cs
@@ -7,7 +7,7 @@ namespace Stripe.Tests.Xunit
     {
         public StripeTransferReversalCreateOptions TransferReversalCreateOptions { get; }
         public StripeTransferReversalUpdateOptions TransferReversalUpdateOptions { get; }
-        public StripeListOptions TransferReversalListOptions { get; }
+        public StripeTransferReversalListOptions TransferReversalListOptions { get; }
 
         public StripeTransfer Transfer { get; }
         public StripeTransferReversal TransferReversal { get; }
@@ -47,6 +47,8 @@ namespace Stripe.Tests.Xunit
                 }
             };
 
+            TransferReversalListOptions = new StripeTransferReversalListOptions();
+
             var service = new StripeTransferReversalService(Cache.ApiKey);
             service.ExpandBalanceTransaction = true;
             service.ExpandTransfer = true;
@@ -54,7 +56,7 @@ namespace Stripe.Tests.Xunit
             TransferReversal = service.Create(Transfer.Id, TransferReversalCreateOptions);
             TransferReversalUpdated = service.Update(Transfer.Id, TransferReversal.Id, TransferReversalUpdateOptions);
             TransferReversalRetrieved = service.Get(Transfer.Id, TransferReversal.Id);
-            TransferReversalList = service.List(Transfer.Id, new StripeListOptions());
+            TransferReversalList = service.List(Transfer.Id, TransferReversalListOptions);
 
             // get the original transfer
             Transfer = new StripeTransferService(Cache.ApiKey).Get(Transfer.Id);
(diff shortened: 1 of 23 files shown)


src/Stripe.Tests.XUnit/transfer_reversals/_fixture.cs
```diff
@@ -54,7 +54,7 @@ public transfer_reversals_fixture()
             TransferReversal = service.Create(Transfer.Id, TransferReversalCreateOptions);
             TransferReversalUpdated = service.Update(Transfer.Id, TransferReversal.Id, TransferReversalUpdateOptions);
             TransferReversalRetrieved = service.Get(Transfer.Id, TransferReversal.Id);
-            TransferReversalList = service.List(Transfer.Id, new StripeListOptions());
+            TransferReversalList = service.List(Transfer.Id, TransferReversalListOptions);
```
remi-stripe: why don't you need a new here?

src/Stripe.net.Tests/countryspecs/when_listing_country_specs.cs
```diff
@@ -17,7 +17,7 @@ public class when_listing_country_specs
         Because of = () =>
         {
             _countrySpecList = _countrySpecService.List(
-                new StripeListOptions() { Limit = 8 }
+                new CountrySpecListOptions() { Limit = 8 }
```
remi-stripe: why not `StripeCountrySpecListOptions`?

src/Stripe.net/Services/BankAccounts/BankAccountListOptions.cs
```diff
@@ -0,0 +1,8 @@
+using Newtonsoft.Json;
+
+namespace Stripe
+{
+    public class BankAccountListOptions : StripeListOptions
```
remi-stripe: Why not `StripeBankAccountListOptions`? I guess it's because of the class name discrepancies but it makes it confusing

src/Stripe.Tests.XUnit/transfer_reversals/_fixture.cs
ob-stripe: Oops! I didn't see that `TransferReversalListOptions` was declared but never initialized. This is now fixed.

src/Stripe.net.Tests/countryspecs/when_listing_country_specs.cs
ob-stripe: To maintain consistency with all the other classes (service and entities). I agree that we should fix this, but it's a job for a different PR and in the meantime I'd rather maintain consistency within each resource.

src/Stripe.net/Services/BankAccounts/BankAccountListOptions.cs
ob-stripe: Same answer as for `CountrySpec`: to maintain internal consistency with the other BankAccount classes (entity, service and other options classes). Also something that we should fix separately.

remi-stripe: I see but it seems better (to me) to keep consistency with the new naming of the `ListOptions` than the parent class itself. Especially as those are the only 2 that are different

ob-stripe: I disagree, but since we're going to release a new major version anyway, I guess now's as good a time as any to fix all those classes :)

remi-stripe: sounds good :)

Issue #1088, opened 2018-01-08
StripeSourceCreateOptions does not have a statement_descriptor property


AFallad:
Hello,

I'm trying to create a source and add data to the top level statement_descriptor property (https://stripe.com/docs/api#create_source-statement_descriptor) however the class StripeSourceCreateOptions.cs does not contains a property that maps to it, I can see that bancontact[statement_descriptor], ideal[statement_descriptor] and sofort[statement_descriptor] are mapped, but not the top level statement_descriptor.

Thanks in advance for your help!

remi-stripe:
@AFallad Thanks a lot for the report! I'll work on a PR and hopefully push a fix in the next few days!

PR #1089, opened 2018-01-08
Add support for top-level statement descriptor for all sources

remi-stripe:
This removes the type-specific statement descriptor as it is now supported as a top level parameter and moves every type to `StatementDescriptor`.

Fixes https://github.com/stripe/stripe-dotnet/issues/1088

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.Tests.XUnit/sources/creating_and_updating_bancontact_source.cs
+++ b/src/Stripe.Tests.XUnit/sources/creating_and_updating_bancontact_source.cs
@@ -6,27 +6,29 @@ namespace Stripe.Tests.Xunit
     public class creating_and_updating_bancontact_source
     {
         private StripeSource Source { get; set; }
+        private StripeSourceCreateOptions SourceCreateOptions { get; set; }
 
         public creating_and_updating_bancontact_source()
         {
-            var options = new StripeSourceCreateOptions
+            SourceCreateOptions = new StripeSourceCreateOptions
             {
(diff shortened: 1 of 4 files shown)


AFallad:
Hello, I did some testing and I don't think that the type-specific statement_descriptor should be deleted, since the top level and type-specific can be different, check this request
image
image

Regards!

remi-stripe:
@AFallad even if both can be passed to be backwards compatible, the `ideal[statement_descriptor]` would be ignored and we'd default to `statement_descriptor` instead. That's why I think it's cleaner to remove all the other ones and only officially support the top-level one.

AFallad:
More than backward compatibility I was worried that with that change we will be unable to set the type specific `ideal[statement_descriptor]`  which is set separately from the top level 'statement_descriptor' by the Stripe API. 
I don't know what case use would require the dev to set a different descriptor for both, but by removing the specific ones, that use case would not be supported by the stripe-net library but would be supported by the Stripe API
I appreciate a lot the swiftness. I'll keep an eye out for the release that fixes this. Thanks!

remi-stripe:
@AFallad I hear you but ultimately if you were to pass both, one would be ignored silently. The top-level is the _right_ one to set and you should ignore that the type-specific ones even exist. It's for an old compatibility reason and `statement_descriptor` is always the preferred/correct one.

stan-stripe:
LGTM

PR #1091, opened 2018-01-10
Fix synchronous login link creation

remi-stripe:
* The method for synchronous Creation was not working at all and is fixed
* Remove old test approach from the old framework
* Moved new test to XUnit
* Comment out the test as it would break everyone's local test suite.

r? @ob-stripe 
cc @stripe/api-libraries


remi-stripe:
@ob-stripe do you think you can do a release of this one since it's a real bug?

Issue #1090, opened 2018-01-10
Card is null in customer source


BendicoE:
I'm having trouble with accessing the card information in a customer source. (I'm using the new sources workflow in Stripe.net.) I create a source with javascript in the frontend and on the server I use StripeSourceService.Get() to access it. The .Card member of the StripeSource object now seems to contain the Card information, with Fingerprint, etc. I then attach this source to a StripeCustomer.

If I now loop through the StripeCustomer.Sources, I can see that the source with that Id was added. However, the .Card member now seems to be null in the source object. I also noticed that the list of sources on the customer object is of a different type: Source instead of StripeSource.

I wonder, is this inherent behaviour from Stripe or could it be something missing in Stripe.net?

```c#
StripeSourceService sourceService = new StripeSourceService("sk_myPrivateApiKey");
StripeCustomerService customerService = new StripeCustomerService("sk_myPrivateApiKey");
StripeSource source = sourceService.Get("src_theSourceId");
// source.Card is an object with card information
customerService.Update("cus_theCustomerId", new StripeCustomerUpdateOptions() {  DefaultSource = "src_theSourceId" });
StripeCustomer customer = customerService.Get("cus_theCustomerId");
Source customerSource = customer.Sources.Where(x => x.Id == "src_theSourceId").First();
// customerSource.Card is null
```

(If this is expected behaviour, I apologize for posting an issue here. I have been working with Stripe and Stripe.net for a while, but I am by no means a Stripe pro.)

Regards,
Bendik Engebretsen

ob-stripe:
Hi @BendicoE, thanks for reporting this.

There are two different classes with very similar names but different behaviors:
- `StripeSource` represents a [source object](https://stripe.com/docs/api#source_object)
- `Source` is a wrapper class that is used to represent any customer's [payment source](https://stripe.com/docs/api#customer_object-sources), which may be a card (`StripeCard`), a bank account (`StripeBankAccount`) or... a source (`StripeSource`)

So you'd need to modify your code like this:
```c#
Source customerSource = customer.Sources.Where(x => x.Id == "src_theSourceId").First();
StripeSource customerCardSource = source.Source;
// customerCardSource.Card will have the values you expect
```

However, that won't work either because right now, `Source` does not support wrapping `StripeSource`. This is an issue with the library, we will push a fix and release a new version shortly.

BendicoE:
Thanks, Oliver!
[...]

PR #1096, opened 2018-01-19
Add support for source objects in Source wrapper

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Tentative fix for #1090.

### DIFF
--- a/src/Stripe.net/Entities/Source.cs
+++ b/src/Stripe.net/Entities/Source.cs
@@ -6,9 +6,10 @@ namespace Stripe
     public enum SourceType
     {
         Account,
-        Card,
         BankAccount,
-        Deleted
+        Card,
+        Deleted,
+        Source,
     }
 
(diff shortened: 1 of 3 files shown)


src/Stripe.net/Entities/Source.cs
```diff
@@ -17,8 +18,9 @@ public class Source : StripeEntityWithId
         public SourceType Type { get; set; }
 
         public StripeAccount Account { get; set; }
-        public StripeDeleted Deleted { get; set; }
-        public StripeCard Card { get; set; }
         public StripeBankAccount BankAccount { get; set; }
+        public StripeCard Card { get; set; }
+        public StripeDeleted Deleted { get; set; }
+        public StripeSource SourceObject { get; set; }
```
ob-stripe: It's not possible to have a property with the same name as the wrapping class, so I named the property `SourceObject` instead of just `Source`.

PR #1068, opened 2017-12-01
[WIP] updates for upcoming Stripe API version bump

jhoon-stripe:
#### Summary

Similar to https://github.com/stripe/stripe-go/pull/496, this PR adds support for an upcoming Stripe API version, tentatively dated 2017-12-01.

 #### Testing

I'll plan to...
 - [ ] update existing test cases for plans and products
 - [ ] add coverage for plan creation via product ID and via inline product fields

cc @stripe/api-libraries

PR #1099, opened 2018-02-08
Pantera plan and product updates

pantera-stripe:
Summary
Replacement for @jhoon-stripe's PR: https://github.com/stripe/stripe-dotnet/pull/1068
This PR adds support for an upcoming Stripe API version, released on `2018-02-05`. Extends the original PR by adding `nickname` to `StripePlan` and `type` to `StripeProduct`. Also adds `BillingCycleAnchor` to `StripeSubscription` which was also released on 2018-02-05

r? @ob-stripe @brandur-stripe (brandur reviewed original PR)
cc @stripe/api-libraries


remi-stripe:
@pantera-stripe wanted to flag https://github.com/stripe/stripe-dotnet/issues/1101 to make sure it could be done in the same PR. Let me know if you need help!

pantera-stripe:
@remi-stripe You're right, we're missing that variable. I was about to cargo-cult the code from the create options, but I just realized... it's buggy. (If I'm reading this correctly)

```
        [JsonProperty("billing_cycle_anchor")]
        internal string BillingCycleAnchorInternal
        {
            get
            {
                if (BillingCycleAnchorNow)
                    return "now";

                if (`)
                    return "unchanged";

                return BillingCycleAnchor?.ConvertDateTimeToEpoch().ToString();
            }
        }
```

If `BillingCycleAnchorNow` is `true` it'll send `now`. If you set both `BillingCycleAnchorNow` and `BillingCycleAnchorUnchanged` to `true` it'll also send `now`! Yay!

I'm fine replicating this not-so-amazing behavior in this PR. But I'd like to push a larger fix off to another PR. I really want to get this reviewed and merged.

remi-stripe:
@pantera-stripe I'm fine with reproducing the current behaviour. It's not perfect but the alternative would be to either send both and rely on the API error or do an assert or similar to leave early. FWIW we have some logic like this in stripe-go where we make a decision on which one to send first and I think it's okay.

src/Stripe.Tests.XUnit/plans/_fixture.cs
```diff
@@ -40,4 +45,43 @@ public plans_fixture()
...
+                Type = "service"
+            };
+
+            var productService = new StripeProductService(Cache.ApiKey);
+            Product = productService.Create(ProductCreateOptions);
+
+            PlanCreateOptions = new StripePlanCreateOptions() {
+                // Add a space at the end to ensure the ID is properly URL encoded
+                // when passed in the URL for other methods
+                Id = "test-plan-" + Guid.NewGuid().ToString() + " ",
```
remi-stripe: We likely can skip enforcing the Id here since it's now optional

src/Stripe.Tests.XUnit/plans/when_creating_and_updating_plans.cs
```diff
@@ -11,14 +11,13 @@ public class creating_and_updating_plans : IClassFixture<plans_fixture>
 
         public creating_and_updating_plans(plans_fixture fixture)
         {
-            this.fixture = fixture;
+            this.fixture = fixture;
```
remi-stripe: remove extra space

src/Stripe.Tests.XUnit/plans/when_creating_and_updating_plans.cs
```diff
@@ -11,14 +11,13 @@ public class creating_and_updating_plans : IClassFixture<plans_fixture>
...
         {
-            this.fixture = fixture;
+            this.fixture = fixture; 
         }
 
         [Fact]
         public void created_has_the_right_details()
         {
             fixture.Plan.Id.Should().Be(fixture.PlanCreateOptions.Id);
-            fixture.Plan.Name.Should().Be(fixture.PlanCreateOptions.Name);
```
remi-stripe: Shouldn't we test the `Nickname` property here?

src/Stripe.Tests.XUnit/plans/when_creating_and_updating_plans.cs
```diff
@@ -28,7 +27,6 @@ public void created_has_the_right_details()
         public void updated_has_the_right_details()
         {
             fixture.PlanUpdated.Id.Should().Be(fixture.Plan.Id);
-            fixture.PlanUpdated.Name.Should().BeEquivalentTo(fixture.PlanUpdateOptions.Name);
```
remi-stripe: same as above, we should look at `Nickname`

src/Stripe.net/Entities/StripePlan.cs
```diff
@@ -35,10 +35,16 @@ public class StripePlan : StripeEntityWithId
         [JsonProperty("name")]
         public string Name { get; set; }
```
remi-stripe: Shouldn't we remove `Name` completely?

src/Stripe.net/Entities/StripeProduct.cs
```diff
@@ -81,14 +81,26 @@ public class StripeProduct : StripeEntityWithId
...
+        public bool? Shippable { get; set; }
 
         /// <summary>
         /// A sublist of active SKUs associated with this product.
         /// </summary>
         [JsonProperty("skus")]
         public StripeList<StripeSku> Skus { get; set; }
 
+        /// <summary>
+        /// What appears on a customers credit card statement. Max of 22 characters.
```
remi-stripe: customer's
I'd also remove the 22 chars mention, better to keep this in the official docs.

src/Stripe.net/Entities/StripeProduct.cs
```diff
@@ -81,14 +81,26 @@ public class StripeProduct : StripeEntityWithId
...
+        /// </summary>
+        [JsonProperty("statement_descriptor")]
+        public string StatementDescriptor { get; set; }
+
+        /// <summary>
+        /// The type of the Product. Either 'good' or 'service'
```
remi-stripe: missing trailing .

src/Stripe.net/Services/Plans/StripePlanCreateOptions.cs
```diff
@@ -16,20 +16,23 @@ public class StripePlanCreateOptions : StripeBaseOptions
...
         public int? TrialPeriodDays { get; set; }
 
-        [JsonProperty("statement_descriptor")]
-        public string StatementDescriptor { get; set; }
-
         [JsonProperty("metadata")]
         public Dictionary<string, string> Metadata { get; set; }
+
+        [JsonProperty("nickname")]
+        public string Nickname { get; set; }
```
remi-stripe: should be in alphabetical order if we can

src/Stripe.net/Services/Products/StripeProducNestedCreateOptions.cs
```diff
@@ -0,0 +1,13 @@
+using Newtonsoft.Json;
+
+namespace Stripe
+{
+    public class StripeProductNestedCreateOptions : INestedOptions
```
remi-stripe: Since it's nested under the plan options I think we might want to name it `StripePlanProductCreatOptions` to indicate the relation.

Not a huge deal though as naming is at best inconsistent in the library today already.

src/Stripe.net/Services/Products/StripeProducNestedCreateOptions.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        [JsonProperty("product[name]")]
+        public string Name { get; set; }
+
+        [JsonProperty("product[statement_descriptor]")]
+        public string StatementDescriptor { get; set; }
+    }
```
remi-stripe: We should support all options here and not just name and statement_descriptor. The API [docs](https://stripe.com/docs/api#create_plan-product) indicate we support `id` and `metadata` too.

src/Stripe.net/Entities/StripePlan.cs
pantera-stripe: Wouldn't that prevent existing API integrations from upgrading their bindings?

remi-stripe: I don't entirely follow. Anyone who gets that new version of the code will already have to change their integration since it's a new API version and we've added/moved/removed multiple properties.

Name does not exist on the new API version, only Nickname does so `Name` should go.

This will all be behind a major version

pantera-stripe: Right, I'm talking the other way. Do merchants ever update their bindings independent of their API version? If not, then I can remove it. But if they do, then updating bindings to this version would eliminate their ability to access `Name`

remi-stripe: @pantera-stripe stripe-dotnet enforces an API version for the entire library. There is no way to "choose" an API version. Every user of stripe-dotnet is locked to the API version that the version of the library they use has been built for.

In another part of your PR you are now enforcing `2018-02-06` so every developer who upgrades their integration to that new version of stripe-dotnet will use that version on every API request. So `Name` will never exist or be set.

pantera-stripe: Yeah. Totally overlooked that. Will remove.

src/Stripe.net/Services/Plans/StripePlanCreateOptions.cs
pantera-stripe: Ah. Right. Alphabetical, but metadata last.

src/Stripe.net/Services/Products/StripeProducNestedCreateOptions.cs
pantera-stripe: My thought was there is nothing specific about the nesting to plans -- though you're right, that's the only place it's used and probably the only place it'll ever be used.

src/Stripe.net/Services/Products/StripeProducNestedCreateOptions.cs
pantera-stripe: Good catch.

pantera-stripe:
Applied the feedback.

r? @remi-stripe

remi-stripe:
Looks good to me. I don't like giving the only +1 on PRs though.

r? @ob-stripe

src/Stripe.net/Entities/StripePlan.cs
```diff
@@ -32,13 +32,16 @@ public class StripePlan : StripeEntityWithId
...
-        [JsonProperty("name")]
-        public string Name { get; set; }
+        [JsonProperty("nickname")]
+        public string Nickname { get; set; }
+
+        [JsonProperty("product")]
+        public string ProductId { get; set; }
 
         [JsonProperty("statement_descriptor")]
         public string StatementDescriptor { get; set; }
```
ob-stripe: Shouldn't we remove this property?

pantera-stripe: Yup. Will remove.

ob-stripe:
@pantera-stripe LGTM. Mind squashing the commits before I merge?

pantera-stripe:
@ob-stripe squashed

PR #1105, opened 2018-02-16
#1104: Added real parameterless constructors to all services

### LINKED ISSUE #1104: Suggestion: Services should have a real parameterless constructor
Having a real parameterless constructor allows classes to be used as generic types in methods with a `where T : new()` clause. A good example in the context of this library would be a service builder method that ensures only one instance of any given service is created:

```csharp
        private TService GetStripeService<TService>() where TService : StripeService, new()
        {
            var key = typeof(TService);

            if (!_services.TryGetValue(key, out var service))
            {
                lock(_services)
                {
                    if (!_services.TryGetValue(key, out service))
                    {
                        _services[key] = service = new TService();
                    }
                }
            }

            return service as TService;
        }
```

Since all Stripe .NET services have an optional `apiKey` parameter, replaced by `null` at compile-time, the above construction cannot be made. We have to remove the `new()` constraint and either use builders or reflection to create instances, which is less "elegant".

slimCODE:
This allows these services to serve as a generic type in methods with a `where TService : new()` clause.

Fixes #1104.

### DIFF
--- a/src/Stripe.net/Services/3DSecure/Stripe3DSecureService.cs
+++ b/src/Stripe.net/Services/3DSecure/Stripe3DSecureService.cs
@@ -8,7 +8,10 @@ namespace Stripe
     [Obsolete("Use the StripeSourceService instead.")]
     public class Stripe3DSecureService : StripeBasicService<Stripe3DSecure>
     {
-        public Stripe3DSecureService(string apiKey = null) : base(apiKey) { }
+        public Stripe3DSecureService() : base(null) { }
+        public Stripe3DSecureService(string apiKey) : base(apiKey) { }
+
+
 
         // Sync
         public virtual Stripe3DSecure Create(Stripe3DSecureCreateOptions createOptions, StripeRequestOptions requestOptions = null)
--- a/src/Stripe.net/Services/Account/StripeAccountService.cs
+++ b/src/Stripe.net/Services/Account/StripeAccountService.cs
@@ -7,7 +7,9 @@ namespace Stripe
 {
     public class StripeAccountService : StripeService
     {
-        public StripeAccountService(string apiKey = null) : base(apiKey) { }
+        public StripeAccountService() : base(null) { }
+        public StripeAccountService(string apiKey) : base(apiKey) { }
+
 
 
         //Sync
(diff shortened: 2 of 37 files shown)


brandur-stripe:
Thanks @slimCODE. This seems like it'll be useful and doesn't appear to be harmful in any way.

The one thing I noticed is that it seems like you've been relatively selective about which services get this treatment. Is there any reason for that? Your patch shows 20 files changed, but there are more like 36 services that are not one of the base classes:

```
$ ls -1 -R  src/Stripe.net/Services/ | grep 'Service.cs'
StripeBasicService.cs
StripeService.cs
Stripe3DSecureService.cs
StripeAccountService.cs
StripeApplePayDomainService.cs
StripeApplicationFeeRefundService.cs
StripeApplicationFeeService.cs
StripeBalanceService.cs
BankAccountService.cs
StripeCardService.cs
StripeChargeService.cs
CountrySpecService.cs
StripeCouponService.cs
StripeCustomerService.cs
StripeDiscountService.cs
StripeDisputeService.cs
StripeEphemeralKeyService.cs
StripeEventService.cs
StripeExchangeRateService.cs
StripeExternalAccountService.cs
StripeFileUploadService.cs
StripeInvoiceItemService.cs
StripeInvoiceService.cs
StripeLoginLinkService.cs
StripeOAuthTokenService.cs
StripeOrderService.cs
StripePayoutService.cs
StripePlanService.cs
StripeProductService.cs
StripeRefundService.cs
StripeSkuService.cs
StripeSourceTransactionService.cs
StripeSourceService.cs
StripeSubscriptionItemService.cs
StripeSubscriptionService.cs
StripeTokenService.cs
StripeTransferReversalService.cs
StripeTransferService.cs
```

(Also, you might want to add a "Fixes #1104" in your description above just so the issue gets linked somewhere.)

slimCODE:
@brandur-stripe That was a louzy patch, I missed the ones deriving from `StripeBasicService<>`!

brandur-stripe:
@slimCODE Ah I see what happened. Thanks for the quick fix!

Would you mind rebasing to get your two commits squashed down into one? That'll also give us a chance to try the CI build again, which seems to have hit an intermittent network problem and failed.

slimCODE:
@ob-stripe I added those lines to normalize, as there were three lines everywhere else between the ctor (and props) and the "Sync" methods. If you prefer, I'll remove them.

ob-stripe:
@slimCODE Ah no, that's fine then. Thanks for going the extra mile :)

ob-stripe:
Released in 13.1.0. Thanks again for the contribution!

PR #1107, opened 2018-02-19
Minor webhook signing improvements

fred-stripe:
Two improvements here:

- (minor) `ConstructEvent` now throws a `StripeException` instead of a generic `Exception`, which prevents calling code from catching non-SDK related errors when constructing events.
- (minor) Refactor the test suite to eliminate the `EpochUtcNowOverride` class member, which was only used by the test suite.

I had to add a couple more constructors to `StripeException`, as the existing constructor was only for errors during HTTP requests. As a bigger project I'd like to see us leave `StripeException` as a generic exception and derive an `StripeHTTPException` class from `StripeException`, which would cleanly break up the class responsibilities. It'd also allow us to add custom exception formatting and better surface non-200 HTTP responses and API error messages.

r? @ob-stripe 
cc @stripe/api-libraries @rpaul-stripe


ob-stripe:
@fred-stripe Great changes -- +1 to removing the special `EpochUtcNowOverride` attribute and adding a signature to inject the time instead, much cleaner!

ob-stripe:
Released in 13.0.2.

Issue #1116, opened 2018-02-27
Suggestion: Entities with a Metadata property should implement a common interface


slimCODE:
Many (most) entities have a `Dictionary<string, string> Metadata { get; set; }` property. Writing extension methods to help push and pull strongly-typed information from these fields is very useful. I know I've created myself a set of `GetIntMetadata()`, `GetLocalizedMetadata()` or `GetDateMetadata()` extensions.

Unfortunately, because all these entities don't share a base common type declaring that property, those extension methods must be copied for each entity type.

Having a common `ISupportMetadata` interface would be a simple and non-breaking addition.
(Note: I already did it)

ob-stripe:
This sounds like a great improvement. We should absolutely do this!

ob-stripe:
Implemented in 13.3.0. Thanks for the suggestion and the patch @slimCODE! :)

semenenkov:
I've upgraded from 19.6 which had ISupportMetadata to 27.3 and it has no ISupportMetadata. First surprise is that I haven't found any description on this neither in change log nor in major migration instructions for v 20, 22 and 27.
Then I've found something similar - IHasMetadata but it is missing for some objects that in fact have Dictionary<string, string>Metadata property, for example for CustomerCreateOptions class. So my suggestion is to reopen this ticket or raise another one to fix this.

ob-stripe:
Hi @semenenkov. `IHasMetadata` is applied to all entity classes (i.e. model classes representing resources returned by Stripe's API, e.g. `Customer`) but not to options classes (i.e. classes representing request parameters sent to Stripe's API, e.g. `CustomerCreateOptions`).

Could you explain why you think it would be useful to have the interface on options classes?

semenenkov:
@ob-stripe thanks for quick reply.
CustomerCreateOptions represents create data for Customer object. We can consider it as entity, just unsaved yet. And if we look at Stripe docs - 
https://stripe.com/docs/api/metadata
their .net example is for ChargeCreateOptions, not for Charge. 
This lib is a wrapper around api. Some api calls allow to save metadata. All corresponding objects should have Metadata property (and I think most of them have). In fact that's one feature supported by many calls and it would make sence to keep it under one common interface.

semenenkov:
.. and as I said at some point of time it became a breaking change, and now we'll have to add some ugly relfection-based code to process metadata of such objects which don't inherit this interface

ob-stripe:
[...]

PR #1117, opened 2018-02-27
Added ISupportMetadata interface

slimCODE:
Added that contract to all entities with a `Metadata` property.

This implements suggestion #1116.


ob-stripe:
Awesome! Thanks for the quick patch. LGTM!

PR #1119, opened 2018-02-28
Adds the `FailureReason` property to `StripeRedirect`

### LINKED ISSUE #1118: StripeRedirect does not have a property for failure_reason
Hello again,

I noticed the failure_reason does not have a mapping to the StripeRedirect model, hopefully it can be added.
https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Entities/Sources/StripeRedirect.cs

image

Thanks in advance for the help!

brandur:
We're currently missing the `FailureReason` property which is documented
here [1]. This patch adds it.

Fixes #1118.

[1] https://stripe.com/docs/api#source_object-redirect-failure_reason

r? @ob-stripe

### DIFF
--- a/src/Stripe.net/Entities/Sources/StripeRedirect.cs
+++ b/src/Stripe.net/Entities/Sources/StripeRedirect.cs
@@ -5,6 +5,12 @@ namespace Stripe
     public class StripeRedirect : StripeEntity
     {
         /// <summary>
+        /// The failure reason for the redirect. One of <see cref="StripeRedirectFailureReason" />.
+        /// </summary>
+        [JsonProperty("failure_reason")]
+        public StripeRedirectFailureReason? FailureReason { get; set; }
+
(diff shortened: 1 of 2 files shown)


ob-stripe:
@brandur-stripe @AFallad asked in #1121 if we could declare constants with the possible values for this property. I think we can do one better and create an enum type -- we currently have only one in the entire library (`StripeBilling`), but ideally we should have one for each of these attributes where the value is constrained like this.

brandur-stripe:
@ob-stripe Yeah, great point. There's a slight disadvantage in the enum in that there's no way for users to access a new value without first reporting the problem to us (given our current machinery, it's unlikely we'll notice ourselves), but we have this problem already for properties on classes anyway, so I wouldn't have a problem with trying this out.

I've updated the pull request. Can you take a look?

PR #1129, opened 2018-03-08
(WIP) Consider improving StripeEvent polymorphic reference

fred-stripe:
Currently in order to do something useful with a webhook event's data you need to jump through a few hoops:

```c#
StripeEvent event = StripeEventUtility.ConstructEvent(…);

// ConstructedEvent.Data.Object contains the event details
// Since JSON.NET parsed *all* of the event JSON, this is a Newtonsoft.Json.Linq.JToken object[1]
// Since we've already registered the Stripe.NET deserializers we can dynamically
// cast this into the type of object we need.

if (event.Type == "charge.successful") {
   StripeCharge Charge = ConstructedEvent.Data.Object.ToObject(typeof(StripeCharge));
}
else if (event.Type == …) {
  …
}
```

I don't like this approach for two reasons:

1. You end up with an if/else type ladder (or switch) mapping event types to entity types. This feels like something that could be encoded better, possibly by customizing StripeEvent's deserializer.
2. In order to get the type of object you *actually* want, you have to also know internal details of the StripeEventData object as well as how JSON.NET works internally, which feels like two leaky abstractions.

I'm not sure what approach we'd favor long-run for #1, but for the second I propose this as an API:

```c#
StripeEvent event = StripeEventUtility.ConstructEvent(…);

if (event.Type == "charge.successful") {
   StripeCharge Charge = (StripeCharge)ConstructedEvent.Data;
}
else if (event.Type == …) {
  …
}
```

This is actually pretty easy to accomplish with by defining an explicit conversion operator[2] for each kind of object that can be included in StripeEventData, and also has the nice property of causing a compile-time warning if one attempts to cast to a type of object that can't occur in an event's payload. I wrote an example in this PR. The downside is that this requires one conversion method per event payload type.

I'm also not sure that event type -> event payload type is well mapped in the API, which might be a future enhancement for the OpenAPI spec.

This PR is mainly speculative, but I'd love thoughts on this line of improvement.

At the very least I want to commit the example of how to typecast `event.Data.Object` so that we can use it as a reference later on.

[1] https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_Linq_JToken_ToObject.htm
[2] https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/explicit

### DIFF
--- a/src/Stripe.Tests.XUnit/events/_fixture.cs
+++ b/src/Stripe.Tests.XUnit/events/_fixture.cs
@@ -7,7 +7,7 @@ namespace Stripe.Tests.Xunit
     public class events_fixture
     {
         // this was sent in the header Stripe-Signature along with the body of event.json in a test webhook
-        internal string StripeSignature => "t=1493329224,v1=dca14f723dfa3bf47a310c3d6c3aff8bdb2534263d051dd2613ece097b8bdea4,v0=63f3a72374a733066c4be69ed7f8e5ac85c22c9f0a6a612ab9a025a9e4ee7eef";
+        internal string StripeSignature => "t=1493329224,v1=eb27ce0245b4a641852b443c47054f6ca2719048270028ce9fcf240c8c77f9e6,v0=63f3a72374a733066c4be69ed7f8e5ac85c22c9f0a6a612ab9a025a9e4ee7eef";
         internal int EventTimestamp => 1493329224;
         internal string StripeJson { get; set; }
         internal string StripeSecret => "whsec_H68eTX02a4bCbiQOOoSAsIytOvuWZrQC";
--- a/src/Stripe.Tests.XUnit/events/event.json
+++ b/src/Stripe.Tests.XUnit/events/event.json
@@ -1,43 +1,89 @@
 ﻿{
-  "id": "evt_1ADKLLLWlyqCpbLp0qRyqExL",
-  "object": "event",
-  "account": "acct_CONNECT",
-  "api_version": "2017-05-25",
-  "created": 1493329207,
-  "data": {
-    "object": {
-      "id": "ii_1ACxoELWlyqCpbLpBSzyGWHX",
-      "object": "invoiceitem",
-      "amount": 1000,
-      "currency": "usd",
-      "customer": "cus_AY3wPzBBv9DU2H",
-      "date": 1493242586,
-      "description": "Test Invoice Item",
-      "discountable": false,
-      "invoice": "in_1ADKLKLWlyqCpbLpMhrRBk7u",
-      "livemode": false,
-      "metadata": {
-        "A": "Value-A",
-        "B": "Value-B"
-      },
-      "period": {
-        "start": 1493242586,
-        "end": 1493242586
-      },
-      "plan": null,
-      "proration": false,
-      "quantity": null,
-      "subscription": null
+    "id": "evt_1C3T3NDTLHiiDklrSQa8SGw8",
+    "object": "event",
+    "account": "acct_CONNECT",
+    "api_version": "2017-05-25",
+    "created": 1520532081,
+    "data": {
+        "object": {
+            "id": "ch_1C3T3MDTLHiiDklrEkWcj4XK",
+            "object": "charge",
+            "amount": 1000,
+            "amount_refunded": 0,
+            "application": "ca_8JqC6RrO75GM2pIx953mFymukVxpJt4n",
+            "application_fee": null,
+            "balance_transaction": "txn_1C3T3MDTLHiiDklrjexvdDXi",
+            "captured": true,
+            "created": 1520532080,
+            "currency": "usd",
+            "customer": null,
+            "description": null,
+            "destination": null,
+            "dispute": null,
+            "failure_code": null,
+            "failure_message": null,
+            "fraud_details": {},
+            "invoice": null,
+            "livemode": false,
+            "metadata": {},
+            "on_behalf_of": null,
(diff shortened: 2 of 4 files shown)


fred-stripe:
Heh, well, that seems to have made AppVeyor unhappy with a linker error.

```events\when_constructing_an_event.cs(79,35): error CS0656: Missing compiler required member 'Microsoft.CSharp.RuntimeBinder.CSharpArgumentInfo.Create' [C:\projects\stripe-dotnet\src\Stripe.Tests.Xunit\Stripe.Tests.XUnit.csproj]
```

Ignoring that, ptal @anelder-stripe @ob-stripe @den-stripe

brandur-stripe:
Neat. I didn't know about `explicit operator`. A few questions:

1. Just to be clear, I guess you didn't end up trying an implementation here? The diff looks like it's just a few changes to fixtures and tests.
2. Would we need to define an `explicit operator` for every possible type that can go in an event's data? If so, it seems to me that the current implementation has a bid of an advantage in that there's a little less maintenance overhead and risk of falling out-of-date.

fred-stripe:
> Just to be clear, I guess you didn't end up trying an implementation here? The diff looks like it's just a few changes to fixtures and tests.

@brandur-stripe apologies, I forgot to push up the last commit I made. Here's the `explicit operator` implementation:

https://github.com/stripe/stripe-dotnet/pull/1129/files#diff-a03534421cece7cddff7563292301859R13

> Would we need to define an explicit operator for every possible type that can go in an event's data? If so, it seems to me that the current implementation has a bid of an advantage in that there's a little less maintenance overhead and risk of falling out-of-date.

Unfortunately yes, there's no way to make a templated cast operator like in C++ e.g. `explicit operator<T>`, nor is there `friend` to allow the conversion to be defined alongside any new StripeObject we define. Ultimately this just ends up being syntactic sugar, and you're right that it risks falling out of date.

I might favor a version that allows for the following syntax instead? 

```ConstructedEvent.Data.Object.As(typeof(StripeCharge))```

That would require some more specific Data.Object type that has an  "As" method but it would decouple us a bit from JSON.NET. I don't know that swapping out JSON libraries is a thing we expect to do, but it feels weird that you have to "know" that Data.Object is a JObject. Another thing that just occurred to me is that because `ToObject` only exists on `JObject` you 1/ won't get Intellisense on `ConstructedEvent.Data.Object` and so can't discover it can be cast, and 2/ It appears that the existence of this method is only tested for at runtime, leading to a latent & surprising issue if the programmer doesn't know what methods are available here.

So, dropping `explicit operator` (because it's going to be hard to maintain) there are two alternative changes that I think would help here:

1/ Type StripeEventData.Object as JObject, which would allow Intellisense and make this moderately better -or-
2/ Introduce an intermediate "box" type for Data.Object which provides a discoverable wrapper around JObject.ToObject().

Of these I slightly prefer the latter because returning an intermediate JSON object from a different library feels wrong.

**Comment by fred-stripe** [comment]
Also, just wanted to note that these are equivalent with JObject:

```c#
var charge = ConstructedEvent.Data.Object.ToObject(typeof(StripeCharge));
var charge = ConstructedEvent.Data.Object.ToObject<StripeCharge>();
```

brandur-stripe:
> So, dropping explicit operator (because it's going to be hard to maintain) there are two alternative changes that I think would help here:
>
> 1/ Type StripeEventData.Object as JObject, which would allow Intellisense and make this moderately better -or-
> 2/ Introduce an intermediate "box" type for Data.Object which provides a discoverable wrapper around JObject.ToObject().

Hm, yeah, it might be a good idea to consider changing the type of `Object`. I might be missing something, but I don't think there's really a good reason to make it `dynamic` — it doesn't really buy you anything, and makes use more dangerous (in the sense that it can throw runtime errors) compared to just using the standard C# `object`.

Just as a slight alternative proposal, could we just implement a helper on `StripeEventData` to make typed access easier? Maybe with this signature:

``` c#
        public T ObjectAs<T>() where T: StripeEntity {
            ...
        }
```

It could also be well-documented for IntelliSense friendliness and perform the cast safety and on failure, extract the actual name of the object and tell the user something like `It's not possible to cast an object of type "charge" to type "StripePlan".`

You're saying that the main issue with the current implementation is discoverability right? I think this would fix that problem and also avoid having to maintain a bunch of per-type operator implementations.

den-stripe:
I am also in favor of using `ObjectAs<T>()`, you can also define https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods in the `StripeCharge` itself.

```
public static class StripeChargeExtensions
{
        public static StripeCharge? AsStripeCharge(this StripeEventData data)
        {
              if (!Object.CanBeConvertedToStripeCharge) {
                   return null;
              }
              return (StripeCharge?)Object.ToObject(typeof(StripeCharge));
        }
}   
```

bchavez:
```csharp
var charge1 = StripeEventUtility.ParseEventDataItem<StripeCharge>(webhook.Data.Object);
````
```csharp
if( webhook.Data.Object is StripeCharge charge2 )
{
}
```
https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching

:thinking: :sparkles: ***["Do you ponder the manner of things... yeah yeah... like glitter and gold..."](https://www.youtube.com/watch?v=IrV90gXmOpA)***

ob-stripe:
Superseded by #1322. `Object` is now declared has an `IHasObject` interface rather than `dynamic`, and users can check the concrete type at runtime.

PR #1135, opened 2018-03-15
Add ProductId parameter to StripePlanListOptions

### LINKED ISSUE #1134: Please add support for "product" for the list all plans request. (API 2018-02-28)
I assumed this would be simply adding a field in StripePlanListOptions?

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries @DataEstate

Adds support for the `product` field when listing plans: https://stripe.com/docs/api/dotnet#list_plans-product.

I didn't add any tests because this is fairly trivial and a well-established pattern through the library (e.g. [here](https://github.com/stripe/stripe-dotnet/blob/2d9f5b27ae57b2a98333e538859281a944b9b033/src/Stripe.net/Services/Subscriptions/StripeSubscriptionListOptions.cs#L16-L20) or [here](https://github.com/stripe/stripe-dotnet/blob/45a10753b5ca16be4646f62cf865173b61ac98eb/src/Stripe.net/Services/Invoices/StripeInvoiceListOptions.cs#L31-L32)).

Fixes #1134.

### DIFF
--- a/src/Stripe.net/Services/Plans/StripePlanListOptions.cs
+++ b/src/Stripe.net/Services/Plans/StripePlanListOptions.cs
@@ -4,5 +4,13 @@ namespace Stripe
 {
     public class StripePlanListOptions : StripeListOptions
     {
+        [JsonProperty("created")]
+        public StripeDateFilter Created { get; set; }
+
+        /// <summary>
+        /// Only return plans for the given product.
+        /// </summary>
+        [JsonProperty("product")]
+        public string ProductId { get; set; }
     }
 }


src/Stripe.net/Services/Plans/StripePlanListOptions.cs
```diff
@@ -4,5 +4,10 @@ namespace Stripe
 {
     public class StripePlanListOptions : StripeListOptions
     {
+        /// <summary>
+        /// Only return plans for the given product.
+        /// </summary>
+        [JsonProperty("product")]
+        public string ProductId { get; set; }
```
remi-stripe: we should add `created` while we're at it like we do for charges: https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/Charges/StripeChargeListOptions.cs

remi-stripe:
LGTM if you want to go ahead though I mentioned a small addition

src/Stripe.net/Services/Plans/StripePlanListOptions.cs
ob-stripe: 👍  Nice catch!

PR #1140, opened 2018-03-17
Add support for multi-level Expand feature

remi-stripe:
All options class now have the ability to pass a list of properties to expand. Until then, we had to add the `ExpandXXXX` property to the service and it only supported one level.

One downside is that this could cause crashes if you're on a version of the library that does not expand a specific property to be expandable.

The way I added this means that it is not a breaking change and both versions are supported. It would likely be best to remove the old way afterwards (or in this PR) to avoid conflicts though.

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.Tests.XUnit/_infrastructure/expanding_resources.cs
+++ b/src/Stripe.Tests.XUnit/_infrastructure/expanding_resources.cs
@@ -1,0 +1,49 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using FluentAssertions;
+using Xunit;
+using Stripe.Infrastructure;
+
+namespace Stripe.Tests.Xunit
+{
+    public class expanding_resources
+    {
+        public class TestOptions : StripeBaseOptions
+        {
+            [JsonProperty("a_string")]
+            public string AString { get; set; }
+        }
+
+        public class TestService : StripeService
+        {
+            public TestService() : base(null)
+            {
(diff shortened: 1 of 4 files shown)


ob-stripe:
Nice!

I'm happy to add this to the library, however I wonder if we should make this the "official" way of expanding attributes as it makes the developer responsible for passing proper expandable values. That said, we do the same thing in stripe-go, and I'm really not sure what the interface would look like if we wanted to add support for multi-level expansions in a type-safe manner.

cc @brandur-stripe @fred-stripe @anelder-stripe in case you have any thoughts on this topic.

remi-stripe:
@ob-stripe part of the reason I added was that I discovered that our library had multiple bugs already. In the other PR you can see that some services had the `ExpandXX` property while that property was not defined as Expandable in the object class or the opposite where the property supported expansion but the service did not.

I could not come up with a viable way to support multi-level expansion without this unless we explicitly added things like `ExpandTransferAndDestinationPaymentAndBalanceTransaction` which seemed unmaintainable.

ob-stripe:
> I could not come up with a viable way to support multi-level expansion without this unless we explicitly added things like ExpandTransferAndDestinationPaymentAndBalanceTransaction which seemed unmaintainable.

Oh, definitely. I was thinking it would be neat to have something like `Expand(Transfer.DestinationPayment.BalanceTransaction)`, and leverage the type system to catch unexpandable attributes at compile time.

That would likely be significantly more complex to implement though, so probably not worth the effort. In the meantime, this PR solves a common pain point for stripe-dotnet users, so I'm going to pull this in.

PR #1143, opened 2018-03-20
Use parser plugins during SubscriptionItemUpdatedPlugin item creation

### LINKED ISSUE #1142: POST /v1/subscriptions/{Id} isn't serializing Metadata dictionary.
It looks like the Metadata property in [StripeSubscriptionItemUpdateOption.cs](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/Subscriptions/StripeSubscriptionItemUpdateOption.cs) isn't being serialized to a JSON object when sent to the Stripe API.
I believe this is due to StripeSubscriptionItemUpdateOption.cs implements the flag interface INestedOptions.cs which changes property serialization handling in [ParameterBuilder.cs](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Infrastructure/ParameterBuilder.cs#L30)

Stripe.net 14.0.0
dotnet core 2.x
Windows/kestrel host.

(Sorry about the horrendous snipping tool job)
image

TobiasMelen:
This PR makes the SubscriptionItemPlugin use the parser plugins in RequestStringBuilder when creating request values for items in subscription update. This solves an issue with the input metadata parsing. The Metadata dictionary property needs to be parsed by DictionaryParserPlugin to create a valid Stripe request.
An inline JsonPropertyAttribute is created, which isn't pretty. Any other change would require a larger refactor. 
I had to run the project locally with XUnit 2.3.0 due to Visual Studio issues, so the testcode should probably be validated on XUnit 2.2.0.

Fixes #1142.

### DIFF
--- a/src/Stripe.Tests.XUnit/subscriptions/_fixture.cs
+++ b/src/Stripe.Tests.XUnit/subscriptions/_fixture.cs
@@ -37,7 +37,12 @@ namespace Stripe.Tests.Xunit
                 Items = new List<StripeSubscriptionItemUpdateOption>
                 {
                     new StripeSubscriptionItemUpdateOption { Id = Subscription.Items.Data[0].Id, Deleted = true },
-                    new StripeSubscriptionItemUpdateOption { Id = Subscription.Items.Data[1].Id, Quantity = 5 }
+                    new StripeSubscriptionItemUpdateOption
+                    {
+                        Id = Subscription.Items.Data[1].Id,
+                        Quantity = 5,
+                        Metadata = new Dictionary<string, string>{ {"key", "value"} }
+                    }
                 }
             };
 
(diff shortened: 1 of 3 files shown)


TobiasMelen:
#1142

ob-stripe:
Hey @TobiasMelen, thanks for the contribution!

> An inline JsonPropertyAttribute is created, which isn't pretty. Any other change would require a larger refactor.

Yeah, it's slightly hacky, but the state of parameter encoding in stripe-dotnet isn't great (there's a lot of special casing) and subscription items encoding in particular has required special casing anyway in our other language libraries. Your fix is fine with me, pending a larger rewrite of the parameter encoding logic.

PR #1145, opened 2018-03-22
Added ConfigureAwait(false) to all awaited tasks.

### LINKED ISSUE #1141: Use ConfigureAwait(false) for all async method calls
Libraries are encouraged to always use `ConfigureAwait(false)` on async methods to prevent possible deadlocks. This doesn't seem to be happening at all in the .NET Stripe client library. For example, this method from `StripeBasicService`:

```c#
public virtual async Task<EntityReturned> GetEntityAsync(string url, StripeRequestOptions requestOptions, CancellationToken cancellationToken, StripeBaseOptions options = null)
{
    return Mapper<EntityReturned>.MapFromJson(
        await Requestor.GetStringAsync(
                this.ApplyAllParameters(options, url),
                SetupRequestOptions(requestOptions),
                cancellationToken
        ).ConfigureAwait(false)
    );
}
```

I wouldn't mind submitting a pull request, if I can find some time to make these changes myself. My organization is simply considering using Stripe at this point, so until or if we actually pull the trigger on that, I'd have to do this in my personal time. Still, let me know if you'd like any help.

For what it's worth, overall I'm very pleased with how well the Stripe .NET library is implemented. Braintree's is a nightmare. So, kudos.

More information:
* https://blogs.msdn.microsoft.com/seteplia/2018/02/14/one-user-scenario-to-rule-them-all/
* https://channel9.msdn.com/Series/Three-Essential-Tips-for-Async/Async-library-methods-should-consider-using-Task-ConfigureAwait-false-

chrisdpratt:
This PR adds `ConfigureAwait(false)` to all awaited tasks, throughout the library. It is not necessary to add this to tasks that are not awaited, so in that regard, I also took the opportunity to optimize a few methods that had the `async` keyword only because they were awaiting a task before returning it. Now, all library methods only await tasks when necessary, and simply return the task directly otherwise.

Fixes #1141

### DIFF
--- a/src/Stripe.net/Infrastructure/Requestor.cs
+++ b/src/Stripe.net/Infrastructure/Requestor.cs
@@ -61,8 +61,8 @@ namespace Stripe.Infrastructure
 
         private static StripeResponse ExecuteRequest(HttpRequestMessage requestMessage)
         {
-            var response = HttpClient.SendAsync(requestMessage).Result;
-            var responseText = response.Content.ReadAsStringAsync().Result;
+            var response = HttpClient.SendAsync(requestMessage).ConfigureAwait(false).GetAwaiter().GetResult();
+            var responseText = response.Content.ReadAsStringAsync().ConfigureAwait(false).GetAwaiter().GetResult();
 
             var result = BuildResponseData(response, responseText);
 
@@ -105,14 +105,15 @@ namespace Stripe.Infrastructure
 
         private static async Task<StripeResponse> ExecuteRequestAsync(HttpRequestMessage requestMessage, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var response = await HttpClient.SendAsync(requestMessage, cancellationToken);
+            var response = await HttpClient.SendAsync(requestMessage, cancellationToken).ConfigureAwait(false);
+            var responseText = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
 
-            var result = BuildResponseData(response, await response.Content.ReadAsStringAsync());
+            var result = BuildResponseData(response, responseText);
 
             if (response.IsSuccessStatusCode)
                 return result;
 
-            throw BuildStripeException(result, response.StatusCode, requestMessage.RequestUri.AbsoluteUri, await response.Content.ReadAsStringAsync());
+            throw BuildStripeException(result, response.StatusCode, requestMessage.RequestUri.AbsoluteUri, responseText);
         }
 
 
--- a/src/Stripe.net/Services/Account/StripeAccountService.cs
+++ b/src/Stripe.net/Services/Account/StripeAccountService.cs
@@ -80,7 +80,7 @@ namespace Stripe
             return Mapper<StripeAccount>.MapFromJson(
                 await Requestor.PostStringAsync(this.ApplyAllParameters(createOptions, $"{Urls.BaseUrl}/accounts", false),
                 SetupRequestOptions(requestOptions),
-                cancellationToken)
+                cancellationToken).ConfigureAwait(false)
             );
         }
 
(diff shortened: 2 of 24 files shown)


jaymedavis:
Hmm... I’m not sure you want to do this (outside of the async context). The danger of ConfigureAwait in a synchronous context is that the thread can jump ship. The reason it has ExecuteRequest and ExecuteRequestAsync is for this reason. The changes to async make sense, but I would worry that asp.net users would lose blocking when they want it. Granted - there is an issue now were a deadlock can occur from the block, perhaps this fixes that? 

referenced all the comments here: https://stackoverflow.com/questions/13489065/best-practice-to-call-configureawait-for-all-server-side-code

jaymedavis:
Are you writing a UI app or web @chrisdpratt?

chrisdpratt:
`ConfigureAwait` doesn't have anything to do with releasing a thread or not. By default, an async task attempts to complete on the original thread. `ConfigureAwait(false)` merely allows it to complete on any thread, throwing away the synchronization context. In a sync method, there's no opportunity for a thread jump, so work is always completed on the original thread. That means there's especially no need for a synchronization context in this case, and `ConfigureAwait(false)` is actually *more important* when blocking on async tasks, than actually awaiting them.

The *best* thing you can do is don't block on async tasks. However, given that the Stripe library provides synchronous methods but uses async-only `HttpClient`, it does this. Adding `ConfigureAwait(false)` helps to minimize the chance of deadlocks (though deadlocks are still possible).

jaymedavis:
From the source I posted:

‘’’
However, with ASP.NET Web Api, if your request is coming in on one thread, and you await some function and call ConfigureAwait(false) that could potentially put you on a different thread when you are returning the final result of your ApiController function.
‘’’

I wouldn’t trust this until it’s tested on an asp.net site that use controllers with sync code.

Just my 2 cents

chrisdpratt:
That's talking about actually using async as async, not blocking on async. Regardless, you can't use `ConfigureAwait(false) in ASP.NET (not Core) because `HttpContext` is thread local. In other words, you *need* the synchronization context to continue on the original thread. However, the synchronization context only needs to be kept on the main async task. Anything in a library can safely use `ConfigureAwait(false)` for it's own internal stuff. But, yes, if you were to actually utilize one of the Stripe methods in you ASP.NET Web App, you should not use `ConfigureAwait(false)` *there*. In fact, you'll raise an exception if you try.

brandur-stripe:
Thanks @chrisdpratt! Reading through the sources, this change looks sane to me.

> But, yes, if you were to actually utilize one of the Stripe methods in you ASP.NET Web App, you should not use ConfigureAwait(false) there. In fact, you'll raise an exception if you try.

Just for clarification, this is an internal error thrown by `ConfigureAwait(false)` throw because it's smart enough to know that it shouldn't be called from user code right?

I just want to make sure we never go from a situation where a previous usage `stripe-dotnet` code goes from not throwing an exception to throwing an exception, even if it was being used in a degenerate way.

chrisdpratt:
@brandur-stripe No exception will ever be thrown simply because `ConfigureAwait(false)` is used internally in the Stripe library. The exception would only occur if the end-user uses `ConfigureAwait(false)` in *their* own code calling the Stripe library, *and* they are running in a context where the synchronization context must be preserved (ASP.NET, UI threads, etc.). Regardless, it would be user code, not library code, responsible. As long as the synchronization context is captured on the main call, things like ASP.NET are good.

The reason this is important for library code is because of the "first call" problem. Since tasks return "hot" (i.e. already running), the synchronization context is discarded only *after* the first task that uses `ConfigureAwait(false)` completes. If async call is blocked, then you can end up getting deadlocked before that can occur. The only safe way to prevent deadlocks like this is to use `ConfigureAwait(false)` all the way down the chain. In other words, by *not* implementing `ConfigureAwait(false)` in the Stripe library, you're preventing end-users who *do* need to use `ConfigureAwait(false)` to prevent deadlocks from doing so.

brandur-stripe:
> The reason this is important for library code is because of the "first call" problem. Since tasks return "hot" (i.e. already running), the synchronization context is discarded only after the first task that uses ConfigureAwait(false) completes. If async call is blocked, then you can end up getting deadlocked before that can occur. The only safe way to prevent deadlocks like this is to use ConfigureAwait(false) all the way down the chain. In other words, by not implementing ConfigureAwait(false) in the Stripe library, you're preventing end-users who do need to use ConfigureAwait(false) to prevent deadlocks from doing so.

Great explanation. Thanks!

brandur-stripe:
Released as 15.2.0.

Issue #990, opened 2017-08-15
getFrameworkFromRegistry() throwing null reference exception on Mac


mickbyrne:
I'm trying to use the Stripe.net package in our server-side API. I started with a simple `customerService.create(...)` call, but got an exception with the following stack trace:

> Object reference not set to an instance of an object
> at Stripe.Infrastructure.Client.getFrameworkFromRegistry () [0x0001a] in <d01b79e1f9c84725be1db806a232e677>:0 \n  at Stripe.Infrastructure.Client.getSystemInformation () [0x00033] in <d01b79e1f9c84725be1db806a232e677>:0 \n  at Stripe.Infrastructure.Client.GetClientUserAgentString () [0x00077] in <d01b79e1f9c84725be1db806a232e677>:0 \n  at Stripe.Infrastructure.Client.ApplyClientData () [0x0000b] in <d01b79e1f9c84725be1db806a232e677>:0 \n  at Stripe.Infrastructure.Requestor.GetRequestMessage (System.String url, System.Net.Http.HttpMethod method, Stripe.StripeRequestOptions requestOptions, System.Boolean useBearer) [0x000af] in <d01b79e1f9c84725be1db806a232e677>:0 \n  at Stripe.Infrastructure.Requestor.PostString (System.String url, Stripe.StripeRequestOptions requestOptions) [0x00006] in <d01b79e1f9c84725be1db806a232e677>:0 \n  at Stripe.StripeCustomerService.Create (Stripe.StripeCustomerCreateOptions createOptions, Stripe.StripeRequestOptions requestOptions) [0x00014] in <d01b79e1f9c84725be1db806a232e677>:0 \n  at [ my code ]

I'm running Visual Studio for Mac using the .NET Framework 4.5.2 (Visual Studio for Mac is what has replaced Xamarin Studio). This uses the Mono 5.0.1.1 runtime.

The actual code that triggers this issue is super basic. Here's a gist of the relevant bits, https://gist.github.com/mickbyrne/df3835f34f5100a45e58d68002981213

This is a blocker for us at the moment. Do we need to roll our own HTTP handler for this (similar to issue https://github.com/stripe/stripe-dotnet/issues/567 re TLS?)

mickbyrne:
I've verified that the code in the `Stripe.Infrastructure.Client.getFrameworkFromRegistry()` is throwing the error when running on a Mac. This is a simple method:

	private static string getFrameworkFromRegistry()
	{
[...]

PR #1151, opened 2018-04-06
Fix GetClientUserAgentString()

ob-stripe:
Tentative fix for #990.

### DIFF
--- a/src/Stripe.net.Tests/infrastructure/when_building_the_client.cs
+++ b/src/Stripe.net.Tests/infrastructure/when_building_the_client.cs
@@ -33,11 +33,5 @@ namespace Stripe.Tests
         {
             var clientData = JToken.Parse(request.Headers.GetValues("X-Stripe-Client-User-Agent").FirstOrDefault());
         };
-
-        It should_have_the_net45_for_uname = () =>
-        {
-            var clientData = JToken.Parse(request.Headers.GetValues("X-Stripe-Client-User-Agent").FirstOrDefault());
-            clientData.SelectToken("uname").ToString().ShouldContain("net45.");
-        };
     }
 }
--- a/src/Stripe.net/Infrastructure/Client.cs
+++ b/src/Stripe.net/Infrastructure/Client.cs
@@ -6,6 +6,8 @@ using System.Reflection;
 using Newtonsoft.Json;
 #if NET45
 using Microsoft.Win32;
+#else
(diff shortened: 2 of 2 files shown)


ob-stripe:
After a lot of efforts, I was able to run the XUnit test suite using Visual Studio for Mac and .NET Framework 4.6! AFAICT it works fine.

For reference, here are the differences in the `X-Stripe-Client-User-Agent` string:

- OS X, target `net46` (from Visual Studio for Mac):
```
"lang_version": ".NET Framework 4.5+",
"os_version": "Unix 17.4.0.0",
"mono_version": "5.8.1.0 (2017-10/6bf3922f3fd Thu Mar  8 16:47:23 EST 2018)"
```

- OS X, target `netcoreapp1.1`:
```
"lang_version": ".NET Core 4.6.26011.01",
"os_version": "Darwin 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64"
```

- Windows, target `net46` (from Appveyor):
```
"lang_version": ".NET Framework 4.5+",
"os_version": "Microsoft Windows 10.0.14393 "
```

- Windows, target `netcoreapp1.1` (from Appveyor):
```
"lang_version": ".NET Core 4.6.25211.01",
"os_version": "Microsoft Windows 10.0.14393 "
```

I've removed the registry lookup entirely -- while it was possible to simply add a check for non-Windows platform, I'm fairly sure the registry lookup was ~useless because it was not returning the current runtime, but rather the most recent framework version available on the machine. (E.g. on Appveyor, it reported "4.7.1 or later" even when targeting 4.6.)

As far as I can tell, there's no good way of checking the actual .NET Framework version at runtime -- there's [`System.Environment.Version`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.version?view=netframework-4.5.1#System_Environment_Version), but it only lets you distinguish between 4.5 and 4.6, and Microsoft actually recommends using the registry instead. At that point I threw my hands in the air and decided "4.5+" was good enough.

r? @brandur-stripe 
cc @stripe/api-libraries

brandur-stripe:
Wow, you are a hero for tracking this one down @ob-stripe! Thanks.

I'll admit up front that I've never looked at these internals before, but this diff and your rationale look correct. I can't believe how difficult it is to do this!

LGTM.

Issue #1153, opened 2018-04-12
"sepa_debit[ideal]" property is missing in StripeSourceCreateOptions 


ramedey:
This property is required for Recurring iDEAL Payments with Sources:

https://stripe.com/docs/sources/ideal/recurring#create-sepa-source-server-side

There should be something like:
    
```
public class StripeSourceCreateOptions : StripeBaseOptions, ISupportMetadata
{
       ....
        [JsonProperty("sepa_debit[ideal]")]
        public string SepaDebitIdeal { get; set; }
       .....
}
```

Thanks,

ob-stripe:
Fixed in 15.4.0 which will be available on NuGet shortly.

PR #1154, opened 2018-04-13
Add the ability to create a SEPA debit source from an Ideal one.

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/1153

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Services/Sources/StripeSourceCreateOptions.cs
+++ b/src/Stripe.net/Services/Sources/StripeSourceCreateOptions.cs
@@ -44,6 +44,9 @@ namespace Stripe
         [JsonProperty("sepa_debit[iban]")]
         public string SepaDebitIban { get; set; }
 
+        [JsonProperty("sepa_debit[ideal]")]
+        public string SepaDebitIdealSourceId { get; set; }
+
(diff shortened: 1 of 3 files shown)


src/Stripe.net/Services/Sources/StripeSourceCreateOptions.cs
```diff
@@ -44,6 +44,9 @@ public class StripeSourceCreateOptions : StripeBaseOptions, ISupportMetadata
         [JsonProperty("sepa_debit[iban]")]
         public string SepaDebitIban { get; set; }
 
+        [JsonProperty("sepa_debit[ideal]")]
+        public string SepaDebitIdealSource { get; set; }
```
remi-stripe: I chose to add "Source" at the end as it takes a source id and in the future we might add an `ideal` hash so wanted to protect future us

ob-stripe: I think this should be named `SepaDebitIdealSourceId` in order to be consistent with other parameters that expect an ID.

PR #1160, opened 2018-04-18
Add support for internal properties on the Card resource

remi-stripe:
Those new properties are for internal use but we do need them in stripe-dotnet. We have shipped them a long time ago in stripe-go too: https://github.com/stripe/stripe-go/pull/221

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Released in 15.6.1.

Issue #1162, opened 2018-04-23
StripeBalanceTransaction has incorrect Source type


lxalln:
The [`BalanceTransaction`](https://stripe.com/docs/api#balance_transaction_object) documentation suggests that the expandable `Source` property can basically be anything.

The current type of `Source` is `StripeSource`, which from what I can tell, is always going to be wrong. The source of a BalanceTransaction is not actually a StripeSource, but more likely a `StripeCharge` or `StripeApplicationFee`.

Because of this, the expansion feature of balance history is unusable for us.

From what I can see, the only type that would make sense for `Source` on `StripeBalanceTransaction` would be `StripeEntityWithId` and it should be de-serialized into the correct type. The client would then need to handle this themselves by looking at the `Type` property on `StripeBalanceTransaction`

PR #1165, opened 2018-04-23
Properly deserialize source on BT during expand

remi-stripe:
This PR adds support for deserializing `source` on a Balance Transaction properly. It used to map to a `Source` class which is a "parent class" for payment methods and has nothing to do with balance transaction.

There's a new custom deserializer and type for this that looks at the balance transaction's source `object` property and deserializes the right type based on this.

This PR also adds support for local JSON files as fixtures for the new test suite to properly test deserializing all source types.

Fixes https://github.com/stripe/stripe-dotnet/issues/1162
cc @stripe/api-libraries

### DIFF
--- a/src/StripeTests/Entities/StripeBalanceTransactionTest.cs
+++ b/src/StripeTests/Entities/StripeBalanceTransactionTest.cs
@@ -1,0 +1,99 @@
+namespace StripeTests
+{
+    using System;
+    using System.IO;
+    using System.Reflection;
+    using System.Text;
+
+    using Newtonsoft.Json;
+    using Stripe;
+    using Xunit;
+
+    public class StripeBalanceTransactionTest : BaseStripeTest
+    {
+        [Fact]
+        public void Deserialize()
+        {
+            string json = GetFixture("/v1/balance/history/txn_123");
+            var balanceTransaction = Mapper<StripeBalanceTransaction>.MapFromJson(json);
+            Assert.NotNull(balanceTransaction);
+            Assert.IsType<StripeBalanceTransaction>(balanceTransaction);
+            Assert.NotNull(balanceTransaction.Id);
+            Assert.Equal("balance_transaction", balanceTransaction.Object);
+        }
+
+        [Fact]
+        public void DeserializeWithExpansions()
+        {
+            // We test all balance transaction possible source types to ensure the deserializer
+            // works as expectes.
+            var json = GetResourceAsString("api_fixtures.balance_transaction_with_expansion.json");
+            var balanceTransactions = Mapper<StripeList<StripeBalanceTransaction>>.MapFromJson(json);
+
+            Assert.NotNull(balanceTransactions);
+            Assert.NotNull(balanceTransactions.Data);
+            Assert.Equal(9, balanceTransactions.Data.Count);
+
+            Assert.NotNull(balanceTransactions.Data[0]);
+            Assert.Equal("balance_transaction", balanceTransactions.Data[0].Object);
+            Assert.NotNull(balanceTransactions.Data[0].Source);
+            Assert.Equal(BalanceTransactionSourceType.ApplicationFee, balanceTransactions.Data[0].Source.Type);
+            Assert.NotNull(balanceTransactions.Data[0].Source.ApplicationFee);
+            Assert.Equal("application_fee", balanceTransactions.Data[0].Source.ApplicationFee.Object);
+
+            Assert.NotNull(balanceTransactions.Data[1]);
+            Assert.Equal("balance_transaction", balanceTransactions.Data[1].Object);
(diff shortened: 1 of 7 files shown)


ob-stripe:
This looks mostly on the right track to me. Thanks for taking a stab at this @remi-stripe!

> How can I handle testing this? Should I create BTs for each type and confirm they get deserialized properly on expand like I do in the Charge one?

Ugh. I'd rather avoid adding complex tests that use the real API. The test suite already takes forever to run, and these tests will either be painful to port to stripe-mock or will be scraped entirely. Since it's unlikely stripe-mock will support returning different `source` objects under `balance_transaction` in the near future, I suppose the right thing to do would be to test the various cases using a local set of JSON fixtures, but right now stripe-dotnet does not support stubbing requests...

@brandur-stripe Any thoughts here?

> How can I handle deserializing new types into a generic StripeObject? StripeEntityWithId is abstract so I could create a new type that would be StripeObject that inherits from it but not sure it's the right direction. The goal would be to be able to access the raw JSON if you want things on types we have not yet added to stripe-dotnet (or older versions of the lib). For now I set the type to Unknown.

Honestly, I'd rather throw an exception with a message saying to contact support so we can update the library with missing types. Just returning the raw JSON does not sound like a great developer experience. I understand that updating stripe-dotnet can be painful because of the pinned API version and semi-frequent major version bumps, but I don't think we're going to add new types often enough that we should have an escape hatch specifically for this case.

(Worst case, the JSON of the entire response is always accessible via the `StripeResponse` object.)

> I modeled this on the Source.cs implementation. C# does not seem to have union support but is this the best approach?

Not a C# expert™, but that seems fine to me. C unions are not really compatible with classes anyway (even in C++, you cannot have any kind of virtual functions in a union).

remi-stripe:
> Honestly, I'd rather throw an exception with a message saying to contact support so we can update the library with missing types. Just returning the raw JSON does not sound like a great developer experience. I understand that updating stripe-dotnet can be painful because of the pinned API version and semi-frequent major version bumps, but I don't think we're going to add new types often enough that we should have an escape hatch specifically for this case.

We can not do that. If we did this it'd mean that as soon as we add a new type, anyone could end up with an exception or a crash in their production application right? New types are not a breaking change and could be introduced.
I'm fine leaving as is without a custom type but at least putting `Unknown` as the enum value so that you can do a case on it in the code and look into it as a developer.

ob-stripe:
> If we did this it'd mean that as soon as we add a new type, anyone could end up with an exception or a crash in their production application right?

Ah, good point. Sigh, who decided that an expandable polymorphic attribute was a good idea anyway :shakes_fist:

So yeah, `Type=Unknown` sounds good to me then.

jaymedavis:
this is interesting:
https://social.msdn.microsoft.com/Forums/en-US/60150e7b-665a-49a2-8e2e-2097986142f3/c-equivalent-to-c-union?forum=csharplanguage

Do you think StructLayout has promise here?

jaymedavis:
Thought that might be the case. To my knowledge, .net developers rarely ever use the struct keyword anyway.

remi-stripe:
@ob-stripe I took another stab at this PR now that I better understand the test suite based on stripe-mock. Can you have a look as I think it's ready to merge.

It's a breaking change, but also it never worked so not sure what we want to do here?

src/StripeTests/Entities/StripeBalanceTransactionTest.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        [Fact]
+        public void DeserializeWithExpansions()
+        {
+            // We test all balance transaction possible source types to ensure the deserializer
+            // works as expectes.
+            var json = GetResourceAsString("api_fixtures.balance_transaction_with_expansion.json");
+            var balanceTransactions = Mapper<StripeList<StripeBalanceTransaction>>.MapFromJson(json);
+
```
ob-stripe: I would just add a couple of asserts here:
```c#
Assert.NotNull(balanceTransactions.Data)
Assert.IsTrue(balanceTransactions.Data.Count >= 9)
```
just to avoid exceptions when trying to access something that doesn't exist if the deserialization goes wrong (or if something is wrong with the fixture itself).

ob-stripe:
> It's a breaking change, but also it never worked so not sure what we want to do here?

I think it's fine to release this as a minor version. Technically, the API changed, but as you pointed out no integration could use the previous API anyway, so in practice it's more like a new feature.

src/StripeTests/Entities/StripeBalanceTransactionTest.cs
remi-stripe: @ob-stripe done! PTAL

ob-stripe:
Released as 17.4.0.

PR #1169, opened 2018-04-25
Don't use Bitcoin in tests

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Stripe [no longer supports Bitcoin](https://stripe.com/blog/ending-bitcoin-support). stripe-dotnet's test suite used Bitcoin sources in a few tests. I've either removed or modified tests to use a different payment method (ACH Credit Transfer, mostly because it's a receiver flow like Bitcoin).


brandur-stripe:
Oh, good call on this.

LGTM.

Issue #1171, opened 2018-05-02
Need InvoicePrefix for StripeCustomerCreateOptions


winzig:
When creating a customer, it is possible to [specify an invoice prefix](https://stripe.com/docs/api/dotnet#create_customer). If it's in the existing Stripe.net API, I am not seeing it. Thanks!

ob-stripe:
Released in 15.7.0.

PR #1172, opened 2018-05-02
Add support for the invoice_prefix parameter on Custom creation/update

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/1171

r? @ob-stripe 
cc @stripe/api-libraries


jaymedavis:
You probably want this on [StripeCustomer](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Entities/StripeCustomer.cs) too?

remi-stripe:
@jaymedavis Nice catch! Forgot it was not just a parameter and was also returned in the API! Fixed.

PR #1174, opened 2018-05-06
Add support for Sigma scheduled queries

Mitch528:


remi-stripe:
@Mitch528 Thank you for doing the work here! As you can see, the tests are failing because the endpoint can only be used in Live mode.

Can you confirm that this works in Live mode for you? If so, you would have to comment the test out for now until we move the library to [stripe-mock](https://github.com/stripe/stripe-mock) so that it does not break the build.

cc @stripe/api-libraries for awareness

Mitch528:
Yes, I can confirm that it works in live mode for me.

remi-stripe:
@Mitch528 Can you comment the whole test file? Also only keep the XUnit one, we're likely going to remove the other test framework in the future so one file is enough :)

brandur-stripe:
LGTM as well.

Thanks a lot for the contribution here @Mitch528! Very much appreciated.

brandur-stripe:
We'll have to wait on CI to finish before the new release is pushed out to Nuget, but this will be coming out as 15.8.0.

Issue #1175, opened 2018-05-07
Bank account fields always null in StripeToken object


jathomas9:
When calling the Get on a StripeTokenService a token is returned but all of the the bank account fields are NULL. I can see the values in the StripeResponse object. Below is the code I am using.

```
var tokenService = new StripeTokenService();
var stripeToken = tokenService.Get(*Id*);
```

For now I am just retrieving the values from the StripeReponse object

Any help would be appreciated.

remi-stripe:
@jathomas9 Thanks for the report! I pushed a fix in https://github.com/stripe/stripe-dotnet/pull/1176

jathomas9:
Awesome!

PR #1176, opened 2018-05-07
Properly handle bank account token deserialization

remi-stripe:
The current resource is not deserialized properly. I think the `xxx[yyy]` only works on options classes (for parameters) and not on deserialization.

This fixes https://github.com/stripe/stripe-dotnet/issues/1175 by deserializing the object to the existing `StripeBankAccount` class instead.

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.Tests.XUnit/tokens/creating_a_bank_account_token.cs
+++ b/src/Stripe.Tests.XUnit/tokens/creating_a_bank_account_token.cs
@@ -1,0 +1,42 @@
+using FluentAssertions;
+using Stripe.Tests.Xunit;
+using Xunit;
+
+namespace Stripe.Tests.Xunit
+{
+    public class creating_a_bank_account_token
+    {
+        private StripeToken Token { get; set; }
+
+        public creating_a_bank_account_token()
+        {
+            Token = new StripeTokenService(Cache.ApiKey).CreateAsync(
+                new StripeTokenCreateOptions
+                {
+                    BankAccount = new BankAccountOptions
+                    {
+                        AccountNumber = "000123456789",
+                        Country = "US",
+                        Currency = "USD",
+                        RoutingNumber = "110000000",
+                    }
+                }
+            ).Result;
+        }
+
+        [Fact]
+        public void not_null()
+        {
+            Token.Should().NotBeNull();
+        }
+
+        [Fact]
+        public void has_relevant_properties_set()
+        {
+            Token.Id.Should().NotBeNull();
+            Token.StripeBankAccount.Id.Should().NotBeNull();
+            Token.StripeBankAccount.Last4.Should().Be("6789");
+            Token.StripeBankAccount.RoutingNumber.Should().Be("110000000");
+        }
+    }
+}
--- a/src/Stripe.net/Entities/StripeToken.cs
+++ b/src/Stripe.net/Entities/StripeToken.cs
@@ -9,33 +9,8 @@ namespace Stripe
         [JsonProperty("object")]
         public string Object { get; set; }
 
-        // TODO: use a single property to map bank_account to a StripeBankAccount object
-        [JsonProperty("bank_account[id]")]
-        public string BankAccountId { get; set; }
-
-        [JsonProperty("bank_account[object]")]
-        public string BankAccountObject { get; set; }
-
-        [JsonProperty("bank_account[country]")]
-        public string BankAccountCountry { get; set; }
-
-        [JsonProperty("bank_account[currency]")]
-        public string BankAccountCurrency { get; set; }
-
-        [JsonProperty("bank_account[last4]")]
-        public string BankAccountLast4 { get; set; }
-
-        [JsonProperty("bank_account[status]")]
-        public string BankAccountStatus { get; set; }
-
-        [JsonProperty("bank_account[bank_name]")]
-        public string BankAccountName { get; set; }
-
-        [JsonProperty("bank_account[fingerprint]")]
-        public string BankAccountFingerprint { get; set; }
-
-        [JsonProperty("bank_account[routing_number]")]
-        public string BankAccountRoutingNumber { get; set; }
+        [JsonProperty("bank_account")]
+        public StripeBankAccount StripeBankAccount { get; set; }
 
         [JsonProperty("card")]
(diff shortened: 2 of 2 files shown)


brandur-stripe:
Thanks for the tests Remi! LGTM.

Any comment on how breaking these changes are? We're removing properties, but I guess none of them worked in the first place?

remi-stripe:
yeah I was ambivalent on the breaking change but they've been broken for years based on an old report. I guess it's fine to merge as is because it's self explanatory but wanted to quickly flag that it _does_ break something.

brandur-stripe:
> yeah I was ambivalent on the breaking change but they've been broken for years based on an old report. I guess it's fine to merge as is because it's self explanatory but wanted to quickly flag that it does break something.

Okay cool. If it makes sense to you, I'm thinking about bumping this as a major version just because those bad fields have been in place forever and I wouldn't be surprised if people were referencing them in their projects (even if they're not working correctly). What do you think?

jaymedavis:
If an object has an entity decorated with "bank_account", which has a child entity with a property "routing_number", that's processed the same as the object having a property called "bank_account[routing_number]". This code should work fine. I would recommend major.

remi-stripe:
> If an object has an entity decorated with "bank_account", which has a child entity with a property "routing_number", that's processed the same as the object having a property decorated with "bank_account[routing_number]". This code should work fine.

So that's what I thought too but it does not work today. I've tested it and dug for a while but could not figure it out. That's why I moved to a sub-object like Card which works fine.

My guess is that the nested logic was properly built on *parameters* but not on *properties*.

Do you have any intuition as to why it does not work?

jaymedavis:
A breakpoint [here](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Infrastructure/Public/Mapper.cs#L32) could verify that it's the serializer itself that doesn't like the prior format. All of the logic to build the parameters for sending to stripe is done by hand via plugins. The logic for deserializing the response is done via the Mapper class (which basically wraps json.net). It does surprise me that the nested class works and the properties don't... my best guess is the serializer doesn't like something.

Can you share an example of the full response?

remi-stripe:
> A breakpoint here could verify that it's the serializer itself that doesn't like the prior format. All of the logic to build the parameters for sending to stripe is done by hand via plugins.

Yeah though I work with the command line and unit tests so I haven't been able to figure out where the issue is happening. I dug for a bit but it is the only class with that approach and my guess is that when it was built it was assumed to work but did not and since the tests only looked at the id property and nothing else they passed. It has been reported as broken for a long time I think.

>  The logic for deserializing the response is done via the Mapper class (which basically wraps json.net). It does surprise me that the nested class works and the properties don't... my best guess is the serializer doesn't like something.

Yeah. Basically for plugins you built this notion of nested options which is how it's handled and I felt like for the Mapper class it was never built. I could be wrong though as I haven't dug enough into the internals of the serialization/deserialization yet but that's what it seemed like from skimming.

> Can you share an example of the full response?

You can see an example JSON response in the docs [here](https://stripe.com/docs/api#create_bank_account_token)

jaymedavis:
I checked the docs for the response, but didn’t see “bank_account” anywhere, so figured there was a little more to it. If that response is accurate, try taking the bank_account attribute off StripeBankAccount - I’d be curious to see if it works without that. Seems crazy, but it might work. :)

Did you ever workout how to console log messages with test?

Glad you’ve fixed this either way!

jaymedavis:
> Yeah. Basically for plugins you built this notion of nested options which is how it's handled and I felt like for the Mapper class it was never built. 

That’s correct. The plugin stuff is all for what is sent _to_ Stripe. The Mapper processes what is sent _from_ Stripe. Since the serializer seemed to work out of the box, I never had a reason to add anyrhing else. I thought there were a lot of other places that used similar declarations as this, but they may have been all refactored out over the years.

remi-stripe:
> I checked the docs for the response, but didn’t see “bank_account” anywhere, so figured there was a little more to it.

The docs I linked you to is specifically for the API method to create a bank account token. It shows an example response which has the information you needed. Here's a copy/paste of an example response:

```
{
  "id": "btok_1CPsqr2eZvKYlo2CycJM9ECU",
  "object": "token",
  "bank_account": {
    "id": "ba_1CPsqr2eZvKYlo2CReNfKjWw",
    "object": "bank_account",
    "account_holder_name": "Jane Austen",
    "account_holder_type": "individual",
    "bank_name": "STRIPE TEST BANK",
    "country": "US",
    "currency": "usd",
    "fingerprint": "1JWtPxqbdX5Gamtc",
    "last4": "6789",
    "routing_number": "110000000",
    "status": "new"
  },
  "client_ip": null,
  "created": 1525874465,
  "livemode": false,
  "type": "bank_account",
  "used": false
}
```

You can see that it has the `bank_account` hash the same way the card tokens have the `card` hash. And that's what the code assumed work but did not.

> Did you ever workout how to console log messages with test?

Yes! @brandur-stripe figured this one out a few weeks ago in https://github.com/stripe/stripe-dotnet/pull/1100 where it was missing some package in the project files.

jaymedavis:
Thanks. Yeah - that _should_ have worked as is, but it was probably refactored at some point. Since you mentioned this is the only place left that did not use a child object for the hash, it must have just been missed. Nice catch - this way is much cleaner imo.

Glad you got the console logging worked out @brandur-Stripe! 💪  If you need more debugging support in the future, visual studio code is a nice editor that allows you to set breakpoints as well. Command+shift+p shows everything you can do, much like sublime or atom.

remi-stripe:
Yeah I'll look into setting up VS at some point but since we work on all the libraries using the command line and a basic editor is usually faster in most cases!

cc @ob-stripe for the deserialization issue in case you are interested to investigate in the future.

brandur-stripe:
> Yeah agreed that major version is likely cleaner even though the code never worked before.

Excellent.

> Glad you got the console logging worked out @brandur-stripe! 💪 If you need more debugging support in the future, visual studio code is a nice editor that allows you to set breakpoints as well. Command+shift+p shows everything you can do, much like sublime or atom.

Yeah, thanks! I've had a few people around the office recently show me VS Code and I'm pretty sold, haha. I'll try this out next time I run into trouble.

Issue #1179, opened 2018-05-09
Add ability to set tax amount on Upcoming Invoice request


bcrobinson:
The stripe api has the ability to set a tax rate on the upcoming invoice request
https://stripe.com/docs/api#upcoming_invoice-subscription_tax_percent
But the `StripeUpcomingInvoiceOptions` class is missing this property.

I have a branch that has the fix on it but I am having trouble running the unit tests (https://github.com/bcrobinson/stripe-dotnet/tree/invoice_add_tax). If I can get the unit tests to run the I'll submit the pull request, but as you can see from the change it's rather trivial.

Are there any help pages on how to get the unit tests running? I've set the `STRIPE_TEST_SK` to our test account one but it seems like there's more configuration on that account that I'll need to do.

remi-stripe:
@bcrobinson If you're not sure how to run the unit tests you can also push your branch and our CI will run the tests automatically.

Just in case, here's how I run the tests (though I'm using the command line on OSX)

```
# For XUnit
STRIPE_TEST_SK=sk_test_.... dotnet test ./src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj /p:TargetFramework=netcoreapp1.1

# For MSpec
STRIPE_TEST_SK=sk_test_.... dotnet test ./src/Stripe.net.Tests/Stripe.net.Tests.csproj /p:TargetFramework=netcoreapp1.1
```

PR #1182, opened 2018-05-10
Added ability to set tax on upcoming invoice request

bcrobinson:
The stripe api has the ability to set a tax rate on the upcoming invoice request
https://stripe.com/docs/api#upcoming_invoice-subscription_tax_percent

This pull request adds this property to the StripeUpcomingInvoiceOptions class.

### DIFF
--- a/src/Stripe.net.Tests/invoices/when_getting_an_upcoming_invoice_with_options.cs
+++ b/src/Stripe.net.Tests/invoices/when_getting_an_upcoming_invoice_with_options.cs
@@ -27,6 +27,7 @@ namespace Stripe.Tests
             {
                 SubscriptionId = _stripeCustomer.Subscriptions.Data.First().Id,
                 SubscriptionPlanId = _stripePlan.Id,
+                SubscriptionTaxPercent = 17.5m,
             };
         };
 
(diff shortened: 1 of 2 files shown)


bcrobinson:
This is for issue https://github.com/stripe/stripe-dotnet/issues/1179

remi-stripe:
@bcrobinson Thank you so much for taking the time to submit a PR and adding a test on top of this.

I can see that another parameter is not supported yet (it was added recently). Would you mind also adding support for it: [subscription_trial_from_plan](https://stripe.com/docs/api/dotnet#upcoming_invoice-subscription_trial_from_plan). You don't need a test for that one since it's just a boolean.

remi-stripe:
Thank you so much for the quick work @bcrobinson 

r? @brandur-stripe 
cc @stripe/api-libraries

brandur-stripe:
LGTM. Thanks @bcrobinson @remi-stripe.

brandur-stripe:
Released as 16.1.0.

bcrobinson:
No problem. Do you have an ETA on when the next version of the nuget package will be released, or if there's a nightly package we can use?

brandur-stripe:
@bcrobinson Just as a quick follow up: there's a bit of a delay on the packages because they're pushed by our CI, so the build has to complete before they're available. That said, they're generally available within 30 minutes or so of when I announce one here.

PR #1184, opened 2018-05-13
Handle past time skew in timestamp validation

samuelkdavis:
* Check whether the absolute delta between server and timestamp is greater than the threshold

### DIFF
--- a/src/Stripe.net/Services/Events/StripeEventUtility.cs
+++ b/src/Stripe.net/Services/Events/StripeEventUtility.cs
@@ -47,7 +47,7 @@ namespace Stripe
 
             var webhookUtc = Convert.ToInt32(signatureItems["t"].FirstOrDefault());
 
-            if (utcNow - webhookUtc > tolerance)
+            if (Math.Abs(utcNow - webhookUtc) > tolerance)
                 throw new StripeException("The webhook cannot be processed because the current timestamp is above the allowed tolerance.");
 
             return Mapper<StripeEvent>.MapFromJson(json);


src/Stripe.net/Services/Events/StripeEventUtility.cs
```diff
@@ -47,7 +47,7 @@ public static StripeEvent ConstructEvent(string json, string stripeSignatureHead
 
             var webhookUtc = Convert.ToInt32(signatureItems["t"].FirstOrDefault());
 
-            if (utcNow - webhookUtc > tolerance)
+            if (Math.Abs(utcNow - webhookUtc) > tolerance)
                 throw new StripeException("The webhook cannot be processed because the current timestamp is above the allowed tolerance.");
```
remi-stripe: You'd meed to change the comment too to not just say "above"

remi-stripe:
@samuelkdavis Wanted to see if you'd be able to update your PR? Let me know if not!

jaymedavis:
If you are shooting at a scale of 0...5 > 5, this PR changes that scale effectively to -5...0...5 > 5. I think this is a precarious change.

remi-stripe:
@jaymedavis I disagree on that point but we will review carefully before the change is merged.

samuelkdavis:
Hey, was planning on giving this a proper look on the weekend - the change I'd made was directly through the github UI so I didn't go very deep.

@jaymedavis you are right to be cautious, it does allow for room for the payload to be up to 10 minutes off depending on the servers time skew. In it's current state however it only checks in a single direction, and essentially allows anything through if the time is skewed in the wrong direction.

The logic above is "within 5 minutes either side" which I believe is more accurate to the documentation Stripe supports.

I do have a pretty full weekend (checking out wedding venues) so if I'm not respondent enough feel free to take this from me.

PR #1189, opened 2018-05-21
Add support for SEPA Credit Transfer sources

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

The tests work on the main account as it is gated in the feature. It would fail for other accounts as this feature is still in private beta today.


brandur-stripe:
Thanks again Remi — I left one comment on the internal ticket (DX-1398), but otherwise LGTM.

brandur-stripe:
Released as 16.2.0.

Issue #1190, opened 2018-05-21
BankAccountOptions is not contained in a namespace.


Kurt-Rayner-Alpha:
The following file defines the class BankAccountOptions at the root level.  It shows up outside of the Stripe namespace, polluting the global namespace for the assembly.

Note:  There appears to be another class that IS in the namespace.

  stripe-dotnet\src\Stripe.net\Services\_sources_token\BankAccountOptions.cs

```
Contents follow:

using Newtonsoft.Json;
using Stripe;

public class BankAccountOptions : INestedOptions
{
    [JsonProperty("bank_account")]
    public string TokenId { get; set; }

    [JsonProperty("bank_account[account_holder_name]")]
    public string AccountHolderName { get; set; }

    [JsonProperty("bank_account[account_holder_type]")]
    public string AccountHolderType { get; set; }

    [JsonProperty("bank_account[account_number]")]
    public string AccountNumber { get; set; }

    [JsonProperty("bank_account[country]")]
    public string Country { get; set; }

    [JsonProperty("bank_account[currency]")]
    public string Currency { get; set; }

    [JsonProperty("bank_account[routing_number]")]
    public string RoutingNumber { get; set; }
}
```

remi-stripe:
Thanks for the report, I opened a PR to fix this!

PR #1191, opened 2018-05-21
Move BankAccountOptions to the Stripe namespace

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/1190

r? @brandur-stripe 
cc @stripe/api-libraries 

Do you think it's a breaking change? I don't know much about namespaces in .NET. If not, let's leave as is, if it is we likely should rename the class to start with `Stripe` to be cleaner.

### DIFF
--- a/src/Stripe.net/Services/_sources_token/BankAccountOptions.cs
+++ b/src/Stripe.net/Services/_sources_token/BankAccountOptions.cs
@@ -1,26 +1,28 @@
 ﻿using Newtonsoft.Json;
-using Stripe;
 
-public class BankAccountOptions : INestedOptions
+namespace Stripe
 {
-    [JsonProperty("bank_account")]
-    public string TokenId { get; set; }
+    public class BankAccountOptions : INestedOptions
+    {
+        [JsonProperty("bank_account")]
+        public string TokenId { get; set; }
 
-    [JsonProperty("bank_account[account_holder_name]")]
-    public string AccountHolderName { get; set; }
+        [JsonProperty("bank_account[account_holder_name]")]
+        public string AccountHolderName { get; set; }
 
-    [JsonProperty("bank_account[account_holder_type]")]
-    public string AccountHolderType { get; set; }
+        [JsonProperty("bank_account[account_holder_type]")]
+        public string AccountHolderType { get; set; }
 
-    [JsonProperty("bank_account[account_number]")]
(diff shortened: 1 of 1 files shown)


jaymedavis:
@remi-stripe - this sort of [came up recently](https://github.com/stripe/stripe-dotnet/pull/1176/files) for tokens. You can probably just stick a `StripeBankAccount` on this one too.

remi-stripe:
@jaymedavis I don't think that's correct. `StripeBankAccount` is for deserialization of a resource not for a parameter. They might share some properties but not all. The resource would not have the bank account number and the parameter would not have the last 4. They have to be separate resources.

jaymedavis:
I was editing my response out, but you are too quick! You are correct!

remi-stripe:
Glad we're on the same page! But I agree that there is likely a way to share some of those options around as they are re-used in various cases and would likely benefit from some cleanup!

jaymedavis:
I did something similar [here](https://github.com/stripe/stripe-dotnet/tree/master/src/Stripe.net/Services/_sources) at one point. Although, I'm not sure everywhere that accepts a bank_account hash could also accept a source hash? If not, it can at least serve as an example that it could work! 😄

remi-stripe:
Yeah the idea is good but that's unfortunately a different approach/logic which would not work in the Token case!

brandur-stripe:
Ouch — I guess this is purely an oversight in that it was accidentally injected into the wrong namespace? I'm a little surprised that you didn't have to update any references to the class, but it seems to work, I guess by virtue of the existing set of `using` statements we have.

> Do you think it's a breaking change? I don't know much about namespaces in .NET. If not, let's leave as is, if it is we likely should rename the class to start with Stripe to be cleaner.

Yeah, interesting. I think it technically _is_ breaking, but it's kind of a weird case. Just like you didn't have to update any references to the new namespace, I suspect that a lot of user code is the same because you can reasonably expect that they'll already have a `using Stripe` in pretty much every file that they would have been referencing `BankAccountOptions`. My feeling would be that we make an extra note in the changelog about it, but ship it as a minor version bump. Thoughts?

remi-stripe:
@brandur-stripe that sounds good to me, we can release a minor version I think

brandur-stripe:
Added an extra note in the CHANGELOG and released as 16.3.0.

PR #1192, opened 2018-05-23
Disable Topups tests as it breaks the Test suite because it's brittle

remi-stripe:
Topups creation requires a Source to be chargeable. In Test mode this can take a random time though which leads to sporadic/random build failures.

For now we disable the tests as the resource works already and is still in private beta anyway.

Fixes https://github.com/stripe/stripe-dotnet/pull/1110#issuecomment-391351071

r? @brandur-stripe 
cc @stripe/api-libraries


remi-stripe:
ah neat, I did not know about that feature but I confirmed the tests are skipped locally! 

PTAL

remi-stripe:
ah no I lied. This still creates a topup source in the fixtures I think.

remi-stripe:
Yeah I confirmed in my logs, every time I run that test suite it still creates the topup (or tries) so the issue is still here.

Do you have another idea?

brandur-stripe:
> ah no I lied. This still creates a topup source in the fixtures I think.

Ah, dang. Maybe just minimize the commenting then by just commenting out the contents of `public topups_fixture()` in `src/Stripe.Tests.XUnit/topups/_fixture.cs`? I guess that's still a little better.

remi-stripe:
Okay, done by commenting the minimum amount of code in the fixture. PTAL

brandur-stripe:
LGTM. Thanks.

PR #1199, opened 2018-06-06
add invoice's pdf and url

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/StripeInvoice.cs
+++ b/src/Stripe.net/Entities/StripeInvoice.cs
@@ -92,6 +92,12 @@ namespace Stripe
         [JsonProperty("forgiven")]
         public bool? Forgiven { get; set; }
 
+        [JsonProperty("hosted_invoice_url")]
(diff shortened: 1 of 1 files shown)


src/Stripe.net/Entities/StripeInvoice.cs
```diff
@@ -92,6 +92,12 @@ internal object InternalCustomer
         [JsonProperty("forgiven")]
         public bool? Forgiven { get; set; }
 
+        [JsonProperty("hosted_invoice_url")]
+        public string HostedInvoiceUrl { get; set; }
+
+        [JsonProperty("invoice_pdf")]
+        public string InvoicePDF { get; set; }
```
brandur-stripe: I think the convention would be `InvoicePdf` based off of uses of `Url` and `Id` elsewhere.

brandur-stripe:
Thanks! Released as 16.6.0.

PR #1207, opened 2018-06-13
Add extra properties to the Invoice object

remi-stripe:
Add support for `amount_remaining` and `billing_reason` to the Invoice object.

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/StripeInvoice.cs
+++ b/src/Stripe.net/Entities/StripeInvoice.cs
@@ -16,6 +16,9 @@ namespace Stripe
         [JsonProperty("amount_paid")]
         public int AmountPaid { get; set; }
 
+        [JsonProperty("amount_remaining")]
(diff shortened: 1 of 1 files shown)


src/Stripe.net/Entities/StripeInvoice.cs
```diff
@@ -31,6 +34,8 @@ public class StripeInvoice : StripeEntityWithId, ISupportMetadata
         [JsonProperty("billing")]
         public StripeBilling? Billing { get; set; }
 
+        public string BillingReason { get; set; }
```
remi-stripe: minor but I almost created an Enum and then realized there are like 2 files today. We should discuss the benefits of Enums and whether it's worth porting all the properties to Enums (which I would happily do if so)

ob-stripe:
Released in 16.11.0.

PR #1210, opened 2018-06-15
Add the 'customer.source.expiring' event and fix a typo

karlr-stripe:
Two changes here : 

1. Adding a `CustomerSourceExpiring` StripeEvent for https://stripe.com/docs/api#event_types-customer.source.expiring
2. Updating `CustomerSourcedDeleted` -> `CustomerSourceDeleted`

I can back out the second change if we're worried about backwards compatibility or anything.

### DIFF
--- a/src/Stripe.net/Constants/StripeEvents.cs
+++ b/src/Stripe.net/Constants/StripeEvents.cs
@@ -182,7 +182,12 @@ namespace Stripe
         /// <summary>
         /// Occurs whenever a source is removed from a customer.
         /// </summary>
-        public const string CustomerSourcedDeleted = "customer.source.deleted";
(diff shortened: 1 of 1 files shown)


remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 

@karlr-stripe Thank you for the fix! Since we rename an event property it's likely a breaking change though.

adamtay:
Hi, any updates on adding the `customer.source.deleted` event to the API?

remi-stripe:
@adamtay The event is in the library, it's just with a typo but you can use it. We're going to try to merge this one in the next few days

ob-stripe:
Released in 17.0.0.

PR #1211, opened 2018-06-16
Add name to the Coupon resource

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/StripeCoupon.cs
+++ b/src/Stripe.net/Entities/StripeCoupon.cs
@@ -35,6 +35,9 @@ namespace Stripe
         [JsonProperty("metadata")]
         public Dictionary<string, string> Metadata { get; set; }
 
+        [JsonProperty("name")]
+        public string Name { get; set; }
+
         [JsonProperty("percent_off")]
         public int? PercentOff { get; set; }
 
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released in 16.11.0.

winzig:
How does one update an existing coupon with the Name field? 

Is this Name field intended to be shown in the receipts/invoices instead of the coupon ID code? If so, it doesn't appear to be working at this time. At least in my test dashboard, the Name only shows up when creating the invoice. When I get the actual invoice/receipt email, it still shows the "ID code" for the coupon instead.

Thanks!

remi-stripe:
@winzig This can not be updated today, it can only be set at creation. For the second part, it could be a bug but it's unrelated to the library. I would recommend contacting our support team to discuss this: https://support.stripe.com/email

winzig:
@remi-stripe That's great news, I've filed a support request about the possible bug. 

In terms of the update capability, is this planned soon? I'm a bit squeamish about trying to create a new coupon and replacing the existing coupons on our existing subscriptions. Being able to just update the name param on existing coupons would be highly desirable.

remi-stripe:
@winzig Not something we have planned for now but I will take this to the team and hopefully we can add this in the future

remi-stripe:
@winzig  Issues are specific to bugs with the libraries so I would recommend against it. Since you reporting it to our support team (and I did too with the eng team) for now the only thing to do is wait for us to implement it. I will add it to the libraries once support for it is added in the API!

PR #1213, opened 2018-06-20
Add missing parameter `aggregate_usage`

svedi:
I (hopefully) added support for the parameter aggregate_usage when creating a plan and the parameter tax_percent when creating a subscription.

From the example at https://stripe.com/docs/billing/subscriptions/taxes it seems that tax_percent should be a decimal. I hope this is correct, since I lack any big picture overview of the API.

Finally, I might add that I have not run the test suit since I do most development on linux, and the test suit did not seem to compile.

### DIFF
--- a/src/Stripe.net/Services/Plans/StripePlanCreateOptions.cs
+++ b/src/Stripe.net/Services/Plans/StripePlanCreateOptions.cs
@@ -11,6 +11,9 @@ namespace Stripe
         [JsonProperty("active")]
         public bool? Active { get; set; }
 
+        [JsonProperty("aggregate_usage")]
+        public string AggregateUsage { get; set; }
+
         [JsonProperty("amount")]
         public int? Amount { get; set; }
 


src/Stripe.net/Services/Subscriptions/StripeSubscriptionCreateOptions.cs
```diff
@@ -37,6 +37,9 @@ public class StripeSubscriptionCreateOptions : SubscriptionSharedOptions
         [JsonProperty("subscription_items_array")]
         public List<StripeSubscriptionItemOption> Items { get; set; }
 
+        [JsonProperty("tax_percent")]
```
remi-stripe: The `StripeSubscriptionCreateOptions` class inherits from `SubscriptionSharedOptions` which already defines `TaxPercent` as you can see here: https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/Subscriptions/SubscriptionSharedOptions.cs#L63

svedi:
Right, fair enough. Not sure how I missed that one. I thought I tried copy pasting it, but maybe I misspelled it or something.

I have updated master and removed the TaxPercent commit.

remi-stripe:
@ob-stripe Can you merged this one in?

ob-stripe:
Released as 16.12.0.

Thanks for the contribution @svedi! Much appreciated :)

PR #1216, opened 2018-06-24
Add support for passing level III data when creating a charge

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 
cc @fred-stripe

### DIFF
--- a/src/Stripe.net/Infrastructure/Middleware/ParserPlugin/ChargeLevel3Plugin.cs
+++ b/src/Stripe.net/Infrastructure/Middleware/ParserPlugin/ChargeLevel3Plugin.cs
@@ -1,0 +1,40 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Newtonsoft.Json;
+
+namespace Stripe.Infrastructure.Middleware
+{
+    internal class ChargeLevel3Plugin : IParserPlugin
+    {
+        public bool Parse(ref string requestString, JsonPropertyAttribute attribute, PropertyInfo property, object propertyValue, object propertyParent)
+        {
+            if (attribute.PropertyName != "charge_level3_line_items_array") return false;
+
+            var items = ((List<StripeChargeLevel3LineItemOptions>) property.GetValue(propertyParent, null));
+
+            var itemIndex = 0;
+            foreach (var item in items)
+            {
+                var properties = item.GetType().GetRuntimeProperties();
+
+                foreach (var prop in properties)
+                {
+                    var value = prop.GetValue(item, null);
+                    if (value == null) continue;
+
+                    // it must have a json attribute matching stripe's key, and only one
+                    var attr = prop.GetCustomAttributes<JsonPropertyAttribute>().SingleOrDefault();
+                    if (attr == null) continue;
+
+                    RequestStringBuilder.ApplyParameterToRequestString(ref requestString,
+                        $"level3[line_items][{itemIndex}][{attr.PropertyName}]", value.ToString());
+                }
+
+                itemIndex++;
+            }
+
+            return true;
+        }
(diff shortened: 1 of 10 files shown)


src/Stripe.net/Infrastructure/Middleware/ParserPlugin/ChargeLevel3Plugin.cs
```diff
@@ -0,0 +1,40 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Newtonsoft.Json;
+
+namespace Stripe.Infrastructure.Middleware
+{
+    internal class ChargeLevel3Plugin : IParserPlugin
```
remi-stripe: We really should revisit how plugins work. It's all copy/pasta for the sake of adding one list of items. There has to be a better way

src/Stripe.net/Services/Charges/StripeChargeLevel3LineItemOptions.cs
```diff
@@ -0,0 +1,25 @@
+using Newtonsoft.Json;
+
+namespace Stripe
+{
+    public class StripeChargeLevel3LineItemOptions : INestedOptions
```
remi-stripe: I still don't understand when something should be called Options versus Option.

src/Stripe.net/Services/Charges/StripeChargeLevel3Options.cs
```diff
@@ -0,0 +1,25 @@
+using Newtonsoft.Json;
+using System.Collections.Generic;
+
+namespace Stripe
+{
+    public class StripeChargeLevel3Options : INestedOptions
+    {
+        [JsonProperty("level3[merchant_reference]")]
```
remi-stripe: This works but since the parent has `JsonProperty(level3)` why isn't it passed to the lower level and you have to repeat it? The same happened to shipping.

src/Stripe.net/Services/Charges/StripeChargeLevel3Options.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        [JsonProperty("level3[shipping_amount]")]
+        public int ShippingAmount { get; set; }
+
+        /// <summary>
+        /// List of line items items to pass as level III data when creating the charge
+        /// </summary>
+        // this will actually send `line_items`. this is to flag it for the middleware
+        // to process it as a plugin
+        [JsonProperty("charge_level3_line_items_array")]
```
remi-stripe: This hack makes me sad each time I have to use it (and grep to find it again).

remi-stripe:
changed to [WIP] as this requires changes to the Charge ressource too but those are behind a gate

remi-stripe:
Okay pushed a new version that can handle the returned `level3` hash on the Charge object. This is gated though so I could not push tests for it but I did test locally.

src/Stripe.net/Infrastructure/Middleware/ParserPlugin/ChargeLevel3Plugin.cs
ob-stripe: Agreed. I took a stab at this in #1218.

src/Stripe.net/Services/Charges/StripeChargeLevel3LineItemOptions.cs
ob-stripe: I think it should always be `Options`.

remi-stripe: Yeah that would work. I thought that we started using `Option` for nested ones to differentiate them from the top-level ones.

PR #1218, opened 2018-06-25
Generic plugin for encoding lists

ob-stripe:
This is an attempt at replacing all the custom plugins for encoding lists of hashes by a single generic plugin that will automatically encode all `List`s as integer-indexed maps.

I think this makes sense because this is what Stripe's API expects in most cases. For regular lists, we can just arrays instead (e.g. `string[]` vs. `List<string>`) and the existing `ArrayPlugin` will encode those as regular lists.

To clarify:
- arrays (e.g. `string[]`) are URL-encoded as lists: `foo[]=val1&foo[]=val2`
- lists (e.g. `List<string>` are URL-encoded as integer-indexed maps: `foo[0]=val1&foo[1]=val2`

~Tagged as WIP for now as I broke some additional custom encoding that was handled by the previous plugins (e.g. `up_to` special values for plan's `tiers`)~


ob-stripe:
Fixed the `up_to` encoding. Looks like everything's okay now. This was simpler than I expected!

cc @stripe/api-libraries

src/Stripe.net/Infrastructure/Middleware/ParserPlugin/ListPlugin.cs
```diff
@@ -27,8 +36,9 @@ public bool Parse(ref string requestString, JsonPropertyAttribute attribute, Pro
                     // it must have a json attribute matching stripe's key, and only one
                     var attr = prop.GetCustomAttributes<JsonPropertyAttribute>().SingleOrDefault();
                     if (attr == null) continue;
+
                     // a JsonPropertyAttribute is required to supply a property name to parser plugins.
-                    var indexedItemAttribute = new JsonPropertyAttribute($"items[{itemIndex}][{attr.PropertyName}]");
+                    var indexedItemAttribute = new JsonPropertyAttribute($"{attribute.PropertyName}[{itemIndex}][{attr.PropertyName}]");
```
remi-stripe: minor but `attribute` versus `attr` in this line is a bit confusing/hard to read

ob-stripe:
> The description of the PR should clarify that for list it's not foo[0]=val1 but for[0][attr1]=val1

Well, that's only true for a `List<Something>` where `Something` is a class with an attribute `attr1`. My example is valid for `List`s of primitives e.g. `List<String>`. (Of course in practice all of `List`s in requests are for option classes and not for primitive data types, except for `Expand` in `StripeBaseOptions` which is encoded separately.)

src/Stripe.net/Infrastructure/Middleware/ParserPlugin/ListPlugin.cs
ob-stripe: 👍  I copied/pasted from the previous plugins but we should take this opportunity to clarify the code. I'll update with better variable names.

remi-stripe:
I see. And do you plan to enforce using an Array instead of a list for base types? If not, isn't your code not working since you explicitly for `foo[][attr1]` in the code?

ob-stripe:
> I see. And do you plan to enforce using an Array instead of a list for base types? If not, isn't your code not working since you explicitly for foo[][attr1] in the code?

🤦‍♂️  Yes, absolutely. I updated it handle both cases, and added some tests in `encoding_parameters` and `encoding_lists_of_objects`.

remi-stripe:
Thank you for adding tests! I've noticed lately that we are not "consistent" in the way we pass nested options and define their JSON name in the API. This is likely unrelated but I wanted to raise it just in case it ends up related.

In some cases we "inherit" the parent element from the parent class but in others we explicitly repeat the parent json name. While typing this, I realize this is likely unclear to anyone but me.

For example my recent [level3 changes](https://github.com/stripe/stripe-dotnet/pull/1216) have the [property](https://github.com/stripe/stripe-dotnet/pull/1216/files#diff-aacd6baaf2b13a3b4a76eb477ddcfb73R57) in the top level options class with `[JsonProperty("level3")]`. 
The [parameters](https://github.com/stripe/stripe-dotnet/pull/1216/files#diff-18e3cb471edfc7b83f33b9deadbf9837R8) though have `[JsonProperty("level3[customer_reference]")]`.

ob-stripe:
> In some cases we "inherit" the parent element from the parent class but in others we explicitly repeat the parent json name. While typing this, I realize this is likely unclear to anyone but me.

I don't think it's clear to anyone tbh 😐 

So from what I understand, if the nested options class is directly used, then it includes the parent parameter's name. However, if the class is used in a `List`, then it doesn't (because the plugins take care of computing the URL-encoded parameter name).

Also, not all nested options class implement `INestedOptions` :/

We should definitely clean this up. Nested options class should not have to know about the parent parameter's name (it makes it impossible to reuse the same class for two different parameters).

remi-stripe:
Yep totally. I was flagging to make sure we were not missing anything in the changes you made since it would apply widely to all options class. It seems to clean up something messy though and it does seem correct to me.

remi-stripe:
@ob-stripe I rebased on the most recent master and fixed the Level III part to move to a plugin. I think this one is ready to be merged.

ob-stripe:
Released in 16.16.1.

PR #1219, opened 2018-06-27
Add support for updating a coupon's name

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 

This was just shipped [in the API](https://stripe.com/docs/api#update_coupon-name) and I think it's the only library that needs this since stripe-go has a unique params structure for creation/update/delete.


ob-stripe:
Released in 16.13.0.

PR #1221, opened 2018-07-03
Add support for expanding outcome.rule on a charge

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries @spastorelli-stripe @gh-stripe 

This ensures that a Charge can be deserialized if `outcome[rule]` is already expanded. Before this we only supported a string for this field preventing older API versions from expanding this field or deserializing a `charge.*` event on older API versions.

I don't think we can reliably add tests for this one since it's Radar related so I punted.


ob-stripe:
Released in 16.15.0.

PR #1222, opened 2018-07-04
Add support for the PaymentIntent resource

remi-stripe:
This PR adds support for the Payment Intent resource.

cc @stripe/api-libraries @michelle-stripe @jenan-stripe 
cc @stevene-stripe

### DIFF
--- a/src/Stripe.net/Entities/PaymentIntents/PaymentIntentSourceAction.cs
+++ b/src/Stripe.net/Entities/PaymentIntents/PaymentIntentSourceAction.cs
@@ -1,0 +1,20 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public enum PaymentIntentSourceActionType
+    {
+        Unknown,
+        AuthorizeWithUrl,
+        None,
+    }
+
+    [JsonConverter(typeof(PaymentIntentSourceActionConverter))]
+    public class PaymentIntentSourceAction : StripeEntity
+    {
+        public PaymentIntentSourceActionType Type { get; set; }
+
+        public PaymentIntentSourceActionAuthorizeWithUrl AuthorizeWithUrl { get; set; }
+    }
+}
--- a/src/Stripe.net/Services/PaymentIntents/PaymentIntentConfirmOptions.cs
+++ b/src/Stripe.net/Services/PaymentIntents/PaymentIntentConfirmOptions.cs
@@ -1,0 +1,26 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class PaymentIntentConfirmOptions : StripeBaseOptions
+    {
+        [JsonProperty("invoice")]
+        public string Invoice { get; set; }
+
+        [JsonProperty("invoice_charge_reason")]
+        public string InvoiceChargeReason { get; set; }
+
+        [JsonProperty("receipt_email")]
+        public string ReceiptEmail { get; set; }
+
+        [JsonProperty("return_url")]
+        public string ReturnUrl { get; set; }
+
+        [JsonProperty("save_source_to_customer")]
+        public bool? SaveSourceToCustomer { get; set; }
+
+        [JsonProperty("source")]
+        public string SourceId { get; set; }
+    }
+}
--- a/src/Stripe.net/Entities/PaymentIntents/PaymentIntent.cs
+++ b/src/Stripe.net/Entities/PaymentIntents/PaymentIntent.cs
@@ -1,0 +1,148 @@
+namespace Stripe
+{
+    using System;
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class PaymentIntent : StripeEntityWithId, ISupportMetadata
+    {
+        [JsonProperty("object")]
+        public string Object { get; set; }
+
+        [JsonProperty("allowed_source_types")]
+        public List<string> AllowedSourceTypes { get; set; }
+
+        [JsonProperty("amount")]
+        public int? Amount { get; set; }
+
+        [JsonProperty("amount_capturable")]
+        public int? AmountCapturable { get; set; }
+
+        [JsonProperty("amount_received")]
+        public int? AmountReceived { get; set; }
+
+        #region Expandable Application
+        public string ApplicationId { get; set; }
+
+        [JsonIgnore]
+        public StripeApplication Application { get; set; }
+
+        [JsonProperty("application")]
+        internal object InternalApplication
+        {
+            set
+            {
+                StringOrObject<StripeApplication>.Map(value, s => this.ApplicationId = s, o => this.Application = o);
+            }
+        }
+        #endregion
+
+        [JsonProperty("application_fee_amount")]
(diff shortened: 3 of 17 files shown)


src/Stripe.Tests.XUnit/payment_intents/creating_and_updating_payment_intents.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        [Fact]
+        public void created_has_the_right_information()
+        {
+            fixture.PaymentIntent.Should().NotBeNull();
+            fixture.PaymentIntent.Amount.Should().Be(fixture.PaymentIntentCreateOptions.Amount);
+            fixture.PaymentIntent.CaptureMethod.Should().Be(fixture.PaymentIntentCreateOptions.CaptureMethod);
+            fixture.PaymentIntent.Currency.Should().Be(fixture.PaymentIntentCreateOptions.Currency);
+            fixture.PaymentIntent.AllowedSourceTypes.Count.Should().Be(fixture.PaymentIntentCreateOptions.AllowedSourceTypes.Count);
```
remi-stripe: I could not, for the life of me, figure out how to compare the 2 lists. I tried a lot of options but none worked/compiled for some reasons so I settled for the count for now.

src/Stripe.net/Infrastructure/Middleware/ParserPlugin/PaymentIntentAllowedSourceTypesPlugin.cs
```diff
@@ -0,0 +1,28 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Newtonsoft.Json;
+
+namespace Stripe.Infrastructure.Middleware
+{
+    internal class PaymentIntentAllowedSourceTypesPlugin : IParserPlugin
```
remi-stripe: This one is dirty/sad but it's just temporary because I needed to make it work. We should ship only after @ob-stripe other PR https://github.com/stripe/stripe-dotnet/pull/1218 instead.

src/Stripe.net/Services/PaymentIntents/StripePaymentIntentConfirmOptions.cs
```diff
@@ -0,0 +1,17 @@
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace Stripe
+{
+    public class StripePaymentIntentConfirmOptions : StripeBaseOptions
+    {
```
remi-stripe: Those are a bit frustrating, they each support a subset of the other endpoints params. Not sure this was the best approach

src/Stripe.net/Services/PaymentIntents/StripePaymentIntentTransferDataOptions.cs
```diff
@@ -0,0 +1,10 @@
+using Newtonsoft.Json;
+
+namespace Stripe
+{
+    public class StripePaymentIntentTransferDataOptions : INestedOptions
+    {
+        [JsonProperty("transfer_data[amount]")]
```
remi-stripe: I'll never understand why we need to repeat the parent property's parameter name. Haven't tested either as the test suite would require a custom account and all that.

src/Stripe.net/Entities/PaymentIntents/StripePaymentIntent.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        [JsonProperty("description")]
+        public string Description { get; set; }
+
+        [JsonProperty("livemode")]
+        public bool LiveMode { get; set; }
+
+        [JsonProperty("metadata")]
+        public Dictionary<string, string> Metadata { get; set; }
+
```
remi-stripe: missing support for `next_source_action`

src/Stripe.net/Entities/PaymentIntents/StripePaymentIntent.cs
```diff
@@ -0,0 +1,145 @@
+using System;
```
remi-stripe: Can you confirm that the folder structure is the right approach?

src/Stripe.net/Entities/PaymentIntents/StripePaymentIntent.cs
remi-stripe: Done, though the code feels really ugly 😢

ob-stripe:
@remi-stripe Looks like this one also needs to be rebased and fixed.

remi-stripe:
Yes, we should hold off a bit on that one since the API could be changing in the next few days. I'll rebase once we know the exact/final shape

src/Stripe.net/Entities/PaymentIntents/PaymentIntentSourceAction.cs
```diff
@@ -0,0 +1,20 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public enum PaymentIntentSourceActionType
+    {
+        AuthorizeWithUrl,
+        None,
+        Unknown,
```
ob-stripe: Can we have `Unknown` as the first value in the enum, so that when you initialize a new `PaymentIntentSourceActionType`, the `Type` field defaults to `Unknown`?

src/Stripe.net/Services/PaymentIntents/PaymentIntentConfirmOptions.cs
```diff
@@ -0,0 +1,26 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class PaymentIntentConfirmOptions : StripeBaseOptions
+    {
```
ob-stripe: According to the OpenAPI spec, this request also has an additional `client_secret` string parameter.

src/Stripe.net/Entities/PaymentIntents/PaymentIntentSourceAction.cs
remi-stripe: Hmmm sure but that breaks the "alphabetical order" rule :)

src/Stripe.net/Services/PaymentIntents/PaymentIntentConfirmOptions.cs
remi-stripe: That's a client-side property used with the publishable key, I'm not sure this is a useful parameter in the server-side library. Do you still think I should add it? It's not in our docs.

src/Stripe.net/Entities/PaymentIntents/PaymentIntentSourceAction.cs
ob-stripe: That's fine. I don't think it makes sense for enums anyway -- if we add a new value to an existing enum, we should append it after all existing values, otherwise it's technically a breaking change.

src/Stripe.net/Services/PaymentIntents/PaymentIntentConfirmOptions.cs
ob-stripe: Ack, let's ignore it for now -- we can always append it later if needed.

remi-stripe:
@ob-stripe PTAL

PR #1223, opened 2018-07-06
Add missing property/params to the InvoiceItem resource

remi-stripe:
* Add unit_amount everywhere
* Add quantity on create and update
* Remove currency on update as it's not valid

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Services/InvoiceItems/StripeInvoiceItemUpdateOptions.cs
+++ b/src/Stripe.net/Services/InvoiceItems/StripeInvoiceItemUpdateOptions.cs
@@ -11,13 +11,16 @@ namespace Stripe
         [JsonProperty("description")]
         public string Description { get; set; }
 
-        [JsonProperty("currency")]
-        public string Currency { get; set; }
-
         [JsonProperty("discountable")]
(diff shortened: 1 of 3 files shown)


src/Stripe.net/Services/InvoiceItems/StripeInvoiceItemUpdateOptions.cs
```diff
@@ -11,13 +11,16 @@ public class StripeInvoiceItemUpdateOptions : StripeBaseOptions, ISupportMetadat
         [JsonProperty("description")]
         public string Description { get; set; }
 
-        [JsonProperty("currency")]
```
remi-stripe: This is not valid per https://stripe.com/docs/api#update_invoiceitem which means it never worked so not sure if a breaking change

ob-stripe: If it never worked, then it's not a breaking change IMO.

ob-stripe:
Released in 16.16.0.

PR #1224, opened 2018-07-11
Improve tolerance verification to check both past and future timestamps

### LINKED ISSUE #1183: Signature time validation does not handle server time skew in the past
The Stripe signature docs state "Our libraries have a default tolerance of five minutes between the timestamp and the current time" - [source](https://stripe.com/docs/webhooks/signatures#replay-attacks
)

However the logic around this is ([source](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/Events/StripeEventUtility.cs#L50)):

`if (serverTime - webhookTimestamp > tolerance)`

If the servers time drifts into the future, this will correctly throw an error. If the server time lags into the past however, the time validation becomes significantly less useful.

To fit the definition, absolute value looks like it'd solve this

`if (Math.Absolute(serverTime - webhookTimestamp) > tolerance)`

Stripe version: current master, 16.1.0

remi-stripe:
This increases the real tolerance to 10 minutes instead of just 5 but it is what we do in the other libraries. This is a fix for https://github.com/stripe/stripe-dotnet/pull/1184 which went stale.

Fixes #1183.

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Services/Events/StripeEventUtility.cs
+++ b/src/Stripe.net/Services/Events/StripeEventUtility.cs
@@ -47,8 +47,8 @@ namespace Stripe
 
             var webhookUtc = Convert.ToInt32(signatureItems["t"].FirstOrDefault());
 
-            if (utcNow - webhookUtc > tolerance)
-                throw new StripeException("The webhook cannot be processed because the current timestamp is above the allowed tolerance.");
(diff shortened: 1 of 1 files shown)


ob-stripe:
Released in 17.0.0.

PR #1226, opened 2018-07-11
[WIP] Move part of the test suite to stripe-mock

remi-stripe:
This is an attempt to move part of the test suite to stripe-mock for the resource that it handles without issues.

This uses a trick which configures stripe-mock for that specific test and then goes back to the normal API endpoint for other tests. Thanks to @ob-stripe for the idea for a better approach than just rewriting  the entire test suite at once.

cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.Tests.XUnit/coupons_new/coupons.cs
+++ b/src/Stripe.Tests.XUnit/coupons_new/coupons.cs
@@ -1,0 +1,116 @@
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using Xunit;
+
+namespace Stripe.Tests.Xunit
+{
+    public class coupon_test
+    {
+        private string previousApiBase;
+        private readonly string resourceId = "co_123";
+        private StripeCouponService service;
+
+        public coupon_test()
+        {
+            this.service = new StripeCouponService(Cache.ApiKey);
+            this.previousApiBase = StripeConfiguration.GetApiBase();
+            StripeConfiguration.SetApiBase("http://localhost:12111/v1");
+        }
+
+        ~coupon_test()
+        {
+            StripeConfiguration.SetApiBase(this.previousApiBase);
+        }
+
+        [Fact]
+        public void create_succeeds()
+        {
+            var options = new StripeCouponCreateOptions() {
+                // Add a space at the end to ensure the ID is properly URL encoded
+                // when passed in the URL for other methods
+                Id = "co_123 ",
+                PercentOff = 25,
+                Duration = "repeating",
+                DurationInMonths = 3,
+            };
+
+            var coupon = service.Create(options);
+
+            Assert.NotNull(coupon);
+            coupon.Id.Should().Be(options.Id);
+            coupon.PercentOff.Should().Be(options.PercentOff);
+            coupon.Duration.Should().Be(options.Duration);
+            coupon.DurationInMonths.Should().Be(options.DurationInMonths);
+        }
+
+        [Fact]
+        public void update_succeeds()
+        {
+            var options = new StripeCouponUpdateOptions {
+              Metadata = new Dictionary<string, string>{{"key_1", "value_1"}}
+            };
+
+            var coupon = service.Update(resourceId, options);
+
+            Assert.NotNull(coupon);
+            coupon.Id.Should().Be(resourceId);
+            // stripe-mock does not update metadata so we can't verify this part
+        }
+
+        [Fact]
+        public void get_succeeds()
+        {
+            var coupon = service.Get(resourceId);
+
+            Assert.NotNull(coupon);
+            coupon.Id.Should().Be(resourceId);
+        }
+
+        [Fact]
+        public void delete_succeeds()
+        {
+            var coupon = service.Delete(resourceId);
+
+            Assert.NotNull(coupon);
+            coupon.Deleted.Should().BeTrue();
+        }
+
+        [Fact]
+        public void list_succeeds()
+        {
+            var coupons = service.List();
+
+            Assert.NotNull(coupons);
+
+            var count = 0;
+            IEnumerable<StripeCoupon> enumerable = coupons as IEnumerable<StripeCoupon>;
+            foreach (var obj in enumerable)
+            {
+                count += 1;
+            }
+            Assert.Equal(coupons.ToList().Count > 0, true);
+            Assert.Equal(coupons.ToList().Count, count);
+
+            var datahash = new HashSet<string>();
+            foreach (var obj in coupons.Data)
+            {
+                datahash.Add(obj.Id);
+            }
+
+            var enumhash = new HashSet<string>();
+            IEnumerable<StripeCoupon> enumerable2 = coupons as IEnumerable<StripeCoupon>;
+            foreach (var obj in enumerable2)
+            {
+                enumhash.Add(obj.Id);
+            }
+
+            Assert.Equal(datahash, enumhash);
+
+            Assert.NotNull(coupons.Object);
+            Assert.Equal(coupons.Object, "list");
+            Assert.NotNull(coupons.Data);
+            Assert.NotNull(coupons.Url);
+        }
+    }
+}


src/Stripe.Tests.XUnit/coupons_new/coupons.cs
```diff
@@ -0,0 +1,116 @@
...
+        public void delete_succeeds()
+        {
+            var coupon = service.Delete(resourceId);
+
+            Assert.NotNull(coupon);
+            coupon.Deleted.Should().BeTrue();
+        }
+
+        [Fact]
+        public void list_succeeds()
```
remi-stripe: For now I copy/pasted what we have in the other test. My guess is that this can be heavily simplified though.

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            StripeConfiguration.SetApiBase(this.previousApiBase);
+        }
+
+        [Fact]
+        public void create_succeeds()
+        {
+            var options = new StripeCouponCreateOptions() {
+                // Add a space at the end to ensure the ID is properly URL encoded
+                // when passed in the URL for other methods
```
brandur-stripe: OMG, this comment: I see that this looks copied from elsewhere, but does anyone know what exactly causes this problem?

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
```diff
@@ -0,0 +1,116 @@
...
+                Id = "co_123 ",
```
brandur-stripe: I think we were using the coupon archetype ID of `25OFF` elsewhere. It may be worth reusing that here too.

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
```diff
@@ -0,0 +1,116 @@
...
+    {
+        private string previousApiBase;
+        private readonly string resourceId = "co_123";
+        private StripeCouponService service;
+
+        public coupon_test()
+        {
+            this.service = new StripeCouponService(Cache.ApiKey);
+            this.previousApiBase = StripeConfiguration.GetApiBase();
+            StripeConfiguration.SetApiBase("http://localhost:12111/v1");
```
brandur-stripe: Just as a quick sanity check: is there any possibility that tests run in parallel? If so, this isn't safe.

brandur-stripe:
Thanks for getting the ball rolling Remi! Just since I haven't been following Stripe.NET too closely — Xunit tests are definitely the way we want to go with in the future?

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
remi-stripe: @brandur-stripe it's mostly because some endpoint allow you to pass user-chosen ids and so we had encoding issues. We should re-build the encoding logic to do this later in the pipeline though we haven't scheduled the work.
I'm not planning to keep it, I was really mostly reproducing all the tests for the normal coupon package.

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
remi-stripe: Yeah We likely would just remove Id since it's not a recommended parameter anymore. We needed it before because it was easier/required to have a valid one for other APIs (since we needed real values). Now we can just pass co_123 for all when needed since stripe-mock does not care.

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
remi-stripe: As far as I could tell it's not happening no, and I confirmed that tests fail if I don't clean up after the destruction, so I think it works! We will definitely confirm 100% before merging anything

remi-stripe:
@brandur-stripe It mostly came from ob's idea that we could likely migrate tests slowly which ended up working easily!

As far as I could tell XUnit is still the recommended solution. cc @jaymedavis in case you have more context as we were discussing this just last week!

jaymedavis:
Glad to see you are carefully selecting the new framework! It was disappointing to see mspec fall behind. So many tests. 😞 

[Here are some good reasons to use xUnit](https://dev.to/hatsrumandcode/net-core-2-why-xunit-and-not-nunit-or-mstest--aei). I'm sure you could find an opinionated piece about any of the three choices: xUnit, nUnit, mstest, to be fair though.

I went and looked through the [.NET Foundation](https://github.com/dotnet) projects (corefx, coreclr, cli, orleans, etc), and they are all using xUnit. That's a good reason alone, imo.

I haven't looked through stripe-mock, but do some research around `Theory`'s in xUnit - I bet there's a great integration to make there.

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
ob-stripe: I think Brandur was confused like I was: the comment implies that we _have_ to add a space to work around some URL-encoding problem, not that we're adding a space specifically to test that URL-encoding works as expected. In any case, URL-encoding correctness should be tested separately.

remi-stripe: yep but at the time that was the easiest way to make sure we did not break it again. Totally happy to remove this once we agree this shape of test is the right direction!

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        ~coupon_test()
+        {
+            StripeConfiguration.SetApiBase(this.previousApiBase);
```
ob-stripe: I don't think this works. With xUnit, if you have cleanup code, your test class must implement `IDisposable` and the cleanup code must be placed in a `Dispose()` method. Cf. https://xunit.github.io/docs/shared-context.

remi-stripe: hmmm weirdly it definitely worked locally but I'll investigate further in that case

jaymedavis:
ooc: stripe-mock looks *awesome*. The `image: Visual Studio 2017` might be able to spin this up, but I would consider moving the build to Travis CI along with your other bindings. Just something to think about for #future :)

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
brandur-stripe: I think that either will often work in practice, but the difference is that unlike destructors in C/C++, finalizers in C# are not invoked deterministically — they'll be invoked by the GC only when it deems it's the right time to clean up the instance. For this sort of test teardown, that might produce some unexpected results when you're running a number of tests together that are all modifying this sort of global configuration. ([See docs on finalizers for details](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/destructors).)

As OB suggests, you probably want `Dispose` instead because XUnit can invoke that explicitly as soon as this suite has finished running.

jaymedavis: The destructor calls .Finalize() in .net... that's why this got cleaned up when the garbage collector was finished with it. With `IDisposable`, you have more control. You can call Dispose() or implement your code in a `using` block. The Finalizer (.Finalize() or ~) is going to run when the object gets destroyed, but unfortunately you don't know exactly when that will occur.

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
brandur-stripe: Oh man, yep, I was confused by that comment.

src/Stripe.Tests.XUnit/coupons_new/coupons.cs
jaymedavis: [Your choice](https://stackoverflow.com/questions/1408175/execute-unit-tests-serially-rather-than-in-parallel). If possible, shoot for 100% isolation imo. 👍

brandur-stripe:
@jaymedavis 

> Here are some good reasons to use xUnit. I'm sure you could find an opinionated piece about any of the three choices: xUnit, nUnit, mstest, to be fair though.

Hah, I'd just found that same article via Google before you sent it :) It's the most compelling (relatively recent) argument for any of the three test frameworks that I was able to find.

> I went and looked through the .NET Foundation projects (corefx, coreclr, cli, orleans, etc), and they are all using xUnit. That's a good reason alone, imo.

Nice! That does seem like pretty strong official confirmation. +1 XUnit.

remi-stripe:
closing in favour of https://github.com/stripe/stripe-dotnet/pull/1229

PR #1229, opened 2018-07-11
BaseStripeTest class with stripe-mock setup

ob-stripe:
cc @stripe/api-libraries 

Adds a `BaseStripeTest` class that takes care of setting up and tearing down all the stripe-mock stuff.

Also adds a `StripeCouponServiceTest` to illustrate how to use `BaseStripeTest` and hopefully serve as a template for other tests.

`StripeCouponServiceTest` implements `IClassFixture<BaseStripeTest>` rather than just `BaseStripeTest` so that the initialization is done once for the entire class rather than for every test. 

Unfortunately, because the base URLs can only be set globally, I had to disable parallel test execution by grouping all tests in the same collection (by default, tests are grouped by class [0]). One alternative would be to add the ability to provide the base URLs in `StripeRequestOptions` (as can already be done for the API key) and pass an instance with the correct values in all the stripe-mock tests.

[0] https://xunit.github.io/docs/running-tests-in-parallel.html#runners-and-test-frameworks

### DIFF
--- a/src/StripeTests/BaseStripeTest.cs
+++ b/src/StripeTests/BaseStripeTest.cs
@@ -1,0 +1,150 @@
+namespace StripeTests
+{
+    using System;
+    using System.Linq;
+    using System.Net;
+    using System.Net.Http;
+    using System.Threading;
+
+    using Stripe;
+
+    public class BaseStripeTest
+    {
+        /// <value>Minimum required version of stripe-mock</value>
+        /// <remarks>
+        /// If you bump this, don't forget to bump `STRIPE_MOCK_VERSION` in appveyor.yml as well.
+        /// </remarks>
+        private const string MockMinimumVersion = "0.19.0";
+
+        private static string port;
+
+        // Lazy initializer to ensure that initialization is run only once even when running tests
+        // in parallel.
+        private static Lazy<object> initializer = new Lazy<object>(InitStripeMock);
+
+        public BaseStripeTest()
+        {
+            // This triggers the lazy initialization. We don't actually care about the value of
+            // initialized (it will be null anyway), but simply writing `initializer.Value` is not
+            // a valid statement in C#.
+            var initialized = initializer.Value;
+        }
+
+        /// <summary>
+        /// Gets fixture data from stripe-mock for a resource expected to be at the given API path.
+        /// stripe-mock ignores whether IDs are actually valid, so it's only important to make sure
+        /// that the route exists, rather than the actual resource. It's common to use a symbolic
+        /// ID stand-in like <code>ch_123</code>
+        /// </summary>
+        /// <param name="path">API path to use to get a fixture for stripe-mock</param>
+        /// <returns>Fixture data encoded as JSON</returns>
+        protected static string GetFixture(string path)
+        {
+            return GetFixture(path, null);
+        }
+
+        /// <summary>
+        /// Gets fixture data with expansions specified. Expansions are specified the same way as
+        /// they are in the normal API like <code>customer</code> or <code>data.customer</code>.
+        /// Use the special <code>*</code> character to specify that all fields should be
+        /// expanded.
+        /// </summary>
+        /// <param name="path">API path to use to get a fixture for stripe-mock</param>
+        /// <param name="expansions">Set of expansions that should be applied</param>
+        /// <returns>Fixture data encoded as JSON</returns>
+        protected static string GetFixture(string path, string[] expansions)
+        {
+            string url = $"http://localhost:{port}{path}";
+
+            if (expansions != null)
+            {
+                string query = string.Join("&", expansions.Select(x => $"expand[]={x}").ToArray());
+                url += $"?{query}";
+            }
+
+            using (HttpClient client = new HttpClient())
+            {
+                client.DefaultRequestHeaders.Authorization
+                    = new System.Net.Http.Headers.AuthenticationHeaderValue(
+                        "Bearer",
+                        "sk_test_123");
+
+                HttpResponseMessage response;
+
+                try
+                {
+                    response = client.GetAsync(url).Result;
+                }
+                catch (Exception)
+                {
+                    throw new Exception(
+                        $"Couldn't reach stripe-mock at `localhost:{port}`. "
+                        + "Is it running? Please see README for setup instructions.");
+                }
+
+                if (response.StatusCode != HttpStatusCode.OK)
+                {
+                    throw new Exception(
+                        $"stripe-mock returned status code: {response.StatusCode}.");
+                }
+
+                return response.Content.ReadAsStringAsync().Result;
+            }
+        }
+
+        /// <summary>
+        /// Checks that stripe-mock is running and that it's a recent enough version.
+        /// </summary>
+        private static object InitStripeMock()
+        {
+            port = Environment.GetEnvironmentVariable("STRIPE_MOCK_PORT") ?? "12111";
+            string url = $"http://localhost:{port}";
+
+            using (HttpClient client = new HttpClient())
+            {
+                HttpResponseMessage response;
+
+                try
+                {
+                    response = client.GetAsync(url).Result;
+                }
+                catch (Exception)
+                {
+                    throw new Exception(
+                        $"Couldn't reach stripe-mock at `localhost:{port}`. "
+                        + "Is it running? Please see README for setup instructions.");
+                }
+
+                string version = response.Headers.GetValues("Stripe-Mock-Version").FirstOrDefault();
+
+                if (!version.Equals("master") &&
+                    (CompareVersions(version, MockMinimumVersion) > 0))
+                {
+                    throw new Exception(
+                        $"Your version of stripe-mock ({version}) is too old. The minimum "
+                        + $"version to run this test suite is {MockMinimumVersion}. Please see its "
+                        + "repository for upgrade instructions.");
+                }
+            }
+
+            StripeConfiguration.SetApiBase($"http://localhost:{port}/v1");
+            StripeConfiguration.SetUploadsBase($"http://localhost:{port}/v1");
+            StripeConfiguration.SetApiKey("sk_test_123");
+
+            return null;
+        }
+
+        /// <summary>
+        /// Compares two version strings.
+        /// </summary>
+        /// <param name="a">A version string (e.g. "1.2.3").</param>
+        /// <param name="b">Another version string.</param>
+        /// <returns>-1 if a > b, 1 if a < b, 0 if a == b</returns>
+        private static int CompareVersions(string a, string b)
+        {
+            var version1 = new Version(a);
+            var version2 = new Version(b);
+            return version2.CompareTo(version1);
+        }
+    }
+}
--- a/src/StripeTests/Entities/StripeCouponTest.cs
+++ b/src/StripeTests/Entities/StripeCouponTest.cs
@@ -1,0 +1,24 @@
+namespace StripeTests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    using Newtonsoft.Json;
+    using Stripe;
+    using Xunit;
+
+    public class StripeCouponTest : BaseStripeTest
+    {
+        [Fact]
+        public void Deserialize()
+        {
+            string json = GetFixture("/v1/coupons/co_123");
+            var coupon = Mapper<StripeCoupon>.MapFromJson(json);
+            Assert.NotNull(coupon);
+            Assert.IsType<StripeCoupon>(coupon);
+            Assert.NotNull(coupon.Id);
+            Assert.Equal("coupon", coupon.Object);
+        }
+    }
+}
--- a/src/StripeTests/Services/Coupons/StripeCouponServiceTest.cs
+++ b/src/StripeTests/Services/Coupons/StripeCouponServiceTest.cs
@@ -1,0 +1,112 @@
+namespace StripeTests
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    using Stripe;
+    using Xunit;
+
+    public class StripeCouponServiceTest : BaseStripeTest
+    {
+        private const string CouponId = "co_123";
+
+        private StripeCouponService service;
+        private StripeCouponCreateOptions createOptions;
+        private StripeCouponUpdateOptions updateOptions;
+        private StripeCouponListOptions listOptions;
+
+        public StripeCouponServiceTest()
+        {
+            this.service = new StripeCouponService();
+
+            this.createOptions = new StripeCouponCreateOptions()
+            {
+                PercentOff = 25,
+                Duration = "forever",
+            };
+
+            this.updateOptions = new StripeCouponUpdateOptions()
+            {
+                Metadata = new Dictionary<string, string>()
+                {
+                    { "key", "value" },
+                },
+            };
+
+            this.listOptions = new StripeCouponListOptions()
+            {
+                Limit = 1,
+            };
+        }
+
+        [Fact]
+        public void Create()
+        {
+            var coupon = this.service.Create(this.createOptions);
+            Assert.NotNull(coupon);
+        }
+
+        [Fact]
+        public async Task CreateAsync()
+        {
+            var coupon = await this.service.CreateAsync(this.createOptions);
+            Assert.NotNull(coupon);
+        }
+
+        [Fact]
+        public void Delete()
+        {
+            var deleted = this.service.Delete(CouponId);
+            Assert.NotNull(deleted);
+        }
+
+        [Fact]
+        public async Task DeleteAsync()
+        {
+            var deleted = await this.service.DeleteAsync(CouponId);
+            Assert.NotNull(deleted);
+        }
+
+        [Fact]
+        public void Get()
+        {
+            var coupon = this.service.Get(CouponId);
+            Assert.NotNull(coupon);
+        }
+
+        [Fact]
+        public async Task GetAsync()
+        {
+            var coupon = await this.service.GetAsync(CouponId);
+            Assert.NotNull(coupon);
+        }
+
+        [Fact]
+        public void List()
+        {
+            var coupons = this.service.List(this.listOptions);
+            Assert.NotNull(coupons);
+        }
+
+        [Fact]
+        public async Task ListAsync()
+        {
+            var coupons = await this.service.ListAsync(this.listOptions);
+            Assert.NotNull(coupons);
+        }
+
+        [Fact]
+        public void Update()
+        {
+            var coupon = this.service.Update(CouponId, this.updateOptions);
+            Assert.NotNull(coupon);
+        }
+
+        [Fact]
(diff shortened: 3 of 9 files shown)


brandur-stripe:
Slick OB!

> `StripeCouponServiceTest` implements `IClassFixture<BaseStripeTest>` rather than just `BaseStripeTest` so that the initialization is done once for the entire class rather than for every test. 

Nice touch.

> Unfortunately, because the base URLs can only be set globally, I had to disable parallel test execution by grouping all tests in the same collection (by default, tests are grouped by class [0]). One alternative would be to add the ability to provide the base URLs in `StripeRequestOptions` (as can already be done for the API key) and pass an instance with the correct values in all the stripe-mock tests.

IMO ,adding the ability to set this configuration per-request (or preferably per-client although I don't think we really have such a concept in .NET) would be better.

src/Stripe.Tests.XUnit/BaseStripeTest.cs
```diff
@@ -0,0 +1,106 @@
...
+        private string OrigApiBase;
+        private string OrigUploadBase;
+        private string OrigApiKey;
+
+        public BaseStripeTest()
+        {
+            CheckStripeMock();
+            SetUpStripeMock();
+        }
+
```
jaymedavis: One thing I liked in @remi-stripe's other PR was a destructor. Having additional cleanup area could prove useful. 

~BaseStripeTest()
{
}

src/Stripe.Tests.XUnit/BaseStripeTest.cs
```diff
@@ -0,0 +1,106 @@
...
+        public void Dispose()
+        {
+            TearDownStripeMock();
+        }
```
jaymedavis: You'd need to suppress the garbage collection in dispose for most cases

```GC.SuppressFinalize(this);```

**Comment by ob-stripe** [comment]
@jaymedavis Thanks for the pointers! I found a StackOverflow answer that seems relevant: https://stackoverflow.com/a/151244/5307473. I don't think we need this kind of additional complexity at this point, but I'll keep it in mind.

> IMO ,adding the ability to set this configuration per-request (or preferably per-client although I don't think we really have such a concept in .NET) would be better.

I tried taking a stab at this, but I think it will require some refactoring. Currently, services form the entire URL using the `Urls` package, which directly reads from `StripeConfiguration.GetApiBase()`. I think we should modify it to have services form the _path_ only, and have the `Requestor` class then form the full URL by joining the base URL and the path.

**Comment by ob-stripe** [comment]
I've made a few updates to the PR:
- stripe-mock is now cached on AppVeyor's end, so that it's only downloaded if `appveyor.yml` changes (e.g. when bumping `STRIPE_MOCK_VERSION`)
- added a deserialization test in `Entities/StripeCouponTest.cs`, which uses a `GetFixture()` method similar to what we do in stripe-java
- added a (very basic) serialization test in `Services/StripeCouponCreateOptionsTest.cs`. I'm not 100% sure about this one, in particular I wonder if serialization is deterministic (i.e. if parameters will always be in the same order)
- changed the `BaseStripeTest` class to only run the stripe-mock check once (via static variables). This way tests can directly inherit `BaseStripeTest` and benefit from any helper methods we define (e.g. `GetFixture`)

File: src/Stripe.Tests.XUnit/BaseStripeTest.cs [hunk 3]
```diff
@@ -0,0 +1,161 @@
...
+    {
+        /// <value>Minimum required version of stripe-mock</value>
+        /// <remarks>
+        /// If you bump this, please don't forget to bump STRIPE_MOCK_VERSION in appveyor.yml.
+        /// </remarks>
+        private const string MockMinimumVersion = "0.19.0";
+
+        private static string Port;
+
+        private string OrigApiBase;
```
remi-stripe: Orig => Original ? (stripe-dotnet does not really abbreviate variable names, let's not start that trend)

src/Stripe.Tests.XUnit/BaseStripeTest.cs
```diff
@@ -0,0 +1,161 @@
...
+        private string OrigUploadBase;
+        private string OrigApiKey;
```
remi-stripe: Should be alphabetize while we're at it?

src/Stripe.Tests.XUnit/BaseStripeTest.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        public BaseStripeTest()
+        {
+            if (string.IsNullOrEmpty(Port))
+            {
+                CheckStripeMock();
```
remi-stripe: Start with `this.` on all methods that are in the same class.

src/Stripe.Tests.XUnit/BaseStripeTest.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        private static string Port;
+
+        private string OrigApiBase;
+        private string OrigUploadBase;
+        private string OrigApiKey;
+
+        public BaseStripeTest()
+        {
+            if (string.IsNullOrEmpty(Port))
```
remi-stripe: Start with `this.` for `Port` and all other class members

src/Stripe.Tests.XUnit/BaseStripeTest.cs
```diff
@@ -0,0 +1,161 @@
...
+                }
+
+                if (response.StatusCode != HttpStatusCode.OK)
+                {
+                    throw new Exception(
+                        $"stripe-mock returned status code: {response.StatusCode}."
+                    );
+                }
+
+                return response.Content.ReadAsStringAsync().Result;
```
remi-stripe: This only supports stripe-mock fixtures right? We should add a JSON one like in Java, those really are the most important ones to have/test.

src/Stripe.Tests.XUnit/BaseStripeTest.cs
```diff
@@ -0,0 +1,161 @@
...
+
+                string version = response.Headers.GetValues("Stripe-Mock-Version").FirstOrDefault();
+
+                if (!(version.Equals("master")) &&
+                    (CompareVersions(version, MockMinimumVersion) > 0))
+                {
+                    throw new Exception(
+                        $"Your version of stripe-mock ({version}) is too old. The minimum "
+                        + $"version to run this test suite is {MockMinimumVersion}. Please see its "
+                        + "repository for upgrade instructions."
```
remi-stripe: OOC why don't we require the exact same version? New versions will likely break tests too (that happened to be at least twice in Java)

src/Stripe.Tests.XUnit/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+using Xunit;
+
+// By default, group all tests in a single collection. This ensures the tests are run sequentially,
+// which is necessary because we have to change the global configuration in order to use
+// stripe-mock.
+[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly)]
```
remi-stripe: maybe add a todo to remove that once we have completed the migration to stripe-mock

src/Stripe.Tests.XUnit/Services/Coupons/StripeCouponServiceTest.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            var coupon = await Service.GetAsync(CouponId);
+            Assert.NotNull(coupon);
+        }
+
+        [Fact]
+        public void List()
+        {
+            var coupons = Service.List(ListOptions);
+            Assert.NotNull(coupons);
```
remi-stripe: Should we confirm what's _in_ the list?

src/Stripe.Tests.XUnit/Services/Coupons/StripeCouponCreateOptionsTest.cs
```diff
@@ -0,0 +1,29 @@
...
+        [Fact]
+        public void Serialize()
+        {
+            var options = new StripeCouponCreateOptions() {
+                PercentOff = 25,
+                Duration = "forever",
+            };
+
+            var url = this.Service.ApplyAllParameters(options, "", false);
+            Assert.Equal("?percent_off=25&duration=forever", url);
```
remi-stripe: I like this approach but not entirely sure we want this for every single resource. I think something like what we have in Go to test all "types" could work better.
Otherwise we likely want to only test complex types like lists? We also should test all options class otherwise so create and update here.

src/Stripe.Tests.XUnit/BaseStripeTest.cs
ob-stripe: Yep. Ideally I'd like `BaseStripeTest` to have the same helper methods as we have in stripe-java, so it's still missing things like:
- retrieving contents of a local fixture file
- stubbing requests
- verifying the requested URL / parameters

The last two will be fairly complex to implement (and will likely require using a library like [moq](https://github.com/moq/moq4)) but the first one shouldn't be too hard.

src/Stripe.Tests.XUnit/BaseStripeTest.cs
ob-stripe: Yes, but that's what you want:
1. Upgrade stripe-mock
2. See which tests fail, and fix the code (or the tests themselves)

If the version check prevents the test suite from running at all, then 2. becomes more difficult :)

src/Stripe.Tests.XUnit/Services/Coupons/StripeCouponServiceTest.cs
ob-stripe: Not sure there'd be much value in it since I think whatever stripe-mock returns will be coerced into the expected type. I'll check though.

remi-stripe: yeah you're likely right. I always have a hard time letting go of all the code that tested (and caught) so many thing before but it's likely superfluous and we just need good tests for `StripeList` in general

src/Stripe.Tests.XUnit/Services/Coupons/StripeCouponCreateOptionsTest.cs
ob-stripe: Right, to be clear, this particular test is fairly useless and I completely agree that most of this stuff should be tested only once on `ParameterBuilder`. This was just an example of what a custom serialization test could work. In practice, we'd probably only want these for classes that use custom plugins.

ob-stripe:
Alright, I moved the new tests into a new project `StripeTests`, and disabled `StyleCop.Analyzers` in the existing `Stripe.Tests.Xunit` project. Should make it easier to ensure that our new tests are compliant with the linting rules without getting tens of thousands of warnings.

src/StripeTests/BaseStripeTest.cs
```diff
@@ -0,0 +1,158 @@
...
+    {
+        /// <value>Minimum required version of stripe-mock</value>
+        /// <remarks>
+        /// If you bump this, don't forget to bump `STRIPE_MOCK_VERSION` in appveyor.yml as well.
+        /// </remarks>
+        private const string MockMinimumVersion = "0.19.0";
+
+        private static string port;
+
+        private string originalApiBase;
```
remi-stripe: since we are moving tests to a new package, they will all be based on stripe-mock (or in the future stubbed). Can we get rid of all this in that case? Or are the test suites ran in parallel and so they conflict?

remi-stripe:
I think this PR is almost there. Do we want to slowly clean the relevant files/resources tested in the 2 other test suites? Or do we simply plan to migrate everything first and then later remove the other test suite in one go?

jaymedavis:
You could mark them `[Obselete]` as you go. Perhaps configure the test runner to skip those.

remi-stripe:
@jaymedavis yeah but it's even more work than just removing the files at that point

src/StripeTests/BaseStripeTest.cs
ob-stripe: I think there's _some_ value in cleaning up (though tbh we could simply restore the default values instead of saving the original values).

The tests are run in parallel on AppVeyor (not sure why they aren't on our laptops -- xUnit uses the "number of virtual CPUs in the PC" by default, I guess that xUnit only sees 1 for some reason). It should still work because all the tests in this project will use the same global state values, though ideally we shouldn't rely on this. A better solution would be to be able to provide all options directly in each tests, via `RequestOptions` or an hypothetical `StripeClient`, but that's not possible right now.

ob-stripe:
No strong preference. If we do decide to clean up as we go, flagging the tests as `[Obsolete]` is probably better -- earlier in the PR I tried removing all the coupons related stuff from the old test suite, only to find in CI that I broke other tests that depended on the coupons fixtures.

src/StripeTests/BaseStripeTest.cs
jaymedavis: > xUnit uses the "number of virtual CPUs in the PC" by default
Interesting. Never knew that.

You could create a `TestRequestOptions` of sorts which has properties for api base, upload base, and api key (private strings on BaseStripeTest now). Have the constructor of BaseStripeTest take a parameter of type `TestRequestOptions`. If the test doesn't provide this object, have it use the default set.

src/StripeTests/BaseStripeTest.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                CheckStripeMock();
+            }
+
+            this.SetUpStripeMock();
+        }
+
+        public void Dispose()
+        {
+            this.TearDownStripeMock();
```
remi-stripe: I added extra tests and this line causes one of them to fail. It works fine without it which seems to point to some kind of parallelism issue.

ob-stripe: Ah right, I guess that one test could restore the original state while another test is ongoing.

Honestly, the right answer here is to disable parallel test execution as long as we have to rely on global state. It shouldn't be too much of an issue since stripe-mock tests are fast anyway.

src/StripeTests/BaseStripeTest.cs
ob-stripe: That sounds like a good idea if a specific test needs to use different values, but from past experience from the other language libraries, I don't anticipate that being the case. In other libraries, tests fall into one of two categories:
- most use stripe-mock
- some use stubbed requests (and thus don't care about values for `ApiBase` etc. since they don't make network requests)

src/StripeTests/BaseStripeTest.cs
remi-stripe: But if the test suites runs separately (like as 3 separate projects they must right?) then that new one does not need the hack since it would only ever use stripe-mock right? This was only needed before you decided to add a third project I think.

ob-stripe: Yes, the different test projects are run separately. Here's what's currently happening in the new project:
- T1: call `SetUpStripeMock()`
- T1: run a test
- T2: call `SetUpStripeMock()` (effectively a no-op since the values are the same)
- T1: call `TearDownStripeMock()`
- T2: run a test --> fails because T1 restored the original state values and so T2's requests are sent to the live API servers instead of stripe-mock

We could simply set the global state once then never clean up and that would probably work, but it seems less hacky to me to disable parallel execution until we no longer have to rely on global state (i.e. until we can provide `ApiBase` / `UploadBase` on a per-request or per-service basis).

ob-stripe: It just seems cleaner to me to have isolated tests that don't rely on global state.

That said, it looks like stripe-go does exactly this (i.e. sets the global configuration once and never restore it afterwards) so I've updated the PR to do the same thing.

ob-stripe:
ptal @remi-stripe

ob-stripe:
Sight, this was more annoying than I thought, but it should be working now. I've changed the initialization to use a `Lazy<T>` class. Normally, this is used to defer the initialization of some object until the first time it's needed. In this case, we're not really initializing an object (just setting some global state) so the initializer factory method (`InitStripeMock`) simply returns `null`, and we don't really care about deferring initialization so we call `initializer.Value` (which triggers the lazy initialization) in `BaseStripeTest`'s constructor.

The interesting property of `Lazy<T>` is that they're thread-safe:
> By default, `Lazy<T>` objects are thread-safe. That is, if the constructor does not specify the kind of thread safety, the `Lazy<T>` objects it creates are thread-safe. In multi-threaded scenarios, the first thread to access the `Value` property of a thread-safe `Lazy<T>` object initializes it for all subsequent accesses on all threads, and all threads share the same data. Therefore, it does not matter which thread initializes the object, and race conditions are benign.

So this should ensure that even when running tests in parallel, initialization will only be run once and threads will block until initialization is complete.

src/StripeTests/BaseStripeTest.cs
```diff
@@ -0,0 +1,157 @@
...
+                return response.Content.ReadAsStringAsync().Result;
+            }
+        }
+
+        /// <summary>
+        /// Sets the global configuration parameters to use stripe-mock.
+        /// </summary>
+        private static void SetUpStripeMock()
+        {
+        }
```
remi-stripe: is this still needed?

src/StripeTests/Entities/StripeCouponTest.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class StripeCouponTest : BaseStripeTest
+    {
+        [Fact]
+        public void Deserialize()
+        {
+            string json = GetFixture("/v1/coupons/co_123");
+            var coupon = Mapper<StripeCoupon>.MapFromJson(json);
+            Assert.NotNull(coupon);
+            Assert.IsType<StripeCoupon>(coupon);
+            Assert.NotNull(coupon.Id);
```
remi-stripe: We discussed yesterday that we should test `Assert.Equal("coupon", coupon.Object)` here too

src/StripeTests/Services/Coupons/StripeCouponServiceTest.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                Metadata = new Dictionary<string, string>()
+                {
+                    { "key", "value" },
+                },
+            };
+
+            this.listOptions = new StripeCouponListOptions()
+            {
+                Limit = 1,
```
remi-stripe: Do we care to set that `Limit` if we don't explicitly test it?

src/StripeTests/BaseStripeTest.cs
ob-stripe: Nope. Good catch, thanks.

src/StripeTests/Services/Coupons/StripeCouponServiceTest.cs
ob-stripe: Yeah, I think so -- just like for creation and update requests, it at least ensures that the method does accept a `StripeCouponListOptions` and that the resulting encoding is accepted by stripe-mock.

remi-stripe: sure but I meant you could do this without setting `Limit`.

PR #1231, opened 2018-07-12
Add Stylecop.Analyzers

ob-stripe:
cc @stripe/api-libraries 

Linters! My favorite thing. :)

Right now the linter produces (a lot of) warnings. We'll need to solve all of them (we can customize some of the rules if we think they're not a good fit for us) then change the configuration to produce errors instead of warnings, similar to what we did in stripe-java.

### DIFF
--- a/src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
+++ b/src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
@@ -19,6 +19,10 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
+    <PackageReference Include="Stylecop.Analyzers" Version="1.0.2">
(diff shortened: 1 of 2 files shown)


src/Stripe.Tests.XUnit/Stripe.Tests.XUnit.csproj
```diff
@@ -19,6 +19,10 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
+    <PackageReference Include="Stylecop.Analyzers" Version="1.0.2">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
```
ob-stripe: These two lines instruct NuGet to not add `Stylecop.Analyzers` as a package dependency, so that it's not propagated to Stripe.net users. Cf. https://github.com/NuGet/Home/issues/7051.

PR #1232, opened 2018-07-12
Start fixing the easiest StyleCop issues

remi-stripe:
This PR explicitly ignores the most common StyleCop issues and fixes the less common ones to make sure it's the right direction.

cc @ob-stripe 
cc @stripe/api-libraries


src/Stripe.net/Stripe.net.csproj
```diff
@@ -19,6 +19,7 @@
     <PackageTargetFallback Condition=" '$(TargetFramework)' == 'netstandard1.2' ">$(PackageTargetFallback);netcoreapp1.0</PackageTargetFallback>
     <NetStandardImplicitPackageVersion Condition=" '$(TargetFramework)' == 'netstandard1.2' ">1.6.1</NetStandardImplicitPackageVersion>
     <SignAssembly>True</SignAssembly>
+    <CodeAnalysisRuleSet>Settings.stylecop</CodeAnalysisRuleSet>
```
olivierbellone: I don't think you need this one.

remi-stripe: ah yes sorry forgot that I copy/pasted that everywhere to try to make it work :p

remi-stripe:
Okay I pushed fixes for almost all the rules. There are a few left and I added comments as to what we need to agree on first.

src/Stripe.net/Infrastructure/Middleware/ParserPlugin/DateFilterPlugin.cs
```diff
@@ -14,7 +14,7 @@ public bool Parse(ref string requestString, JsonPropertyAttribute attribute, Pro
                 return false;
             }
 
-            var filter = (StripeDateFilter) propertyValue;
+            var filter = (StripeDateFilter)propertyValue;
```
remi-stripe: While I implemented that one, I feel quite strongly that this missing space is not nice to look at and reason with.

remi-stripe:
Okay almost all rules have been fixed. The ones left are a bit more complex or just painful (headers on all files) and left disabled for now. We also have 2 rules introducing breaking changes that we will fix on the next major release.

r? @ob-stripe

brandur-stripe:
OMG Remi you're a BOSS. Amazing work here.

remi-stripe:
@ob-stripe Okay I was able to fix the rule requiring `this` for local calls/variables after all with a few regexes.

I think this is as far as we can go without breaking changes so it's ready to be reviewed/merged.

ob-stripe:
Amazing @remi-stripe!

Even with `?w=1`, it's kind of difficult to review this PR very closely so I'm largely relying on the test suite here, but lgtm.

PR #1234, opened 2018-07-13
Add support for Customer on Source creation for shared Sources

remi-stripe:
The `customer` parameter can only be used when creating a shared Source with Stripe Connect.

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Released as 17.3.0.

PR #1236, opened 2018-07-16
Move test suite to stripe-mock

remi-stripe:
This PR is used to move all tests to stripe-mock. This will be ongoing work for a while as we move everything over and fix all edge-cases discovered similarly to what we did in Java recently.

cc @stripe/api-libraries


remi-stripe:
@ob-stripe Can you have a look and confirm this all looks good? If so, I'll start porting all the ones that don't require custom work/stubbing

src/StripeTests/StripeTests.csproj
```diff
@@ -13,6 +13,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
```
ob-stripe: I assume that you didn't add this on purpose (it's not necessary since we already import a more recent version of the same package just two lines below). Can you remove the line and run `dotnet restore` from the root of the repo to ensure that the new version is installed?

remi-stripe: Weird, I likely messed up my rebase. Thanks for catching, will fix.

src/StripeTests/Entities/StripeApplicationFeeTest.cs
```diff
@@ -0,0 +1,24 @@
...
+        public void Deserialize()
+        {
+            string json = GetFixture("/v1/application_fees/fee_123");
+            var applicationFee = Mapper<StripeApplicationFee>.MapFromJson(json);
+            Assert.NotNull(applicationFee);
+            Assert.IsType<StripeApplicationFee>(applicationFee);
+            Assert.NotNull(applicationFee.Id);
+            Assert.Equal("application_fee", applicationFee.Object);
+        }
+    }
```
ob-stripe: This class has expandable attributes. Could you add a test like [this one](https://github.com/stripe/stripe-java/blob/master/src/test/java/com/stripe/model/ApplicationFeeTest.java#L28) in stripe-java?

`GetFixture()` already supports a second `string[]` arguments with the list of attributes you want to expand.

remi-stripe: Nice catch, added a test for that!

src/StripeTests/StripeTests.csproj
remi-stripe: fixed

src/StripeTests/Entities/StripeApplicationFeeTest.cs
```diff
@@ -0,0 +1,49 @@
...
+            };
+
+            string json = GetFixture("/v1/application_fees/fee_123", expansions);
+            var applicationFee = Mapper<StripeApplicationFee>.MapFromJson(json);
+            Assert.NotNull(applicationFee);
+            Assert.IsType<StripeApplicationFee>(applicationFee);
+            Assert.NotNull(applicationFee.Id);
+            Assert.Equal("application_fee", applicationFee.Object);
+
+            Assert.Equal("account", applicationFee.Account.Object);
```
ob-stripe: You should probably add `Assert.NotNull(applicationFee.Account)` before this assert (and do something similar for all the following asserts), otherwise the test would crash if the attribute was null for some reason.

It's not a huge deal -- in both cases, the test would fail. It's just cleaner and more explicit to test with a failed assertion rather than with a null pointer exception :)

src/Stripe.net/Entities/StripeBalance.cs
```diff
@@ -5,6 +5,9 @@
 
     public class StripeBalance : StripeEntity
     {
+        [JsonProperty("object")]
+        public string Object { get; set; }
+
```
remi-stripe: Discovered that this class did not have `Object` so added it. Let me know if it should be a separate commit.

src/StripeTests/Entities/CountrySpecTest.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        [Fact]
+        public void Deserialize()
+        {
+            string json = GetFixture("/v1/country_specs/US");
+            var countrySpec = Mapper<CountrySpec>.MapFromJson(json);
+            Assert.NotNull(countrySpec);
+            Assert.IsType<CountrySpec>(countrySpec);
+            Assert.NotNull(countrySpec.Id);
+            Assert.Equal("country_spec", countrySpec.Object);
```
remi-stripe: Part of me still wants to test deserialization of complex sub-resource types but I think we want to avoid this and just trust the framework.

src/StripeTests/Entities/StripeApplePayDomainTest.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        [Fact]
+        public void Deserialize()
+        {
+            string json = GetFixture("/v1/apple_pay/domains/apwc_123");
+            var domain = Mapper<StripeApplePayDomain>.MapFromJson(json);
+            Assert.NotNull(domain);
+            Assert.IsType<StripeApplePayDomain>(domain);
+            Assert.NotNull(domain.Id);
+            Assert.Equal("apple_pay_domain", domain.Object);
```
remi-stripe: "Someone" should take moving things to `apple_pay.domain` for the object name as it would respect the new namespace norms

src/StripeTests/Entities/StripeCustomerTest.cs
```diff
@@ -0,0 +1,46 @@
...
+            };
+
+            string json = GetFixture("/v1/customers/cus_123", expansions);
+            var customer = Mapper<StripeCustomer>.MapFromJson(json);
+            Assert.NotNull(customer);
+            Assert.IsType<StripeCustomer>(customer);
+            Assert.NotNull(customer.Id);
+            Assert.Equal("customer", customer.Object);
+
+            // TODO: Fix this in stripe-mock as Account is not a valid Customer source.
```
remi-stripe: Flagging this one as it's weird. I wanted to make tests pass but we should fix stripe-mock

src/StripeTests/Entities/StripeDiscountTest.cs
```diff
@@ -0,0 +1,19 @@
...
+    using Newtonsoft.Json;
+    using Stripe;
+    using Xunit;
+
+    public class StripeDiscountTest : BaseStripeTest
+    {
+        [Fact]
+        public void Deserialize()
+        {
+            // TODO: Figure out how to test this resource.
```
remi-stripe: Flagging this one as one that needs a local fixture. There are a few of those.

Something tells me we should add fixture logic to stripe-mock for resources that can't be fetched instead of having local JSON files for each test suite where they go out of sync other time.

src/StripeTests/Entities/StripeInvoiceLineItemTest.cs
```diff
@@ -0,0 +1,53 @@
...
+    using Stripe;
+    using Xunit;
+
+    public class StripeInvoiceLineItemTest : BaseStripeTest
+    {
+        [Fact]
+        public void Deserialize()
+        {
+            // TODO: Figure out if we should test it this way as the class
+            // is generic for invoice items and line items.
```
remi-stripe: Flagging this one as stripe-dotnet has one unified class for 2 separate resources which makes testing it weird.

src/StripeTests/Entities/StripeTokenTest.cs
```diff
@@ -0,0 +1,24 @@
...
+    using Newtonsoft.Json;
+    using Stripe;
+    using Xunit;
+
+    public class StripeTokenTest : BaseStripeTest
+    {
+        [Fact]
+        public void Deserialize()
+        {
+            string json = GetFixture("/v1/tokens/tok_123");
```
remi-stripe: Part of me wants to test each type of token. I know you might say that's not what this test suite should be about but it's the last time I'll raise it since it's the last library we migrate :)

src/StripeTests/Services/FileUploads/StripeFileUploadServiceTest.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        private Stream stream;
+        private StripeFileUploadService service;
+        private StripeFileUploadListOptions listOptions;
+
+        public StripeFileUploadServiceTest()
+        {
+            this.service = new StripeFileUploadService();
+
+            this.stream = typeof(StripeFileUploadServiceTest).GetTypeInfo().Assembly.GetManifestResourceStream(FileName);
```
remi-stripe: I mimicked what was done in the Stripe.Net.Tests test suite, do you think it's the right solution?

remi-stripe:
@ob-stripe Can you do a full review on that new one? Want to make sure the approach is correct. I'll continue to add commits and not squash until we're ready to merge to make reviews easier.

src/Stripe.net/Entities/StripeBalance.cs
ob-stripe: 👍 That's fine, thanks for the heads up!

src/StripeTests/Entities/CountrySpecTest.cs
ob-stripe: Yeah, I think we'd only need in-depth tests for cases where we have custom deserializers. (Thankfully that doesn't seem to be the case in stripe-dotnet thanks to API version pinning.) In "simple" cases like this, we'd only be testing the Newtonsoft.Json library rather than stripe-dotnet.

src/StripeTests/Entities/StripeDiscountTest.cs
ob-stripe: I'm not sure that we want stripe-mock to be able to return any arbitrary resource without a "real" endpoint, but maybe we could embed the OpenAPI fixtures file in the library's test project. That would make it much easier to maintain and update.

src/StripeTests/Entities/StripeInvoiceLineItemTest.cs
ob-stripe: IMO that's something we should fix as it's inconsistent with the API and every other library, but in the meantime this is fine.

src/StripeTests/Services/FileUploads/StripeFileUploadServiceTest.cs
ob-stripe: Yep, that seems fine.

ob-stripe:
@remi-stripe Amazing work! Looks great so far.

src/StripeTests/Entities/CountrySpecTest.cs
remi-stripe: Yeah so not entirely. We'd also be testing the proper nesting and naming of properties too for example which is easy to mess up. But yes I get your point, I just could not resist raising it one last time :p

src/StripeTests/Entities/StripeDiscountTest.cs
remi-stripe: So it's not really "arbitrary resource". It's also "resources that are only returned on creation" for example like usage record. It forces us to have local files that go out of sync quickly (or are manually written incorrectly).

remi-stripe:
@ob-stripe Okay I went ahead and deleted a large number of tests from the other two test suites to make them faster/more reliable.

I left tests that we have not yet moved to stripe-mock and core tests that we need to put "somewhere" like the infrastructure/encoding ones.

Can you have a look? What do you think of merging this to start?

Separately my branch seems in a weird state as it mentions your merge commits above:

commits

Do you know if it's just a UI glitch? I tried to rebase/cherry-pick from master but no change.

ob-stripe:
Not sure what's up with your branch, but the diff looks okay. I guess you can just squash everything in a single commit before merging ¯\\\_(ツ)\_/¯

👍 for merging this

PR #1241, opened 2018-07-30
Move Sources to stripe-mock and fix an existing issue

remi-stripe:
Properly encode `owner` hash on source creation or update.

r? @ob-stripe 
cc @stripe/api-libraries 

Do you think it's worth waiting for that one and doing other major/breaking changes at the same time?


ob-stripe:
> Do you think it's worth waiting for that one and doing other major/breaking changes at the same time?

If you're planning on updating many more tests, I'd say yes, let's wait just in case you find other similar issues.

ob-stripe:
@remi-stripe We can close this one, right? The issue with `owner` should be fixed in the new major version.

PR #1242, opened 2018-07-31
Add account.created field

karlr-stripe:
Added the created field of Account to the relevant model.
https://stripe.com/docs/api/dotnet#account_object-created

r? @remi-stripe
cc @stripe/api-libraries


ob-stripe:
Released as 17.7.0.

PR #1243, opened 2018-07-31
Move infrastructure tests to our new test suite

remi-stripe:
Continue moving more tests to the new test suite

Focused on Infrastructure tests for now
- Ensure user agent and custom headers are set properly
- Ensure simple expansion is supported on a resource and on a list
- Ensure that multi-level expansion is supported
- Ensure parameter encoding works as expected

cc @ob-stripe to have a look
cc @stripe/api-libraries


src/StripeTests/Infrastructure/test_data/TestService.cs
```diff
@@ -0,0 +1,16 @@
+namespace StripeTests.Infrastructure.test_data
```
ob-stripe: Can you rename this to `TestData` instead (and also change the directory's name)? I don't think the use of snake case in namespaces is idiomatic in C#.

remi-stripe: fixed!

remi-stripe:
@ob-stripe Could I ask you to have a look? We're almost there, I need to add tests for Cards, Events and one or two encoding logic and we're done.

remi-stripe:
Okay this work is complete. The old test suites have been removed, all existing tests have been ported to the new test suite and it can now be reviewed and merged.

r? @ob-stripe

ob-stripe:
Pulled this in. Once again, amazing work @remi-stripe 🎉

PR #1244, opened 2018-08-02
file_link support

ob-stripe:
cc @stripe/api-libraries 

Adds support for file_links.

### DIFF
--- a/src/Stripe.net/Entities/StripeFileUpload.cs
+++ b/src/Stripe.net/Entities/StripeFileUpload.cs
@@ -13,6 +13,12 @@
         [JsonConverter(typeof(StripeDateTimeConverter))]
         public DateTime Created { get; set; }
 
+        [JsonProperty("filename")]
+        public string Filename { get; set; }
+
+        [JsonProperty("links")]
+        public StripeList<StripeFileLink> Links { get; set; }
+
         [JsonProperty("purpose")]
         public string Purpose { get; set; }
 
--- a/src/Stripe.net/Services/FileLinks/StripeFileLinkCreateOptions.cs
+++ b/src/Stripe.net/Services/FileLinks/StripeFileLinkCreateOptions.cs
@@ -1,0 +1,16 @@
+namespace Stripe
+{
+    using System;
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class StripeFileLinkCreateOptions : StripeFileLinkSharedOptions
+    {
+        /// <summary>
+        /// The ID of the file.
+        /// </summary>
+        [JsonProperty("file")]
+        public string FileId { get; set; }
+    }
+}
--- a/src/Stripe.net/Services/FileLinks/StripeFileLinkListOptions.cs
+++ b/src/Stripe.net/Services/FileLinks/StripeFileLinkListOptions.cs
@@ -1,0 +1,13 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+
+    public class StripeFileLinkListOptions : StripeListOptionsWithCreated
+    {
+        [JsonProperty("expired")]
+        public bool? Expired { get; set; }
+
+        [JsonProperty("file")]
+        public string FileId { get; set; }
+    }
+}
--- a/src/Stripe.net/Services/FileLinks/StripeFileLinkService.cs
+++ b/src/Stripe.net/Services/FileLinks/StripeFileLinkService.cs
@@ -1,0 +1,91 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using System.Net;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Stripe.Infrastructure;
+
+    public class StripeFileLinkService : StripeService
+    {
+        private static string classUrl = Urls.BaseUrl + "/file_links";
+
+        public StripeFileLinkService()
+            : base(null)
+        {
+        }
+
+        public StripeFileLinkService(string apiKey)
+            : base(apiKey)
+        {
+        }
+
+        public virtual StripeFileLink Create(StripeFileLinkCreateOptions createOptions, StripeRequestOptions requestOptions = null)
+        {
+            return Mapper<StripeFileLink>.MapFromJson(
+                Requestor.PostString(
+                    this.ApplyAllParameters(createOptions, classUrl, false),
+                    this.SetupRequestOptions(requestOptions)));
+        }
+
+        public virtual async Task<StripeFileLink> CreateAsync(StripeFileLinkCreateOptions createOptions, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return Mapper<StripeFileLink>.MapFromJson(
+                await Requestor.PostStringAsync(
+                    this.ApplyAllParameters(createOptions, classUrl, false),
+                    this.SetupRequestOptions(requestOptions),
+                    cancellationToken).ConfigureAwait(false));
+        }
+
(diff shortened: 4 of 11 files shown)


src/Stripe.net/Services/FileLinks/StripeFileLinkService.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        public StripeFileLinkService(string apiKey)
+            : base(apiKey)
+        {
+        }
+
+        public virtual StripeFileLink Create(StripeFileLinkCreateOptions createOptions, StripeRequestOptions requestOptions = null)
+        {
+            return CreateAsync(createOptions, requestOptions).Result;
```
ob-stripe: I've used a style I'd like to see us adopt for new services (and apply to the existing services):
1. Methods are in alphabetical order (like in tests)
2. Non-async methods simply call their async version + `.Result`, as suggested by @jaymedavis in #940.

src/Stripe.net/Services/FileLinks/StripeFileLinkUpdateOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class StripeFileLinkUpdateOptions : StripeBaseOptions, ISupportMetadata
+    {
+        /// <summary>
+        /// A future timestamp after which the link will no longer be usable.
+        /// </summary>
+        public DateTime? ExpiresAt { get; set; }
```
remi-stripe: you likely should have a `StripeFileLinkSharedOptions` class since they have 2 parameters in common. Not a huge deal though since it does work that way

src/Stripe.net/Services/FileLinks/StripeFileLinkListOptions.cs
```diff
@@ -0,0 +1,16 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+
+    public class StripeFileLinkListOptions : StripeListOptions
```
remi-stripe: Minor but there's a class called `StripeListOptionsWithCreated` that should either be used everywhere we have created or canned.

src/Stripe.net/Services/FileLinks/StripeFileLinkService.cs
remi-stripe: Are we sure it works across all .NET frameworks?

Also, if we approve this, should we convert the whole library once and for all?

jaymedavis: Thanks @ob-stripe for keeping things nice and neat and caring about the details. Warms my ❤️

src/Stripe.net/Services/FileLinks/StripeFileLinkService.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        public virtual async Task<StripeFileLink> CreateAsync(StripeFileLinkCreateOptions createOptions, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return Mapper<StripeFileLink>.MapFromJson(
+                await Requestor.PostStringAsync(
+                    this.ApplyAllParameters(createOptions, classUrl, false),
+                    this.SetupRequestOptions(requestOptions),
+                    cancellationToken).ConfigureAwait(false));
```
jaymedavis: I would format the last line (ConfigureAwait) like so:

```
return Mapper<StripeFileLink>.MapFromJson(
                 await Requestor.PostStringAsync(
                     this.ApplyAllParameters(createOptions, classUrl, false),
                     this.SetupRequestOptions(requestOptions),
                     cancellationToken
                 ).ConfigureAwait(false));
```

I like this JavaScript/jQuery style formatting practice. A quick glance shows where the function call ends imo

src/Stripe.net/Services/FileLinks/StripeFileLinkService.cs
jaymedavis: @remi-stripe The library almost [does this anyway](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Infrastructure/Requestor.cs#L63) on non async calls. Notice it uses the addition of `GetAwaiter().GetResult()`. I believe this was determined good practice from #1145. [This stack overflow post](https://stackoverflow.com/questions/17284517/is-task-result-the-same-as-getawaiter-getresult) sheds some light about the differences... which makes me wonder if [this change](https://github.com/stripe/stripe-dotnet/pull/1145/files#diff-f4f707e39cfd8d37180bd288044fc813R64) was necessary.

src/Stripe.net/Services/FileLinks/StripeFileLinkService.cs
remi-stripe: @jaymedavis We now use StyleCop for the coding styles and one of the rules requires that the last bit is on the same line as cancelation token so it would have to look like what @ob-stripe did.

src/Stripe.net/Services/FileLinks/StripeFileLinkUpdateOptions.cs
ob-stripe: Ack. I've updated the PR to use it.

ob-stripe:
I've reverted the sync/async thing for now. I still want to move forward with this, but this PR isn't the best place for it.

ob-stripe:
ptal @remi-stripe

src/Stripe.net/Entities/StripeFileUpload.cs
```diff
@@ -1,6 +1,7 @@
 ﻿namespace Stripe
 {
     using System;
+
```
remi-stripe: curious about that one, is it a convention you want to start respecting? I don't think it was in any of the StyleCop rules

src/Stripe.net/Services/FileLinks/StripeFileLinkCreateOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class StripeFileLinkCreateOptions : StripeFileLinkSharedOptions
+    {
+        /// <summary>
+        /// The ID of the file.
+        /// </summary>
+        [JsonProperty("file")]
+        public string File { get; set; }
```
remi-stripe: The convention in stripe-dotnet is to name things like this with `Id` at the end.

src/Stripe.net/Services/FileLinks/StripeFileLinkListOptions.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    using Newtonsoft.Json;
+
+    public class StripeFileLinkListOptions : StripeListOptionsWithCreated
+    {
+        [JsonProperty("expired")]
+        public bool? Expired { get; set; }
+
+        [JsonProperty("file")]
+        public string File { get; set; }
```
remi-stripe: same as above

src/Stripe.net/Services/FileLinks/StripeFileLinkCreateOptions.cs
ob-stripe: Nice catch. Fixed.

src/Stripe.net/Services/FileLinks/StripeFileLinkListOptions.cs
ob-stripe: 👍 Fixed.

src/Stripe.net/Entities/StripeFileUpload.cs
ob-stripe: It doesn't seem to conventional in C#, so I've reverted.

PR #1245, opened 2018-08-03
Add cancel support for topups

ob-stripe:
cc @stripe/api-libraries @foresthu-stripe 

Add cancel support for topups.

I've also alphabetized methods in `StripeTopupService`.

### DIFF
--- a/src/Stripe.net/Services/Topups/StripeTopupService.cs
+++ b/src/Stripe.net/Services/Topups/StripeTopupService.cs
@@ -15,35 +15,28 @@ namespace Stripe
 
         public bool ExpandSource { get; set; }
 
+        public virtual StripeTopup Cancel(string topupId, StripeRequestOptions requestOptions = null)
+        {
+            return Mapper<StripeTopup>.MapFromJson(
+                Requestor.PostString(
+                    this.ApplyAllParameters(null, $"{Urls.Topups}/{topupId}/cancel", false),
+                    this.SetupRequestOptions(requestOptions)));
+        }
+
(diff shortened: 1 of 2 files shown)


src/Stripe.net/Services/Topups/StripeTopupService.cs
```diff
@@ -15,35 +15,28 @@ public StripeTopupService(string apiKey = null)
...
         public bool ExpandSource { get; set; }
 
-        public virtual StripeTopup Create(StripeTopupCreateOptions createOptions, StripeRequestOptions requestOptions = null)
+        public virtual StripeTopup Cancel(string topupId, StripeRequestOptions requestOptions = null)
         {
             return Mapper<StripeTopup>.MapFromJson(
                 Requestor.PostString(
-                    this.ApplyAllParameters(createOptions, Urls.Topups, false),
+                    this.ApplyAllParameters(null, $"{Urls.Topups}/{topupId}/cancel", false),
                     this.SetupRequestOptions(requestOptions)));
```
remi-stripe: In other services like the Payout one we did
```
return this.Post($"{Urls.BaseUrl}/payouts/{payoutId}/cancel", requestOptions);
```
which is  bit more readable, Not sure what the difference is between `StripeService` and `StripeBasicService` though

ob-stripe: Ack. I'll leave this one as-is for now. Let's add this to the list of things that should be standardized throughout the library.

PR #1250, opened 2018-08-16
Missing and un-mapped properties

xrisdoc:
I noticed that there was no property for `UnitLabel` being mapped from `unit_label` on StripeProduct. So, I have added this in.

There was also no JSON mapping on StripeInvoice to map the `billing_reason` JSON property to the `BillingReason` property on the `StripeIncvoice` class. So, I have updated this to include the mapping for this too.


remi-stripe:
assigning to @brandur-stripe to release!

brandur-stripe:
Released as 17.9.0.

PR #1252, opened 2018-08-21
[WIP] Remove Stripe prefix on all API classes

remi-stripe:
cc @stripe/api-libraries


src/StripeTests/Services/Issuing/Cards/IssuingCardServiceTest.cs
```diff
@@ -6,7 +6,7 @@ namespace StripeTests
     using Stripe.Issuing;
     using Xunit;
 
-    public class CardServiceTest : BaseStripeTest
+    public class IssuingCardServiceTest : BaseStripeTest
```
remi-stripe: Flagging that there's a name conflict in the test suite as we already have `CardServiceTest`. For now, I just prefixed that one and the DisputeServiceTest one to avoid conflicts until we find a better solution

ob-stripe: We can probably reuse the namespacing in tests, e.g. we'd have `StripeTests.CardServiceTest` and `StripeTests.Issuing.CardServiceTest`.

remi-stripe:
r? @ob-stripe 

Can you have a thorough look on this one? I think it's mostly done, there are some classes I did not rename like `StripeEntity` or `StripeError` but everything else should be done.

There's also `Stripe3DSecure` not renamed because classes can't start with a 3. I could rename to ThreeDSecure though let me know what you think is best.

remi-stripe:
> Kind of difficult to be very thorough with such PRs

Yeah I hear you this is really tricky to review unfortunately. I'm hopeful that the test suite covers most edge-cases but I'm sure we'll have miss some things.

> Yeah, I think ThreeDSecure is preferable, and consistent with our other libraries.

Toyed with adding "deprecated" somewhere in the name but went with `ThreeDSecure` after all.

PTAL

ob-stripe:
`StripeReturn` should be renamed to `OrderReturn`.

Aside from that, all the remaining `StripeSomething` classes are high-level abstractions like `StripeEntity` or `StripeService`, so I think it's fine to keep the `Stripe` prefix for those.

remi-stripe:
huh nice catch. We already have OrderReturn so not yet sure what StripeReturn is but will fix

remi-stripe:
> StripeReturn should be renamed to OrderReturn.

Okay `OrderReturn` was never used so I just removed that one and renamed the other

ob-stripe:
@remi-stripe Looks like you didn't push that last change.

remi-stripe:
@ob-stripe I did I just forgot I had switched branch already to try something else 🤦‍♂️. Fixed, PTYAL.

Issue #1066, opened 2017-11-29
Clean up payment instruments code


ob-stripe:
The current implementation of payment instruments (i.e. cards, bank accounts, sources) has a lot of inconsistencies and is hard to maintain.

### Cards

- Entity: `StripeCard`

- API methods:

   - Creation:
       - On a customer: `StripeCardService.Create()` with `StripeCardCreateOptions`
       - On a recipient: `StripeCardService.Create()` with `StripeCreditCardOptions`
       - On an account: not possible, but you can pass a card when creating an account via `StripeAccountCardOptions`

    - Retrieval:
        - On a customer: `StripeCardService.Get()` with `isRecipient=false`
        - On a recipient: `StripeCardService.Get()` with `isRecipient=true`
        - On an account: not possible, but you can access the `ExternalAccounts` attribute which is a `StripeList<Source>` where `Source` is a wrapper over `StripeCard` / `StripeBankAccount` / `StripeDeleted`

    - Update:
        - On a customer: `StripeCardService.Update()` with `StripeCardUpdateOptions` and `isRecipient=false`
        - On a recipient: `StripeCardService.Update()` with `StripeCardUpdateOptions` and `isRecipient=true`
        - On an account: not possible

    - Delete:
        - On a customer: `StripeCardService.Delete()` with `isRecipient=false`
        - On a recipient: `StripeCardService.Delete()` with `isRecipient=true`
        - On an account: not possible

    - List:
        - On a customer: `StripeCardService.List()` with `StripeListOptions` and `isRecipient=false`
            - **NB**: currently this will hit `/v1/customers/{CUSTOMER_ID}/sources` which will return _all_ payment sources, and non-cards will be incorrectly deserialized. Cf. #989
        - On a recipient: `StripeCardService.List()` with `StripeListOptions` and `isRecipient=true`
        - On an account: not possible, but you can access the `ExternalAccounts` attribute (cf. "Retrieval" above)

### Bank accounts

- Entities:
    - `CustomerBankAccount` when the bank account is a payment source on a customer
    - `StripeBankAccount` when the bank account is an external account on a custom account

- API methods:

    - Creation:
        - On a customer: `BankAccountService.Create()` with `BankAccountCreateOptions`
        - On an account: not possible, but you can pass a bank account when creating an account via `StripeAccountBankAccountOptions`

    - Retrieval:
        - On a customer: `BankAccountService.Get()`
[...]

PR #1253, opened 2018-08-21
Properly re-architecture the library to handle various source abstractions

remi-stripe:
Today, the library has multiple separate classes that "abstract" away the complexity in the API around payment sources and payout destinations.

This is partly described in details in https://github.com/stripe/stripe-dotnet/issues/1066 though it was written before External Accounts existed.

This PR will attempt to rewrite or merge classes so that we end up with only three concepts:
* `ExternalAccount` representing the objects where funds can be paid out
* `PaymentSource` representing the objects acting as a Source on a Charge
* `Source` the existing API resource which would be a subset of `PaymentSource`

cc @stripe/api-libraries 

- [x] Remove the Stripe prefix from `StripeBankAccount` and `StripeExternalAccount` resources
- [x] A `Payout` destination should be an `ExternalAccount`
- [x] Get rid of `CustomerBankAccount` and move everything over to StripeBankAccount
- [x] Add tests for `ExternalAccount`
- [x] Remove Recipient-related operations on `StripeCardService`. Those have almost never been used in a few months and this has been deprecated for over 3 years now.
- [ ] ~Figure out a way to merge `StripeCardCreateOptions` and `StripeCreditCardOptions`~. We-re going to punt on this for now until we can support nested options properly (sharing parameter names under a different parent parameter name)


remi-stripe:
r? @ob-stripe 

Make sure to **not** merge this one as we need to do more work first.

remi-stripe:
cc @stripe/api-libraries

src/Stripe.net/Entities/PaymentSource.cs
```diff
@@ -3,7 +3,7 @@
     using Newtonsoft.Json;
     using Stripe.Infrastructure;
 
-    public enum SourceType
+    public enum PaymentSourceType
     {
         Account,
```
ob-stripe: Not a blocker for this PR, but we talked about adding an `Unknown` value in the enum, to avoid defaulting to `Account` in case the API returns a type that the library doesn't know about.

src/Stripe.net/Entities/ExternalAccount.cs
```diff
@@ -11,14 +11,14 @@ public enum ExternalAccountType
     }
```
ob-stripe: Same here, we should add an `Unknown` value at the start of the enum.

src/Stripe.net/Services/Cards/StripeCardService.cs
```diff
@@ -107,9 +86,9 @@ public virtual async Task<StripeCard> CreateAsync(string recipientId, StripeCred
                     cancellationToken).ConfigureAwait(false));
         }
 
-        public virtual async Task<StripeCard> GetAsync(string customerOrRecipientId, string cardId, bool isRecipient = false, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
+        public virtual async Task<StripeCard> GetAsync(string customerId, string cardId, bool isRecipient = false, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
```
ob-stripe: `isRecipient` should be removed from the list of arguments.

src/StripeTests/Entities/BankAccountTest.cs
```diff
@@ -8,31 +8,42 @@ namespace StripeTests
...
             Assert.NotNull(bankAccount);
-            Assert.IsType<CustomerBankAccount>(bankAccount);
+            Assert.IsType<BankAccount>(bankAccount);
             Assert.NotNull(bankAccount.Id);
             Assert.Equal("bank_account", bankAccount.Object);
         }
 
         [Fact]
-        public void DeserializeWithExpansions()
+        public void DeserializeWithExpansionsForCustomer()
```
ob-stripe: Feel free to ignore for this PR, but `Account` should also be expandable on `BankAccount`.

remi-stripe: Ah yes that's because BankAccount is not customer specific anymore. FWIW I added the tests in `ExternalAccount` instead as I think it makes more sense in general. I also had to add support for `Account` on `BankAccount`.

remi-stripe:
@ob-stripe Fixed some and replied to others. PTAL

Issue #1254, opened 2018-08-22
Bug StripeInvoiceItemCreateOptions


Klynk:
Hello,

I am having trouble with the library, due to what seems to be a bug.

https://stripe.com/docs/api#create_invoiceitem

Here you see 'Amount' is an optional property.

But for your source file, it is not:

https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/InvoiceItems/StripeInvoiceItemCreateOptions.cs

`[JsonProperty("amount")]
public int Amount { get; set; }`

It should be changed to:

`[JsonProperty("amount")]
public int? Amount { get; set; }`

I am having trouble setting UnitAmount because of this, as it throws this exception:
Stripe.StripeException: 'You may only specify one of these parameters: amount, unit_amount.'

OS: Windows 10 (17134.228)
.NET Version: Microsoft.AspNetCore.All (v2.1.2)
Library Version: Stripe.net (v17.10.0)

ob-stripe:
Hi @Klynk, thanks a lot for the detailed report! I think your analysis is spot on. We'll release a fix promptly.

ob-stripe:
Fixed in 17.11.0.

Klynk:
Thank you, it works now.

PR #1255, opened 2018-08-22
Amount for invoice item is now optional

remi-stripe:
This fixes https://github.com/stripe/stripe-dotnet/issues/1254

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Released as 17.11.0.

PR #1257, opened 2018-08-23
Added support for options on the Pay Invoice API

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Services/Invoices/StripeInvoiceService.cs
+++ b/src/Stripe.net/Services/Invoices/StripeInvoiceService.cs
@@ -49,11 +49,11 @@
                     this.SetupRequestOptions(requestOptions)));
         }
 
-        public virtual StripeInvoice Pay(string invoiceId, StripeRequestOptions requestOptions = null)
+        public virtual StripeInvoice Pay(string invoiceId, StripeInvoicePayOptions payOptions, StripeRequestOptions requestOptions = null)
         {
             return Mapper<StripeInvoice>.MapFromJson(
                 Requestor.PostString(
-                    this.ApplyAllParameters(null, $"{Urls.Invoices}/{invoiceId}/pay", false),
+                    this.ApplyAllParameters(payOptions, $"{Urls.Invoices}/{invoiceId}/pay", false),
                     this.SetupRequestOptions(requestOptions)));
         }
 
@@ -122,11 +122,11 @@
                     cancellationToken).ConfigureAwait(false));
         }
 
-        public virtual async Task<StripeInvoice> PayAsync(string invoiceId, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
+        public virtual async Task<StripeInvoice> PayAsync(string invoiceId, StripeInvoicePayOptions payOptions, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
             return Mapper<StripeInvoice>.MapFromJson(
                 await Requestor.PostStringAsync(
-                    this.ApplyAllParameters(null, $"{Urls.Invoices}/{invoiceId}/pay", false),
+                    this.ApplyAllParameters(payOptions, $"{Urls.Invoices}/{invoiceId}/pay", false),
                     this.SetupRequestOptions(requestOptions),
                     cancellationToken).ConfigureAwait(false));
         }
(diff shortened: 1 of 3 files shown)


src/Stripe.net/Services/Invoices/StripeInvoiceService.cs
```diff
@@ -50,10 +50,15 @@ public virtual StripeInvoice Update(string invoiceId, StripeInvoiceUpdateOptions
         }
 
         public virtual StripeInvoice Pay(string invoiceId, StripeRequestOptions requestOptions = null)
+        {
+            return this.Pay(invoiceId, null, requestOptions);
+        }
+
+        public virtual StripeInvoice Pay(string invoiceId, StripeInvoicePayOptions payOptions, StripeRequestOptions requestOptions = null)
```
remi-stripe: I did this to avoid a breaking change but maybe it'd be cleaner to force the option?

src/Stripe.net/Services/Invoices/StripeInvoiceService.cs
```diff
@@ -131,6 +136,15 @@ public virtual async Task<StripeInvoice> PayAsync(string invoiceId, StripeReques
                     cancellationToken).ConfigureAwait(false));
         }
 
+        public virtual async Task<StripeInvoice> PayAsync(string invoiceId, StripeInvoicePayOptions payOptions, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
```
remi-stripe: It seems you an not call PayAsync with the option from the one without, it returns a compilation error

```

Services/Invoices/StripeInvoiceService.cs(132,20): error CS4016: Since this is an async method, the return expression must be of type 'StripeInvoice' rather than 'Task<StripeInvoice>' [/Users/remi/Workspace/bindings/stripe-dotnet/src/Stripe.net/Stripe.net.csproj]
```

ob-stripe: Can you change the first `PayAsync` method to call this one?

src/StripeTests/Services/Invoices/StripeInvoiceServiceTest.cs
```diff
@@ -151,6 +158,14 @@ public void Pay()
             Assert.Equal("invoice", invoice.Object);
         }
 
+        [Fact]
+        public void PayOptions()
```
ob-stripe: I don't think we need separate tests for both methods, esp. if one simply calls the other. Can you just update the `Pay` test to add options?

src/Stripe.net/Services/Invoices/StripeInvoiceService.cs
ob-stripe: You want `return await PayAsync(...)`;

remi-stripe: damn I tried a few variations but did not think it just wanted the `await`. Thanks!

remi-stripe:
Fixed PTAL @ob-stripe

src/Stripe.net/Services/Invoices/StripeInvoiceService.cs
```diff
@@ -123,10 +128,15 @@ public virtual async Task<StripeInvoice> UpdateAsync(string invoiceId, StripeInv
         }
 
         public virtual async Task<StripeInvoice> PayAsync(string invoiceId, StripeRequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return await this.PayAsync(invoiceId, null, requestOptions, cancellationToken);
```
ob-stripe: Actually, I did some research and the right approach here is not declare this method as `async` and not use `await`. Can you update accordingly? Sorry for the extra work :/

remi-stripe: Hmmm, are you sure that method will still appear as Async though and work as expected?

ob-stripe: On third thought, it would work, but it would be a breaking change :/

Let's just get rid of the overloads and release this as a major version?

remi-stripe: Done!

remi-stripe:
@ob-stripe fixed, PTAL

ob-stripe:
Released as 18.0.0.

PR #1259, opened 2018-08-23
Upgrade to the latest API version 2018-08-23

remi-stripe:
This PR has multiple major changes due to the move to the latest API version and some new properties/breaking changes. I tried to split the logic in smaller commits to make it easier to review

* Add support for `tax_info` and `tax_info_verification` on the customer
* Remove obsolete subscription service as we're making changes to the prototypes
* Remove `business_vat_id` on the Customer
* Rename `amount` to `unit_amount` on Plan's tiers
* Remove the ability to call Cancel on a Subscription with `at_period_end`. This involved some changes to the prototypes. For that reason I added a new options class as we will soon add parameters on subscription cancelation and it avoids the second major change. I also took this as an opportunity to remove the prototype with plan and customer. It did not make sense to keep supporting this and to remove taking `customer` as a method parameter. The options classes have `CustomerId` for that purpose.
* `percent_off` on Coupons is now a decimal
* Products do not that the list of SKUs embedded anymore.

r? @ob-stripe 
cc @stripe/api-libraries


src/Stripe.net/Services/Subscriptions/StripeSubscriptionService.cs
```diff
@@ -61,13 +48,12 @@ public virtual StripeSubscription Update(string subscriptionId, StripeSubscripti
...
 
-        public virtual StripeSubscription Cancel(string subscriptionId, bool cancelAtPeriodEnd = false, StripeRequestOptions requestOptions = null)
+        public virtual StripeSubscription Cancel(string subscriptionId, StripeSubscriptionCancelOptions cancelOptions, StripeRequestOptions requestOptions = null)
         {
-            var url = string.Format(Urls.Subscriptions + "/{0}", subscriptionId);
-            url = ParameterBuilder.ApplyParameterToUrl(url, "at_period_end", cancelAtPeriodEnd.ToString());
-
             return Mapper<StripeSubscription>.MapFromJson(
-                Requestor.Delete(url, this.SetupRequestOptions(requestOptions)));
+                Requestor.Delete(
```
remi-stripe: I am a _little_ bit worried about this line. I tried testing it fully and I think it works but I could not force stripe-mock to reject my options (When sending fake parameters). It either means it does not validate GET params properly (my guess) or that something truncates the GET parameters when sent with Delete. I have spent a while on this and need a second pair of eyes to make sure.

remi-stripe: Flagging that with the latest stripe-mock I have been able to confirm the params would be sent properly so it seems to work fine!

src/Stripe.net/Services/Coupons/StripeCouponCreateOptions.cs
```diff
@@ -17,7 +17,7 @@ public class StripeCouponCreateOptions : StripeBaseOptions, ISupportMetadata
         public string Currency { get; set; }
 
         [JsonProperty("percent_off")]
-        public int? PercentOff { get; set; }
+        public decimal? PercentOff { get; set; }
```
ob-stripe: This was always a decimal and the type was just incorrect in the .NET library, correct?

remi-stripe: No it was always an integer until API version [2018-07-27](https://stripe.com/docs/upgrades#2018-07-27)

remi-stripe:
@ob-stripe done!

Issue #1265, opened 2018-08-29
Error on customer get with subscription coupon


kawaicheung:
I'm receiving this error when calling `StripeCustomerService.Get()`:
```
JsonReaderException: Input string '40.0' is not a valid integer. Path 'subscriptions.data[0].discount.coupon.percent_off', line 108, position 31.
```
I am using v19.1.0 and .net 4.5 framework and using the latest API version (2018-08-23)

In my customer's subscription, they have a 40% discount coupon associated to their subscription. I know there was a recent int -> decimal upgrade for coupons in this version, but looks like the .NET API needs the type update from `int?` to `decimal?` in `Stripe.StripeCoupon`.

Thanks!
-KC

remi-stripe:
@kawaicheung thanks a lot for the detailed report and sorry about the bug on that one, that's my bad. I just opened a PR to fix this and we'll try to release it shortly.

PR #1266, opened 2018-08-29
Fix the Coupon resource as percent_off is now a decimal

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/1265

While this is a breaking change, it's currently broken in the current version so I think a minor would be fine.

I wonder if we can/should pull the broken version in this case (and in related cases)

r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
LGTM! Thanks for the quick fix Remi.

brandur-stripe:
(And agreed on minor release.)

Released as 19.2.0.

PR #1268, opened 2018-09-03
Add support for mandate/receiver hash on source creation/update

remi-stripe:
This mirrors what we did for stripe-go here: https://github.com/stripe/stripe-go/pull/686

I think it's worth keeping the extra parameters in the tests as it gives a good example of passing a date and just making sure that does not break in the future when we fix the nested options approach.

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Released as 19.3.0.

Issue #1260, opened 2018-08-24
Extract sync/async interfaces from Stripe services


SeanFarrow:
It would be nice if we could use stripe services in unit tests. I'm happy using stripe-mock in integration tests, but the way things are currently mean I'll have to abstract the stripe services if I want to unit test them.

In the project I'm working on it is mandatory that we test as much as is possible.

remi-stripe:
@SeanFarrow You can write unit tests using the real API instead of stripe-mock without issues. This is not something we have blocked you from doing today and using the normal methods would let you hit the API as expected.

I'm likely misunderstanding what is blocking you though. If that's the case, could you provide some examples of what does not work for you?

SeanFarrow:
@ remi-stripe,

Whilst I could hit the API without using stripe-mock, I don’t really want to do that in unit tests, as this then elevates them to integration tests.
I’m looking to mock out the stripe services to ensure they are called and the return values I expect are handled using something like NSubstitute. Not to mention this, this would then help dependency injection.
Does that clarify things?

remi-stripe:
@SeanFarrow Okay I see what you mean now. I do see how it could be useful but I don't think it's something we'd ship in the near future. It's not something we have in any of the other libraries either today. I'll flag it as Future though in case it's something we could tackle in the future.

SeanFarrow:
@ remi-stripe,

It’s not that hard, I’m happy to do the work to save you doing it yourself, if a pr would be accepted.
Would probably take me a couple of days max.

remi-stripe:
@SeanFarrow We're in the middle of a pretty big refactor of the entire library so it might be worth waiting a few weeks. You can follow along in https://github.com/stripe/stripe-dotnet/pull/1253 and https://github.com/stripe/stripe-dotnet/pull/1252

We would definitely be interested in a PR though! What I'd recommend is to make a simple proof of concept on one service (or even one method) to showcase what you're aiming for and make sure it's the right direction. That way you don't do all the work for nothing

Let me know what you think!

SeanFarrow:
@remi-stripe,

That timescale works for me as I don't need this straight away.

I'll create interfaces from a single service as a proof of concept, do you have any suggestions as to what the best one would be?

remi-stripe:
@sean-stripe I'd use one of the most common ones like Customer or Charge!

SeanFarrow:
@ remi-stripe,

Customers looks a goodbet for a proof of concept.
[...]

PR #1269, opened 2018-09-04
Service interfaces

ob-stripe:
cc @stripe/api-libraries 

Just an example of what interfaces for Stripe services could look like. "Fixes" #1260.

@SeanFarrow Is what you had in mind something like this?

### DIFF
--- a/src/Stripe.net/Services/UsageRecordSummaries/UsageRecordSummaryService.cs
+++ b/src/Stripe.net/Services/UsageRecordSummaries/UsageRecordSummaryService.cs
@@ -5,7 +5,8 @@ namespace Stripe
     using System.Threading.Tasks;
     using Stripe.Infrastructure;
 
-    public class UsageRecordSummaryService : BasicService<UsageRecord>
+    public class UsageRecordSummaryService : BasicService<UsageRecord>,
+        INestedListable<UsageRecordSummary, UsageRecordSummaryListOptions>
     {
         public UsageRecordSummaryService()
             : base(null)
--- a/src/Stripe.net/Services/_interfaces/INestedCreatable.cs
+++ b/src/Stripe.net/Services/_interfaces/INestedCreatable.cs
@@ -1,0 +1,14 @@
+namespace Stripe
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public interface INestedCreatable<T, O>
+        where T : StripeEntity
+        where O : BaseOptions
+    {
+        T Create(string parentId, O createOptions, RequestOptions requestOptions = null);
+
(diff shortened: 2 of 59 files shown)


brandur-stripe:
(Nice — this seems like a really nice idea to me.)

SeanFarrow:
@ob-stripe,
Yes, this is exactly what I had in mind!
Would you like me to work on converting the other interfaces based on this?

ob-stripe:
@SeanFarrow Thanks for the offer! However, I'm fairly sure that some of the services have method signatures that will be incompatible with the ones defined in the interfaces, and I'd rather we make the calls to change those (if necessary) ourselves since they'll likely be breaking changes.

I'll try to work on this this week.

ob-stripe:
I rebased the PR on the integration branch for the upcoming big release, and added interfaces wherever possible. It actually went pretty smoothly -- I thought there would be a lot more inconsistencies!

One big remaining todo: handle nested resources. We'd need new interfaces, e.g. `INestedCreatable`, with similar signatures except that there's an additional `parentResourceId` argument.

SeanFarrow:
Is this likely to happen for the big release, or later?

remi-stripe:
@SeanFarrow We will try to make this part of the next major version!

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries

src/Stripe.net/Services/UsageRecordSummaries/UsageRecordSummaryService.cs
```diff
@@ -5,7 +5,8 @@ namespace Stripe
     using System.Threading.Tasks;
     using Stripe.Infrastructure;
 
-    public class UsageRecordSummaryService : BasicService<UsageRecord>
+    public class UsageRecordSummaryService : BasicService<UsageRecord>,
+        INestedListable<UsageRecordSummary, UsageRecordSummaryListOptions>
```
remi-stripe: Are you sure this one is correct? This service is weird because it does not take the record id as a first parameter. It likely should though and we should fix it but it makes me think this current setup would not work?

src/Stripe.net/Services/_interfaces/INestedCreatable.cs
```diff
@@ -0,0 +1,14 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public interface INestedCreatable<T, O>
+        where T : StripeEntity
+        where O : BaseOptions
+    {
+        T Create(string parentId, O createOptions, RequestOptions requestOptions = null);
+
+        Task<T> CreateAsync(string parentId, O createOptions, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken));
```
remi-stripe: Reading this makes me think it should have failed for UsageRecordService after all. Is there a bug somewhere that does not raise an error if the interface does not have a defined method?

PR #1272, opened 2018-09-05
Add support for Reporting resources

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries @brahn-stripe

### DIFF
--- a/src/Stripe.net/Entities/Reporting/ReportRun.cs
+++ b/src/Stripe.net/Entities/Reporting/ReportRun.cs
@@ -1,0 +1,39 @@
+namespace Stripe.Reporting
+{
+    using System;
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class ReportRun : StripeEntityWithId
+    {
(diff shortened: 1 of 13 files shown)


src/Stripe.net/Entities/Reporting/ReportRun.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        [JsonProperty("created")]
+        [JsonConverter(typeof(StripeDateTimeConverter))]
+        public DateTime Created { get; set; }
+
+        [JsonProperty("error")]
+        public string Error { get; set; }
+
+        [JsonProperty("livemode")]
+        public bool LiveMode { get; set; }
```
ob-stripe: Ugh, I don't like that the capitalization doesn't match the snake_case -> CamelCase rule here, but it's consistent with every other resource, so let's keep it as is.

remi-stripe: yup I think that every time I do a new resource. Adding to the list of things to change in the next major revision

ob-stripe:
Released as 19.4.0.

PR #1274, opened 2018-09-08
added brand and last4

pradeepannepu:


pradeepannepu:
thanks for the update @ob-stripe.
But we do receive the fields even with sources, we are making use of those two columns and unable to read through SDK, but they are available when we parse the ResponseJson.

```
 "three_d_secure": {
    "card": "src_1D98WfIlmURI1pIsfU3oTlEI",
    "brand": "Visa",
    "card_automatically_updated": false,
    "country": "US",
    "cvc_check": "unchecked",
    "exp_month": 4,
    "exp_year": 2024,
    "fingerprint": "mSnFOnHKG5jhhGdA",
    "funding": "credit",
    "last4": "3063",
    "three_d_secure": "required",
    "customer": null,
    "authenticated": false,
    "name": null,
    "address_line1_check": null,
    "address_zip_check": null,
    "tokenization_method": null,
    "dynamic_last4": null
  },
```

ob-stripe:
@pradeepannepu Apologies, looks like I was doubly mistaken: you're using the sources API, not the old 3DS API, and you did update the right file. Sorry about that!

I'll open another PR and add all the other missing attributes while I'm at it.

ob-stripe:
@pradeepannepu I just released version 19.6.0, which includes accessors for all attributes on the `StripeThreeDSecure` model class. It will be available on NuGet shortly.

PR #1279, opened 2018-09-11
Add missing attributes on StripeThreeDSecure

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries @pradeepannepu

Adds all missing attributes on the `StripeThreeDSecure` model class (including gated ones).

Replaces #1274.


src/Stripe.net/Entities/Sources/StripeThreeDSecure.cs
```diff
@@ -4,13 +4,70 @@
...
+        public int ExpYear { get; set; }
+
+        [JsonProperty("fingerprint")]
+        public string Fingerprint { get; set; }
+
+        [JsonProperty("funding")]
+        public string Funding { get; set; }
+
+        [JsonProperty("google_reference")]
+        public string GoogleReference { get; set; }
```
remi-stripe: I'm not sure what this is but it seems more internal than external I would not surface. I feel the same about IIN + Issuer + Description. It's gated and while it's supported on the Source resource I'm not convinced we should surface it here

ob-stripe: `IIN`, `Issuer` and `Description` are already available on `StripeCard` though.

ob-stripe: I removed `GoogleReference`.

remi-stripe: I know they are visible but I wanted to avoid leaking them further. Can you at least do what we do on the other resource: and group them with a comment, or add a comment to each one about being gated?

ob-stripe:
I added XML comments for all attributes. ptal @remi-stripe

src/Stripe.net/Entities/Sources/StripeThreeDSecure.cs
```diff
@@ -4,13 +4,124 @@
...
+        /// The card used to create this 3DS source.
+        /// </summary>
         [JsonProperty("card")]
         public string CardId { get; set; }
 
+        /// <summary>
+        /// Whether the card was automatically updated at some point by Stripe or not.
+        /// </summary>
+        [JsonProperty("card_automatically_updated")]
+        public bool CardAutomaticallyUpdated { get; set; }
```
remi-stripe: This is purely internal and should not be here (sorry only seeing it now)

ob-stripe: I'll remove it, but it is returned by the public API though (cf. the sample JSON in #1274)

remi-stripe: I know, but I also know it should not and that we flagged it and that it should have been fixed already.

ob-stripe:
ptal @remi-stripe

PR #1280, opened 2018-09-11
Enable XML documentation

### LINKED ISSUE #1261: No code comment or docs in lib
We just starting using stripe and noticed theres nearly no documentation, just as a suggestion or enhancement it would be really helpful if you at least provided some code comments within the library more specifically the .net lib even if only on top of methods etc.

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Generates the XML documentation file. Even though few of our classes are properly documented, it's better than nothing.

Partially fixes #1261.

### DIFF
--- a/src/Stripe.net/Entities/_contracts/ISupportMetadata.cs
+++ b/src/Stripe.net/Entities/_contracts/ISupportMetadata.cs
@@ -3,7 +3,7 @@ namespace Stripe
     using System.Collections.Generic;
 
     /// <summary>
-    /// Interface that identifies entities with a Metadata property of type <see cref="Dictionary{string,string}" />.
+    /// Interface that identifies entities with a Metadata property of type <see cref="Dictionary{String, String}" />.
     /// </summary>
     public interface ISupportMetadata
     {
--- a/src/Stripe.net/Services/Charges/StripeChargeCaptureOptions.cs
+++ b/src/Stripe.net/Services/Charges/StripeChargeCaptureOptions.cs
@@ -39,7 +39,14 @@ namespace Stripe
         public string ReceiptEmail { get; set; }
 
         /// <summary>
-        /// An arbitrary string to be displayed on your customer's credit card statement. This may be up to 22 characters. As an example, if your website is RunClub and the item you're charging for is a race ticket, you may want to specify a statement_descriptor of RunClub 5K race ticket. The statement description may not include <>"' characters, and will appear on your customer's statement in capital letters. Non-ASCII characters are automatically stripped. While most banks display this information consistently, some may display it incorrectly or not at all.
+        /// An arbitrary string to be displayed on your customer's credit card statement. This may
+        /// be up to 22 characters. As an example, if your website is <c>RunClub</c> and the item
+        /// you're charging for is a race ticket, you may want to specify a
+        /// <c>statement_descriptor</c> of <c>RunClub 5K race ticket</c>. The statement description
+        /// may not include <c>&lt;&gt;"'</c> characters, and will appear on your customer's
+        /// statement in capital letters. Non-ASCII characters are automatically stripped. While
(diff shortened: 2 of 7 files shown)


src/Stripe.net/Entities/_contracts/ISupportMetadata.cs
```diff
@@ -3,7 +3,7 @@ namespace Stripe
     using System.Collections.Generic;
 
     /// <summary>
-    /// Interface that identifies entities with a Metadata property of type <see cref="Dictionary{string,string}" />.
+    /// Interface that identifies entities with a Metadata property of type <see cref="Dictionary{String, String}" />.
```
remi-stripe: Why did this change? We explicitly use `string` and not `String` in the library

src/Stripe.net/Services/Charges/StripeChargeCaptureOptions.cs
```diff
@@ -39,7 +39,14 @@ public class StripeChargeCaptureOptions : StripeBaseOptions
...
+        /// An arbitrary string to be displayed on your customer's credit card statement. This may
+        /// be up to 22 characters. As an example, if your website is <c>RunClub</c> and the item
+        /// you're charging for is a race ticket, you may want to specify a
+        /// <c>statement_descriptor</c> of <c>RunClub 5K race ticket</c>. The statement description
+        /// may not include <c>&lt;&gt;"'</c> characters, and will appear on your customer's
+        /// statement in capital letters. Non-ASCII characters are automatically stripped. While
+        /// most banks display this information consistently, some may display it incorrectly or not
+        /// at all.
```
remi-stripe: Is there a reason you broke this one up in multiple lines in this specific PR?

src/Stripe.net/Entities/_contracts/ISupportMetadata.cs
ob-stripe: Not sure why, but the compiler was complaining about the original version. `string` is an alias in `System.String`, so I don't think this makes any difference.

remi-stripe: Sure but it seemed weird to have to change this when the compiler is the one complaining if we use `String` everywhere else.

src/Stripe.net/Services/Charges/StripeChargeCaptureOptions.cs
ob-stripe: The actual issue I fixed is that I replaced `<>` with `&lt;&gt;`, otherwise the compiler would complain about invalid XML.

Breaking up the comment in multiple lines was just for enhanced legibility.

ob-stripe:
> How did you test/confirm this change will do what you expect?

I verified that building the project generates a `Stripe.net.xml` file with the documentation. I assume that the file will automatically be included next time we push a version to NuGet, but that's unfortunately hard to test in advance.

remi-stripe:
Gotcha, I think that makes sense, approving!

src/Stripe.net/Entities/_contracts/ISupportMetadata.cs
ob-stripe: ¯\\\_(ツ)\_/¯ I don't know enough about C# type system to say for sure, but I assume XML comments need to use actual class names and not aliases.

xt0rted:
Confirming this is showing in the NuGet package now.

### 19.5.0
image

### 19.6.0
image

xt0rted:
>I assume that the file will automatically be included next time we push a version to NuGet, but that's unfortunately hard to test in advance.

You can open the NuGet package from the CI build with [NuGet Package Explorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer) if you ever need to test things like that.

ob-stripe:
\o/ Awesome, thanks @xt0rted!

PR #1282, opened 2018-09-12
Rename event to match type name and add a related one.

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Constants/StripeEvents.cs
+++ b/src/Stripe.net/Constants/StripeEvents.cs
@@ -452,7 +452,12 @@
         /// <summary>
         /// Occurs whenever a source transaction is created.
         /// </summary>
-        public const string SourceTransactionCanceled = "source.transaction.created";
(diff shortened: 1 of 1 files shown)


src/Stripe.net/Constants/StripeEvents.cs
```diff
@@ -452,7 +452,12 @@ public static class StripeEvents
         /// <summary>
         /// Occurs whenever a source transaction is created.
         /// </summary>
-        public const string SourceTransactionCanceled = "source.transaction.created";
+        public const string SourceTransactionCreated = "source.transaction.created";
+
+        /// <summary>
+        /// Occurs whenever a source transaction is created.
```
ob-stripe: Should be "is updated".

remi-stripe: damn it, fixed

remi-stripe:
@ob-stripe PTAL (also go to sleep!)

PR #1283, opened 2018-09-13
 Handle `DateTime`s directly in the encoder

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Handle `DateTime`s directly in the encoder, and update all `DateTime`s fields in options classes so that the conversion to long/string is no longer done internally (except for those that also have special values like `"now"`).


ob-stripe:
> I'm assuming that trial_end is left. What would be the right approach to handle it?

Off the top of my head, we could define a special class `DateTimeOrNow` that could take care of its own encoding via its `ToString` method, or add a new case for the class in `RequestStringBuilder`. I'm not sure it's worth it if it's just used in one or two places though.

PR #1285, opened 2018-09-13
Rename two properties that start with a lower case letter

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Nice! You can probably remove the line for `SA1300` from `StyleCopRules.ruleset` too.

remi-stripe:
Damn it I had completely forgotten about that rule. I knew I had logged which properties to fix somewhere but did not think to look there.

Fixed, PTAL!

PR #1286, opened 2018-09-17
Fix ScheduledQueryRun naming

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

- Renames `ScheduledQuery*` to `ScheduledQueryRun*`.
- Creates a `Stripe.Sigma` namespace for `ScheduledQueryRun` entity / service / options classes
- Renames files accordingly

### DIFF
--- a/src/Stripe.net/Services/Sigma/ScheduledQueryRuns/ScheduledQueryRunService.cs
+++ b/src/Stripe.net/Services/Sigma/ScheduledQueryRuns/ScheduledQueryRunService.cs
@@ -1,17 +1,17 @@
-﻿namespace Stripe
+﻿namespace Stripe.Sigma
 {
     using System.Threading;
     using System.Threading.Tasks;
     using Stripe.Infrastructure;
 
-    public class ScheduledQueryService : BasicService<ScheduledQueryRun>
+    public class ScheduledQueryRunService : BasicService<ScheduledQueryRun>
     {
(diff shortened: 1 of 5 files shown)


ob-stripe:
Added another commit to remove the unnecessary `cancelationToken` argument from the non-async methods, and reorganize the service class a bit. PTAL @remi-stripe

src/Stripe.net/Services/Sigma/ScheduledQueryRuns/ScheduledQueryRunService.cs
```diff
@@ -0,0 +1,41 @@
...
+            : base(null)
+        {
+        }
+
+        public ScheduledQueryRunService(string apiKey)
+            : base(apiKey)
+        {
+        }
+
+        public virtual ScheduledQueryRun Get(string queryRunId, RequestOptions requestOptions = null)
```
remi-stripe: Flagging that in the breaking change we likely should add an optional options class for all methods so that if we ever need to add options it's not a breaking change?

ob-stripe: I'm not necessarily opposed to the idea, but the reason we did this in stripe-java to support expansions in retrieve requests. In stripe-dotnet we already support this via the `ExpandFoo` fields.

PR #1287, opened 2018-09-17
Fix signatures in `InvoiceService` and `RefundService`

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Fixes signatures in `InvoiceService` and `RefundService` so that  required parameters are passed in options classes rather than as their own arguments.

Also a bunch of cleanup:
- organize fields / methods alphabetically
- updated fields in `UpcomingInvoiceOptions`
- added XML comments

### DIFF
--- a/src/Stripe.net/Services/Invoices/InvoiceCreateOptions.cs
+++ b/src/Stripe.net/Services/Invoices/InvoiceCreateOptions.cs
@@ -7,17 +7,33 @@
 
     public class InvoiceCreateOptions : BaseOptions, ISupportMetadata
     {
+        /// <summary>
+        /// A fee in cents that will be applied to the invoice and transferred to the application
+        /// owner’s Stripe account. The request must be made with an OAuth key or the Stripe-Account
+        /// header in order to take an application fee. For more information, see the application
+        /// fees <see href="https://stripe.com/docs/connect/subscriptions#working-with-invoices">documentation</see>.
+        /// </summary>
         [JsonProperty("application_fee")]
         public int? ApplicationFee { get; set; }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Stripe.net/Services/Invoices/InvoiceCreateOptions.cs
```diff
@@ -7,17 +7,33 @@
...
+        /// this invoice using the default source attached to the customer. When sending an invoice,
+        /// Stripe will email this invoice to the customer with payment instructions. Defaults to
+        /// <c>charge_automatically</c>.
         /// </summary>
         [JsonProperty("billing")]
         public Billing? Billing { get; set; }
 
         /// <summary>
-        /// The number of days from which the invoice is created until it is due. Only valid for invoices where billing=send_invoice.
+        /// REQUIRED
```
remi-stripe: Not a fan of the `REQUIRED` part either

src/Stripe.net/Services/Invoices/InvoiceService.cs
```diff
@@ -23,38 +23,38 @@ public InvoiceService(string apiKey)
...
         public bool ExpandSubscription { get; set; }
 
-        public virtual Invoice Get(string invoiceId, RequestOptions requestOptions = null)
+        public virtual Invoice Create(InvoiceCreateOptions createOptions = null, RequestOptions requestOptions = null)
         {
             return Mapper<Invoice>.MapFromJson(
-                Requestor.GetString(
-                    this.ApplyAllParameters(null, $"{Urls.Invoices}/{invoiceId}", false),
+                Requestor.PostString(
+                    this.ApplyAllParameters(createOptions, Urls.Invoices, false),
```
remi-stripe: Flagging for later, can we find a way to have the URL be an a class method/instance instead while we do that work of porting to the new services logic?

remi-stripe: well noticing you did it for refund below

PR #1288, opened 2018-09-17
Fix FileUploadService

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Changes `FileUploadService.Create` to accept a new `FileUploadCreateOptions` class instead of separate arguments.


src/Stripe.net/Services/FileUploads/FileUploadService.cs
```diff
@@ -18,34 +18,35 @@ public FileUploadService(string apiKey)
         {
         }
 
-        public virtual FileUpload Create(string fileName, Stream fileStream, string purpose, RequestOptions requestOptions = null)
+        public virtual FileUpload Create(FileUploadCreateOptions options = null, RequestOptions requestOptions = null)
```
remi-stripe: wouldn't options be required for file upload creation?

src/Stripe.net/Infrastructure/Requestor.cs
```diff
@@ -187,16 +187,24 @@ private static string GetAuthorizationHeaderValue(string apiKey)
...
+
+            #if NET45
+            // Doing this on .NET Standard would require us to bump the minimum framework version
+            // to .NET Standard 1.3, which isn't worth it since the filename is basically ignored
+            // by the server.
+            FileStream fileStream = stream as FileStream;
+            if ((fileStream != null) && (!string.IsNullOrEmpty(fileStream.Name)))
             {
-                fileName = "blob";
+                fileName = fileStream.Name;
```
remi-stripe: Isn't it easier to just put the filename in the options instead of trying to get it from the stream? And if it's not in the options we just use blob? I am not a fan of those pre-processor directives as we don't really test them thoroughly

src/StripeTests/Services/FileUploads/FileUploadServiceTest.cs
```diff
@@ -14,15 +14,19 @@ public class FileUploadServiceTest : BaseStripeTest
         private const string FileUploadId = "file_123";
         private const string FileName = "StripeTests.Resources.file_upload_logo.png";
 
-        private Stream stream;
         private FileUploadService service;
+        private FileUploadCreateOptions createOptions;
```
remi-stripe: I don't know why the file above had permissions changed?

src/Stripe.net/Services/FileUploads/FileUploadService.cs
ob-stripe: Yep. I'll fix.

src/Stripe.net/Infrastructure/Requestor.cs
ob-stripe: Re. testing, we do test with .NET Framework on AppVeyor, so we know that the code compiles and runs without errors at least (which in this instance is as much as we can say with .NET Core anyway).

Mono is supposed to be largely compatible with .NET Framework, but we've never tested against it (or offered any guarantee that the library works with Mono for that matter).

> Why don't we just make filename an optional parameter by the way in the API itself?

The file is encoded as described in [RFC 2388](https://tools.ietf.org/html/rfc2388), which already includes a way of (optionally) providing the filename.

In that regard, I guess we _could_ add a field for the filename in the options class, but even if we do I'd still be in favor of keeping the code I added to automatically provide the filename if it's not manually provided.

remi-stripe: Yeah I mean I'm fine with what you do, I'm just raising that .NET is still a platform we don't know as much about where environments would/could matter.

But I'm fine with merging as is, I just wanted to raise an alternative because I never understood why we didn't just make the filename a real argument instead.

PR #1297, opened 2018-09-25
Cross target netstandard2.0

iamcarbon:
This PR adds netstandard2.0 to the target frameworks. This simplifies the dependency tree on modern runtimes.

### DIFF
--- a/src/Stripe.net/Stripe.net.csproj
+++ b/src/Stripe.net/Stripe.net.csproj
@@ -6,7 +6,7 @@
     <VersionPrefix>19.8.0</VersionPrefix>
     <Version>19.8.0</Version>
     <Authors>Stripe, Jayme Davis</Authors>
-    <TargetFrameworks>netstandard1.2;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard1.2;netstandard2.0;net45</TargetFrameworks>
     <AssemblyName>Stripe.net</AssemblyName>
     <PackageId>Stripe.net</PackageId>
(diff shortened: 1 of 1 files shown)


ob-stripe:
Hi @iamcarbon, thanks for the contribution!

Can you clarify how this helps exactly? I'd've thought that libraries would only need to target the _minimum_ version of .NET Standard that they require. Is that not the case?

iamcarbon:
Hey there. Targeting NETSTANDARD2.0 directly eliminates the 50+ dependencies that come from .NETSTANDARD 1.x libraries. .NETSTANDARD2.0 reduces this to a single dependency (netstsandard.dll) that is already included in modern runtimes (netfx472, netcoreapp2.0, etc).

The official guidance from MSFT is now to re-baseline at 2.0x or multi-target:
https://twitter.com/davidfowl/status/1008453902058917888

ob-stripe:
Cool, thanks @iamcarbon. Going to pull this in.

ob-stripe:
Released as 19.9.2.

ob-stripe:
Just for future reference in case someone stumbles upon this: https://github.com/dotnet/standard/blob/master/docs/versions.md#how-do-i-know-which-net-standard-version-i-should-target

> * You should target the lowest version you can get away with.
> * If you target .NET Standard 1.x, add .NET Standard 2.0 as [another target](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-target-frameworks).

Issue #971, opened 2017-07-24
Fix Full-Permissions Requirement


anelder-stripe:
Right now there is a full-permissions requirement on the library, which means it doesn't run on every framework.  There are two things causing the problem:

### Registry Access

https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Infrastructure/Client.cs#L88-L95

This code is attempting to access the registry; which means it requires UnmanagedCode permissions.  A possible fix (provided by the amazing Engineers at dotnetfiddle) would be something like this

```csharp
 private static string getFrameworkFromRegistry()
{
    if (!AppDomain.CurrentDomain.IsFullyTrusted)
        return "unknown";
```

### Internal Modifier for SourceConverter\StripeDateTimeConverter + Certain Properties

When Newtonsoft.Json tries to create converters with internal access or set values to internal properties, it fails.  Placing this inside of an `InternalsVisibleTo`-file should fix it:

```csharp
[assembly: InternalsVisibleTo("Newtonsoft.Json")]
```

I think this file -- https://github.com/stripe/stripe-dotnet/blob/212b53759ca190f1a2ede3c65d40c3c08a6ef103/src/Stripe.net/Properties/InternalsVisibleTo.cs

anelder-stripe:
Relevant to our interests: http://www.bricelam.net/2012/08/partial-trust-testing-with-xunitnet.html

mickbyrne:
The registry access also means the Stripe.net library doesn't work when running Visual Studio for Mac. I've just logged a new issue for you about it, https://github.com/stripe/stripe-dotnet/issues/990.

ob-stripe:
@anelder-stripe Heads up that I removed the registry access entirely in 15.3.1.

ob-stripe:
Closing this issue due to staleness. The registry access was fixed a long time ago.

Note that Stripe.net is still not usable on .NET Fiddle, but I believe this is because:
1. Stripe.net uses Newtonsoft.Json
2. Newtonsoft.Json provides both strong named and not strong named packages
3. Stripe.net only provides not strong named packages
4. The version of Newtonsoft.Json on .NET Fiddle is the strong named one
5. Due to the way strong naming works, a not strong named package cannot use a strong named package

[...]

PR #1300, opened 2018-09-27
Make internals visible to Newtonsoft.Json

ob-stripe:
r? @anelder-stripe @remi-stripe 
cc @stripe/api-libraries 

Make internals visible to Newtonsoft.Json, as suggested by @anelder-stripe in #971.

Unfortunately, I'm not sure if it's possible to write a test for this, but the change does make sense to me: `StripeDateTimeConverter` is declared as an `internal` class, but it needs to be instantiated by Newtonsoft.Json when deserializing.


remi-stripe:
This seems fine to me but deferring to @anelder-stripe on this since he had the original suggestion.

PR #1301, opened 2018-09-27
Convert all options to have one class per level in the JSON params

remi-stripe:
This PR takes all existing Options classes and splits the properties per level in the JSON. This is to ensure that our class structure matches our API naming as in other libraries. It also makes the code easier to re-use when hashes are shared across resources.

cc @stripe/api-libraries


remi-stripe:
@ob-stripe I did the majority of those. It's missing the `AmountOne` and `AmountTwo` for bank account verification but I'm not sure we can fix that one without hard-coding the 0/1 index in the property? Is it the right approach?
It's also missing the one for the list dates gt/gte/etc.

remi-stripe:
@ob-stripe Okay I did all the changes we discussed. I also took this as an opportunity to add new lines to all Options files which should make you happy.

Can you do a full review?

src/Stripe.net/Services/Invoices/InvoiceListOptions.cs
```diff
@@ -14,13 +15,22 @@ public class InvoiceListOptions : ListOptions
...
 
         [JsonProperty("date")]
-        public DateFilter Date { get; set; }
+        public DateTime? Date { get; set; }
+
+        /// <summary>
+        /// A filter on the list based on the object date field.
+        /// </summary>
+        [JsonProperty("date")]
+        public DateFilterOptions DateFilter { get; set; }
```
ob-stripe: Rename to `DateRange` for consistency.

src/Stripe.net/Services/Invoices/InvoiceListOptions.cs
```diff
@@ -14,13 +15,22 @@ public class InvoiceListOptions : ListOptions
...
+
+        [JsonProperty("due_date")]
+        public DateTime? DueDate { get; set; }
 
         /// <summary>
         /// A filter on the list based on the object due_date field.
         /// </summary>
         [JsonProperty("due_date")]
-        public DateFilter DueDate { get; set; }
+        public DateFilterOptions DueDateFilter { get; set; }
```
ob-stripe: Rename to `DueDateRange` for consistency.

src/Stripe.net/Services/DateFilterOptions.cs
```diff
@@ -3,21 +3,18 @@
     using System;
     using Newtonsoft.Json;
 
-    public class DateFilter
+    public class DateFilterOptions : INestedOptions
```
ob-stripe: I think `DateRangeOptions` would be a better name: filtering on an exact date is also a "date filter", but this class now only handles ranges. It would also make the class name consistent with the properties names.

remi-stripe:
@ob-stripe Fixed, PTYAL

ob-stripe:
LGTM!

ob-stripe:
Just for future reference in case someone stumbles upon this: https://github.com/dotnet/standard/blob/master/docs/versions.md#how-do-i-know-which-net-standard-version-i-should-target

> * You should target the lowest version you can get away with.
> * If you target .NET Standard 1.x, add .NET Standard 2.0 as [another target](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-target-frameworks).

PR #1304, opened 2018-09-28
Sanitize files

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

- Remove UTF-8 BOMs
- Add missing final newlines

Clearly a high impact PR that will definitely make a lot of people happy and not just me.


ob-stripe:
> The sadness is that you won't be able to complain on each PR I make that it's missing the new line at the end of the file :p

You can always create new files 😛

PR #1308, opened 2018-09-29
Better URL handling in services

ob-stripe:
Move URL creation in `Service` as much as possible, so that concrete services don't need to create their own URLs in every method.

Most services will only need to declare their `ObjectName`, then use the new parent methods for CRUD + list requests, or call `this.ClassUrl()` / `this.InstanceUrl(id)` as appropriate.

Some services will need to override `ClassUrl` if their object name doesn't cleanly map to the URL.

I'm not sure yet how to handle services for nested resources -- maybe another parent class `ServiceNested` with more helper methods, similar to these ones but with an additional argument for the parent resource's ID.

ptal @remi-stripe

### DIFF
--- a/src/Stripe.net/Services/ApplePayDomains/ApplePayDomainService.cs
+++ b/src/Stripe.net/Services/ApplePayDomains/ApplePayDomainService.cs
@@ -21,44 +21,46 @@ namespace Stripe
         {
         }
 
+        public override string BasePath => "/apple_pay/domains";
+
         public virtual ApplePayDomain Create(ApplePayDomainCreateOptions options, RequestOptions requestOptions = null)
         {
-            return this.Post($"{Urls.BaseUrl}/apple_pay/domains", requestOptions, options);
+            return this.CreateEntity(options, requestOptions);
         }
 
         public virtual Task<ApplePayDomain> CreateAsync(ApplePayDomainCreateOptions options, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.PostAsync($"{Urls.BaseUrl}/apple_pay/domains", requestOptions, cancellationToken, options);
+            return this.CreateEntityAsync(options, requestOptions, cancellationToken);
         }
 
         public virtual ApplePayDomain Delete(string domainId, RequestOptions requestOptions = null)
         {
-            return this.DeleteEntity($"{Urls.BaseUrl}/apple_pay/domains/{domainId}", requestOptions);
+            return this.DeleteEntity(domainId, null, requestOptions);
         }
 
         public virtual Task<ApplePayDomain> DeleteAsync(string domainId, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.DeleteEntityAsync($"{Urls.BaseUrl}/apple_pay/domains/{domainId}", requestOptions, cancellationToken);
+            return this.DeleteEntityAsync(domainId, null, requestOptions, cancellationToken);
         }
 
         public virtual ApplePayDomain Get(string domainId, RequestOptions requestOptions = null)
         {
-            return this.GetEntity($"{Urls.BaseUrl}/apple_pay/domains/{domainId}", requestOptions);
+            return this.GetEntity(domainId, null, requestOptions);
         }
 
         public virtual Task<ApplePayDomain> GetAsync(string domainId, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.GetEntityAsync($"{Urls.BaseUrl}/apple_pay/domains/{domainId}", requestOptions, cancellationToken);
+            return this.GetEntityAsync(domainId, null, requestOptions, cancellationToken);
         }
 
-        public virtual StripeList<ApplePayDomain> List(ApplePayDomainListOptions listOptions = null, RequestOptions requestOptions = null)
+        public virtual StripeList<ApplePayDomain> List(ApplePayDomainListOptions options = null, RequestOptions requestOptions = null)
         {
-            return this.GetEntityList($"{Urls.BaseUrl}/apple_pay/domains", requestOptions, listOptions);
+            return this.ListEntities(options, requestOptions);
         }
 
-        public virtual Task<StripeList<ApplePayDomain>> ListAsync(ApplePayDomainListOptions listOptions = null, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
+        public virtual Task<StripeList<ApplePayDomain>> ListAsync(ApplePayDomainListOptions options = null, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.GetEntityListAsync($"{Urls.BaseUrl}/apple_pay/domains", requestOptions, cancellationToken, listOptions);
+            return this.ListEntitiesAsync(options, requestOptions, cancellationToken);
         }
     }
 }
--- a/src/Stripe.net/Services/Balance/BalanceService.cs
+++ b/src/Stripe.net/Services/Balance/BalanceService.cs
@@ -9,8 +9,6 @@ namespace Stripe
     public class BalanceService : Service<Balance>,
         ISingletonRetrievable<Balance>
     {
-        private static string classUrl = Urls.BaseUrl + "/balance";
-
         public BalanceService()
             : base(null)
         {
@@ -21,14 +19,16 @@ namespace Stripe
         {
         }
 
+        public override string BasePath => "/balance";
+
         public virtual Balance Get(RequestOptions requestOptions = null)
         {
-            return this.GetEntity($"{Urls.BaseUrl}/balance", requestOptions);
+            return this.GetRequest<Balance>(this.ClassUrl(), null, requestOptions);
         }
 
         public virtual Task<Balance> GetAsync(RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.GetEntityAsync($"{Urls.BaseUrl}/balance", requestOptions, cancellationToken);
+            return this.GetRequestAsync<Balance>(this.ClassUrl(), null, requestOptions, cancellationToken);
         }
     }
 }
--- a/src/Stripe.net/Services/Cards/CardService.cs
+++ b/src/Stripe.net/Services/Cards/CardService.cs
@@ -6,7 +6,7 @@ namespace Stripe
     using System.Threading.Tasks;
     using Stripe.Infrastructure;
 
-    public class CardService : Service<Card>,
+    public class CardService : ServiceNested<Card>,
         INestedCreatable<Card, CardCreateOptions>,
         INestedDeletable<Card>,
         INestedListable<Card, CardListOptions>,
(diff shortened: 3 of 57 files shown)


src/Stripe.net/Services/ApplePayDomains/ApplePayDomainService.cs
```diff
@@ -21,44 +21,53 @@ public ApplePayDomainService(string apiKey)
...
         public virtual Task<ApplePayDomain> DeleteAsync(string domainId, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.DeleteEntityAsync($"{Urls.BaseUrl}/apple_pay/domains/{domainId}", requestOptions, cancellationToken);
+            return this.DeleteEntityAsyncNew(domainId, null, requestOptions, cancellationToken);
         }
 
         public virtual ApplePayDomain Get(string domainId, RequestOptions requestOptions = null)
         {
-            return this.GetEntity($"{Urls.BaseUrl}/apple_pay/domains/{domainId}", requestOptions);
+            return this.GetEntityNew(domainId, null, requestOptions);
```
remi-stripe: Why do we need `null` on GET requests?

src/Stripe.net/Services/ApplePayDomains/ApplePayDomainService.cs
```diff
@@ -21,44 +21,53 @@ public ApplePayDomainService(string apiKey)
...
         public virtual Task<ApplePayDomain> CreateAsync(ApplePayDomainCreateOptions options, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.PostAsync($"{Urls.BaseUrl}/apple_pay/domains", requestOptions, cancellationToken, options);
+            return this.CreateEntityAsync(options, requestOptions, cancellationToken);
         }
 
         public virtual ApplePayDomain Delete(string domainId, RequestOptions requestOptions = null)
         {
-            return this.DeleteEntity($"{Urls.BaseUrl}/apple_pay/domains/{domainId}", requestOptions);
+            return this.DeleteEntityNew(domainId, null, requestOptions);
```
remi-stripe: I'm assuming there's no way to get rid of the `null` for most DELETE?

src/Stripe.net/Services/ApplePayDomains/ApplePayDomainService.cs
```diff
@@ -21,44 +21,53 @@ public ApplePayDomainService(string apiKey)
...
         {
-            return this.GetEntityListAsync($"{Urls.BaseUrl}/apple_pay/domains", requestOptions, cancellationToken, listOptions);
+            return this.ListEntitiesAsync(options, requestOptions, cancellationToken);
+        }
+
+        protected override string ClassUrl(string baseUrl = null)
+        {
+            baseUrl = baseUrl ?? StripeConfiguration.GetApiBase();
+
+            return $"{baseUrl}/apple_pay/domains";
```
remi-stripe: Not a fan of that override. I would rather calculate the relative path instead. Can we do "ObjectPath" or something instead?

src/Stripe.net/Services/Issuing/Authorizations/AuthorizationService.cs
```diff
@@ -23,54 +21,56 @@ public AuthorizationService(string apiKey)
...
         public virtual Authorization Approve(string authorizationId, AuthorizationApproveOptions options = null, RequestOptions requestOptions = null)
         {
-            return this.Post($"{classUrl}/{authorizationId}/approve", requestOptions, options);
+            return this.PostRequest<Authorization>($"{this.InstanceUrl(authorizationId)}/approve", options, requestOptions);
         }
 
         public virtual Task<Authorization> ApproveAsync(string authorizationId, AuthorizationApproveOptions options = null, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.PostAsync($"{classUrl}/{authorizationId}/approve", requestOptions, cancellationToken, options);
+            return this.PostRequestAsync<Authorization>($"{this.InstanceUrl(authorizationId)}/approve", options, requestOptions, cancellationToken);
```
remi-stripe: I wonder if it'd be cleaner to pass the authorization id + append name? Not sure what that would look like but I fell like it could be cleaner? Sorry it's a bit vague and likely not possible but I was thinking of `this.Url(authorizationId, "/approve")` or similar?

src/Stripe.net/Services/ApplePayDomains/ApplePayDomainService.cs
ob-stripe: The `GetEntityNew` method is capable of adding parameters to retrieve requests, even though we don't do this in the library (mostly because the only case where this is needed is expansion, and this is handled separately via the `Expand*` properties on the service).

That said, I added an override to be able to call the method without an options class.

src/Stripe.net/Services/ApplePayDomains/ApplePayDomainService.cs
ob-stripe: No, this is possible too. I added an override to be able to call `DeleteEntityNew` without an options class.

src/Stripe.net/Services/ApplePayDomains/ApplePayDomainService.cs
ob-stripe: Yeah, that makes way more sense for a service actually. I changed `ObjectName` to `BasePath`.

src/Stripe.net/Services/Issuing/Authorizations/AuthorizationService.cs
ob-stripe: Hm, not sure it's worth it. `this.Url(authorizationId, "/approve")` would be the same thing as `this.Url(authorizationId) + "/approve"`, which is already the same thing as `$"{this.InstanceUrl(authorizationId)}/approve"`.

src/Stripe.net/Services/ApplePayDomains/ApplePayDomainService.cs
remi-stripe: But why since expansion is on the other options and no GET (non list) method takes an option?

src/Stripe.net/Services/Issuing/Authorizations/AuthorizationService.cs
remi-stripe: yeah fair, it's not the same approach in my head (and the former is easier to mess up). I ask because I made a mistake a few times yesterday with the /approve, /decline and all where stripe-mock still just worked because params were identical and I had messed up my URL calculation

src/Stripe.net/Services/Products/ProductService.cs
```diff
@@ -23,54 +23,56 @@ public ProductService(string apiKey)
...
         public virtual Task<Product> CreateAsync(ProductCreateOptions options, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.PostAsync($"{Urls.BaseUrl}/products", requestOptions, cancellationToken, options);
+            return this.CreateEntityAsync(options, requestOptions, cancellationToken);
         }
 
         public virtual Product Delete(string productId, RequestOptions requestOptions = null)
         {
-            return this.DeleteEntity($"{Urls.BaseUrl}/products/{WebUtility.UrlEncode(productId)}", requestOptions);
+            return this.DeleteEntityNew(productId, requestOptions);
```
remi-stripe: Flagging that I had to remove the encoding for now. I think it makes more sense to do it at the next level. What do you think?

remi-stripe:
@ob-stripe I did most of the services but left the nested ones for now. I think this one needs a thorough review because so much has changed and it's easy to mess up

remi-stripe:
I'm not sure why it fails on appveyor and works locally :(

src/Stripe.net/Services/Products/ProductService.cs
ob-stripe: `InstanceUrl` already takes care of encoding the ID! :)

src/Stripe.net/Services/Balance/BalanceService.cs
```diff
@@ -23,12 +23,12 @@ public BalanceService(string apiKey)
...
         public virtual Balance Get(RequestOptions requestOptions = null)
         {
-            return this.GetRequest<Balance>($"{Urls.BaseUrl}/{this.BasePath}", null, requestOptions);
+            return this.GetRequest<Balance>(this.ClassUrl(), null, requestOptions);
         }
 
         public virtual Task<Balance> GetAsync(RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.GetRequestAsync<Balance>($"{Urls.BaseUrl}/{this.BasePath}", null, requestOptions, cancellationToken);
+            return this.GetRequestAsync<Balance>(this.ClassUrl(), null, requestOptions, cancellationToken);
```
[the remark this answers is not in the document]
ob-stripe: Your path was `/v1//balance`, and stripe-mock doesn't like the double slash.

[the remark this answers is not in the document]
remi-stripe: yeah I know but it worked locally without any issue which is weird

src/Stripe.net/Services/Cards/CardService.cs
```diff
@@ -6,7 +6,7 @@ namespace Stripe
     using System.Threading.Tasks;
     using Stripe.Infrastructure;
 
-    public class CardService : Service<Card>,
+    public class CardService : ServiceNested<Card, Customer>,
```
remi-stripe: Likely not a problem but flagging since I thought about it. What if we have a generic nested resource in the future (like card and bank account used to be) and we want to have it shared? Would we simply add a parent class/interface or similar instead?

src/Stripe.net/Services/ServiceNested.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            return this.GetRequestAsync<StripeList<EntityReturned>>(this.ClassUrl(parentId), options, requestOptions, cancellationToken);
+        }
+
+        protected EntityReturned UpdateNestedEntity(string parentId, string id, BaseOptions options = null, RequestOptions requestOptions = null)
+        {
+            return this.PostRequest<EntityReturned>(this.InstanceUrl(parentId, id), options, requestOptions);
+        }
+
+        protected Task<EntityReturned> UpdateNestedEntityAsync(string parentId, string id, BaseOptions options = null, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
```
remi-stripe: Minor but I wonder if we should avoid the default values on parameters like requestOptions and cancellationToken. Why you ask? Because it's easy to mess up your copy/paste (we both did in this PR) and have `cancellationToken` be the default value instead of what the upper level passed.

src/Stripe.net/Services/TransferReversals/TransferReversalService.cs
```diff
@@ -21,48 +21,50 @@ public TransferReversalService(string apiKey)
...
         public virtual Task<TransferReversal> CreateAsync(string transferId, TransferReversalCreateOptions options = null, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.PostAsync($"{Urls.BaseUrl}/transfers/{transferId}/reversals", requestOptions, cancellationToken, options);
+            return this.CreateNestedEntityAsync(transferId, options, requestOptions, cancellationToken);
         }
 
         public virtual TransferReversal Get(string transferId, string reversalId, RequestOptions requestOptions = null)
         {
-            return this.GetEntity($"{Urls.BaseUrl}/transfers/{transferId}/reversals/{reversalId}", requestOptions);
+            return this.GetNestedEntity(transferId, reversalId, requestOptions);
```
remi-stripe: I see you used the version without `null` for options. but you did not use it everywhere. I know I kinda asked for it but per my previous comment I'm starting to think being explicit with what params we pass and why is likely safer/better after all (soooorry)

src/Stripe.net/Services/Discounts/DiscountService.cs
```diff
@@ -16,24 +16,26 @@ public DiscountService(string apiKey)
         {
         }
 
+        public override string BasePath => null;
```
remi-stripe: Oh well I guess this is the example of a resource nested under different parents?

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries

remi-stripe:
LGTM assuming the tests pass (since we had a few failures on the branch and not locally on this work)

PR #1309, opened 2018-09-29
Group resources with sub-resources under one folder

remi-stripe:
Some resources go together such as `Account` and `LegalEntity` or `PayoutSchedule` so we group all of them under a common folder.

At the same time, I renamed a few fields on some resources to match the API (but then stopped when I realized the diff would not really show that) and added support for `AddressJapan` on the Account resource.

r? @ob-stripe 
cc @stripe/api-libraries


remi-stripe:
One question I have is whether it'd be cleaner to have one folder per resource even if there's just one file? And if so, whether all shared/common resources are better under the resource they are closest to or under a `Common` folder like `Address` or `Inventory`.

ob-stripe:
Rename `Balances/` to `Balance/` to be consistent with the `Services/` tree / the API.

> One question I have is whether it'd be cleaner to have one folder per resource even if there's just one file?

I'd have a slight preference for always having one folder per resource, though I don't feel too strongly about it.

> And if so, whether all shared/common resources are better under the resource they are closest to or under a Common folder like Address or Inventory.

I'd be in favor of using a `Common` folder.

remi-stripe:
@ob-stripe Done! PTAL

ob-stripe:
Rename `ThreeDSecures` to `ThreeDSecure`. Otherwise LGTM!

PR #1310, opened 2018-09-30
Verifying requests in tests

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add support for verifying requests in tests.

Some notes:

- I had to disable parallel test execution. I suspect the mock object breaks thread-safety in some way. The test suite is fast now that we use stripe-mock so this isn't a huge deal.

- Right now, we can only check the HTTP method and path -- the body is stored in an object that is GC'd by the time `AssertRequest` is called. This is probably solvable by replacing the default mock by a more complex object that would store the body.

### DIFF
--- a/src/StripeTests/Properties.cs
+++ b/src/StripeTests/Properties.cs
@@ -1,0 +1,1 @@
+[assembly: Xunit.CollectionBehavior(DisableTestParallelization = true)]
--- a/src/StripeTests/Services/Accounts/AccountServiceTest.cs
+++ b/src/StripeTests/Services/Accounts/AccountServiceTest.cs
@@ -1,6 +1,7 @@
 namespace StripeTests
 {
     using System.Collections.Generic;
+    using System.Net.Http;
     using System.Threading.Tasks;
 
     using Stripe;
@@ -92,6 +93,7 @@ namespace StripeTests
         public void Create()
         {
             var account = this.service.Create(this.createOptions);
+            this.AssertRequest(HttpMethod.Post, "/v1/accounts");
             Assert.NotNull(account);
             Assert.Equal("account", account.Object);
         }
@@ -100,6 +102,7 @@ namespace StripeTests
         public async Task CreateAsync()
         {
             var account = await this.service.CreateAsync(this.createOptions);
+            this.AssertRequest(HttpMethod.Post, "/v1/accounts");
             Assert.NotNull(account);
             Assert.Equal("account", account.Object);
         }
@@ -108,6 +111,7 @@ namespace StripeTests
         public void Delete()
         {
             var deleted = this.service.Delete(AccountId);
+            this.AssertRequest(HttpMethod.Delete, "/v1/accounts/acct_123");
             Assert.NotNull(deleted);
         }
 
@@ -115,6 +119,7 @@ namespace StripeTests
         public async Task DeleteAsync()
         {
             var deleted = await this.service.DeleteAsync(AccountId);
+            this.AssertRequest(HttpMethod.Delete, "/v1/accounts/acct_123");
             Assert.NotNull(deleted);
         }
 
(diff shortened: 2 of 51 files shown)


src/StripeTests/Properties.cs
```diff
@@ -0,0 +1 @@
+[assembly: Xunit.CollectionBehavior(DisableTestParallelization = true)]
```
remi-stripe: There has to be a way to not disable this and have it work right?

src/StripeTests/Services/Accounts/AccountServiceTest.cs
```diff
@@ -108,20 +111,23 @@ public async Task CreateAsync()
         public void Delete()
         {
             var deleted = this.service.Delete(AccountId);
+            this.AssertRequest(HttpMethod.Delete, "/v1/accounts/acct_123");
```
remi-stripe: minor but using `AccountId` might be cleaner (not convinced it would)

src/StripeTests/Services/UsageRecordSummaries/UsageRecordSummaryServiceTest.cs
```diff
@@ -16,13 +17,17 @@ public UsageRecordSummaryServiceTest()
         {
             this.service = new UsageRecordSummaryService();
 
-            this.listOptions = new UsageRecordSummaryListOptions();
+            this.listOptions = new UsageRecordSummaryListOptions()
+            {
+                Limit = 1,
+            };
```
remi-stripe: why did you have to add this?

src/StripeTests/Properties.cs
remi-stripe: Like clarifying: not specifically in this PR per se, it's fine to merge as is, but surprising there isn't a feature for this in the lib

ob-stripe: Sure, but I suspect the way may be "use a lock to ensure thread-safety", which would negate most gains from parallelization since tests would need to wait on the lock to send requests. I might revisit this in the future but for now I didn't want to block on this since the suite still runs very fast.

src/StripeTests/Services/Accounts/AccountServiceTest.cs
ob-stripe: I like the explicitness of using the expected string directly, without interpolation. It's also consistent with what we do in other libraries (previously, in some libraries, we _had_ to use interpolated strings because stripe-mock would return objects with a different ID than the requested one, but 1/ the stripe-mock issue is fixed now and 2/ this doesn't apply to stripe-dotnet in any case as it uses static methods on service objects, not instance methods on resource objects).

src/StripeTests/Services/UsageRecordSummaries/UsageRecordSummaryServiceTest.cs
ob-stripe: I didn't have to -- it's just the first test I played around with in this PR and I added this to ensure that query parameters weren't part of the `AbsoluteUrl` when asserting the URL. I left it since it's consistent with other list tests.

remi-stripe: Ah perfect! Thanks for clarifying, let's merge and have fun rebasing :p

src/StripeTests/Properties.cs
ob-stripe: Actually, I think I know what's happening: stripe-dotnet uses a single `HttpClient` object (which is what Microsoft recommends -- the `SendAsync` method is thread-safe). However, if multiple tests are run in parallel, then we cannot assert on the last request since this may happen:
- Thread 1: reset mock
- Thread 1: send request
- Thread 2: reset mock
- Thread 1: assert on last request -> fails because the mock was reset

Even if we don't reset the mock between every test, this may happen:
- Thread 1: send request
- Thread 2: send request
- Thread 1: assert on last request -> fails because the last request is not the expected one

So I think disabling parallel execution is the right move here.

ob-stripe:
Pulling this in!

PR #1311, opened 2018-10-01
Ensure all properties are on resources

remi-stripe:
Did a pass using the OpenAPI spec + the Docs to ensure that all resources have the right list of properties. This meant adding multiple type-specific classes for Sources too.
Also renamed a few properties to match the API better such as `ExpirationMonth` => `ExpMonth`.

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Looks much better!

I wonder if we could write a wholesome test to ensure that fields are declared in alphabetical order (though probably difficult with `Internal` stuff) / field names must match the `JsonProperty` declaration...

remi-stripe:
Yeah I think lately I've rushed too many PRs and did not pay enough attention to the fields name and I also made a few mistakes so I'm going to start being more thorough.

PR #1313, opened 2018-10-01
[WIP] [experimental] Interfaces for polymorphic API resources

ob-stripe:
An experiment to see if we could replace classes like `ExternalAccount` and `PaymentSource` (which don't really exist in the API) by interfaces.


src/Stripe.net/Infrastructure/JsonConverters/ExternalAccountConverter.cs
```diff
@@ -24,21 +22,17 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
         {
             var incoming = JObject.Load(reader);
 
-            var externalAccount = new ExternalAccount
-            {
-                Id = incoming.SelectToken("id").ToString()
-            };
+            var externalAccount = default(IExternalAccount);
```
remi-stripe: didn't you say you did not need to use `var`?

src/Stripe.net/Infrastructure/Public/Mapper.cs
```diff
@@ -5,9 +5,14 @@ namespace Stripe
     using System.Reflection;
     using Newtonsoft.Json;
     using Newtonsoft.Json.Linq;
+    using Stripe.Infrastructure;
 
     public static class Mapper<T>
     {
+        private static JsonConverter[] converters = {
+            new ExternalAccountConverter(),
```
remi-stripe: so this is the new place where we would list all converters?

src/Stripe.net/Infrastructure/StringOrObject.cs
```diff
@@ -4,20 +4,21 @@ namespace Stripe.Infrastructure
...
+                // T item = ((JToken)value).ToObject<T>();
+                T item = Mapper<T>.MapFromJson(value.ToString());
                 updateId(item.Id);
                 updateObject(item);
             }
             else if (value is string)
             {
                 updateId((string)value);
-                updateObject(null);
+                updateObject(default(T));
```
remi-stripe: can you clarify what this line does/changes?

src/Stripe.net/Infrastructure/StringOrObject.cs
```diff
@@ -4,20 +4,21 @@ namespace Stripe.Infrastructure
...
-        where T : StripeEntityWithId
+        where T : IStripeEntityWithId
     {
         public static void Map(object value, Action<string> updateId, Action<T> updateObject)
         {
             if (value is JObject)
             {
-                T item = ((JToken)value).ToObject<T>();
+                // T item = ((JToken)value).ToObject<T>();
+                T item = Mapper<T>.MapFromJson(value.ToString());
```
remi-stripe: Not fully sure what all of this does

src/StripeTests/Entities/ExternalAccounts/ExternalAccountTest.cs
```diff
@@ -8,17 +8,17 @@ namespace StripeTests
...
             string json = GetFixture("/v1/accounts/acct_123/external_accounts/ba_123");
-            var externalAccount = Mapper<ExternalAccount>.MapFromJson(json);
+            var externalAccount = Mapper<IExternalAccount>.MapFromJson(json);
             Assert.NotNull(externalAccount);
-            Assert.Equal(ExternalAccountType.BankAccount, externalAccount.Type);
-            Assert.NotNull(externalAccount.BankAccount);
-            Assert.Equal("bank_account", externalAccount.BankAccount.Object);
+            Assert.IsType<BankAccount>(externalAccount);
+            var bankAccount = (BankAccount)externalAccount;
+            Assert.Equal("bank_account", bankAccount.Object);
```
remi-stripe: Same you mentioned you did not need to use `var` anymore

src/StripeTests/StripeTests.csproj
```diff
@@ -20,6 +20,7 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net452' ">
+    <Reference Include="Microsoft.CSharp" />
```
remi-stripe: What is this exactly?

src/Stripe.net/Infrastructure/JsonConverters/ExternalAccountConverter.cs
ob-stripe: To clarify: `dynamic` _can_ (but it's not a requirement) be used to tell the compiler that the exact type won't be known until runtime. It's not a silver bullet, because it disable compile-type checks. E.g. if I write this:
```c#
dynamic foo = account.ExternalAccounts.Data[0];
Console.WriteLine(foo.AccountHolderName)
```

The compiler has no way of knowing whether `AccountHolderName` is a valid call. If it's not (because `foo` is a `Card` instead of a `BankAccount`), a runtime exception will be raised.

A safer but more verbose alternative:
```c#
var foo = account.ExternalAccounts.Data[0]; // foo is an `IExternalAccount`
if (foo is BankAccount)
{
  Console.WriteLine(((BankAccount)foo).AccountHolderName); // need to manually cast to BankAccount
}
else if (foo is Card)
{
  Console.WriteLine(((Card)foo).Name); // need to manually cast to Card
}
else
{
  // this should never be reached, unless a new concrete type is introduced for IExternalAccount
}
```

In the specific place of the code, we can use `var` because we don't need to access any of specific properties of the concrete types. We could replace `var` with `IExternalAccount` an it would be a no-op (it's the type that is inferred by the compiler).

src/Stripe.net/Infrastructure/Public/Mapper.cs
ob-stripe: Yep. Not sure if this is the best way to achieve this, but it seems to work and should be relatively efficient (it's `static`, so only initialized once).

src/Stripe.net/Infrastructure/JsonConverters/ExternalAccountConverter.cs
remi-stripe: Ack thanks for the thorough explanation. So the logic becomes similar to what we do in stripe-java. More verbose but safer and cleaner.

src/Stripe.net/Infrastructure/StringOrObject.cs
ob-stripe: In practice, nothing. `default(T)` returns the default value for the type `T`. For reference types, the default value is `null`.

However, since `T` is now constrained to be an interface rather than a class, the compiler complains that it cannot convert from `null` to `T`. This is (I think) because `struct`s are value types but can implement interfaces.

src/Stripe.net/Infrastructure/StringOrObject.cs
ob-stripe: It uses our custom deserialization instead of the default one, to ensure that our registered converters are used.

src/StripeTests/Entities/ExternalAccounts/ExternalAccountTest.cs
ob-stripe: I forgot to update this test. But as said above, using `dynamic` is optional, it's basically syntactic sugar at the cost of type safety.

src/StripeTests/StripeTests.csproj
ob-stripe: This is needed to use `dynamic` with .NET Framework (at least for version 4.5.2). Note that this is only added in the tests project. Users of the library who use .NET Framework and want to use `dynamic` would need to do the same thing, but the library doesn't make this a requirement.

src/StripeTests/Entities/ExternalAccounts/ExternalAccountTest.cs
ob-stripe: Thinking about this a bit more, I think we could make `Object` part of the `IStripeEntity` interface (or maybe on a new `IStripeEntityWithObject` interface) so that we can check its value without using `dynamic`.

src/Stripe.net/Infrastructure/JsonConverters/ExternalAccountConverter.cs
brandur-stripe: Sorry for the lurk, but +1 the choice to avoid `dynamic` if possible. It's one of these tools that can be useful in a huge variety of situations, but probably should be avoided in almost all of them because of loss of typedness is a very significant drawback (no more compiler checks, Intellisense, etc.).

ob-stripe:
Closing this one because rebasing is a pain.

PR #1316, opened 2018-10-02
Automatically convert Unix timestamps to DateTimes

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Infrastructure/StringOrObject.cs
+++ b/src/Stripe.net/Infrastructure/StringOrObject.cs
@@ -10,7 +10,12 @@ namespace Stripe.Infrastructure
         {
             if (value is JObject)
             {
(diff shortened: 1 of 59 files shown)


src/Stripe.net/Infrastructure/StringOrObject.cs
```diff
@@ -10,7 +10,7 @@ public static void Map(object value, Action<string> updateId, Action<T> updateOb
         {
             if (value is JObject)
             {
-                T item = ((JToken)value).ToObject<T>();
+                T item = Mapper<T>.MapFromJson(value.ToString());
```
remi-stripe: Can you document this line explicitly as to what it does, why the change and maybe a TODO to figure out in the future how to avoid the parsed => toString => parsed?

PR #1317, opened 2018-10-02
Revamp interfaces on Stripe model classes

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

- Adds an `IStripeEntity` interface implemented by `StripeEntity`
- Adds `IHasId` and `IHasObject` interfaces
- Removes `StripeEntityWithId`
- Ensures that all model classes derive from `StripeEntity`, and add `IHasId` and `IHasObject` interfaces where needed
- Removes `ISupportMetadata` (`ISupportMetadata` was only useful to properly encode `metadata` with the old plugin system, it's useless now)
- Renames `AdditionalOwners` to `AdditionalOwner` (the model represents a single additional owner)
- Removes support for old format of `request` attribute on event objects (cf. https://stripe.com/docs/upgrades#2017-05-25)


ob-stripe:
Also, to clarify the motivation behind this: having both `IHasId` and `IHasObject` interfaces makes it possible to cleanly define e.g. `IExternalAccount` that implements both interfaces above (since an external account always has both an `id` and an `object` attribute).

`IHasId` can be used for field expansion (I've updated the `StringOrObject` generic in this PR) and `IHasObject` can be used in a generic JSON deserializer that instantiates the correct concrete class based on the value of the `object` attribute.

src/Stripe.net/Entities/Disputes/Dispute.cs
```diff
@@ -5,11 +5,8 @@ namespace Stripe
     using Newtonsoft.Json;
     using Stripe.Infrastructure;
 
-    public class Dispute : StripeEntityWithId, ISupportMetadata
+    public class Dispute : StripeEntityWithIdAndObject
```
remi-stripe: I don't think Discount has an id, at least it's not in the docs.

src/Stripe.net/Entities/Recipients/RecipientActiveAccount.cs
```diff
@@ -3,7 +3,7 @@ namespace Stripe
     using System.Collections.Generic;
     using Newtonsoft.Json;
 
-    public class RecipientActiveAccount : StripeEntityWithId, ISupportMetadata
+    public class RecipientActiveAccount : StripeEntityWithId
```
remi-stripe: Why not `StripeEntityWithIdAndObject`

src/Stripe.net/Entities/_contracts/ISupportMetadata.cs
```diff
@@ -1,12 +0,0 @@
-namespace Stripe
```
remi-stripe: Wasn't this used to ensure we added `Metadata` to the Resource too?

src/Stripe.net/Infrastructure/StringOrObject.cs
```diff
@@ -22,7 +22,7 @@ public static void Map(object value, Action<string> updateId, Action<T> updateOb
             else if (value is string)
             {
                 updateId((string)value);
-                updateObject(null);
+                updateObject(default(T));
```
remi-stripe: Can you put a real comment on this line?

src/Stripe.net/Entities/Disputes/Dispute.cs
ob-stripe: 👍  Fixed.

src/Stripe.net/Entities/Recipients/RecipientActiveAccount.cs
ob-stripe: 👍  Fixed.

src/Stripe.net/Entities/_contracts/ISupportMetadata.cs
ob-stripe: I'm not sure why resource classes used this interface at all, but I don't think this is a good reason: the interface would only force you add the `Metadata` property if you remembered to add the interface...

src/Stripe.net/Infrastructure/StringOrObject.cs
ob-stripe: 👍

remi-stripe:
> Yeah, tbh I'm already not thrilled about all the StripeEntityWith... superclasses. I'd rather we always derive from StripeEntity and add the interfaces implemented by each class.

Should we do `StripeEntity` and `StripeEntityWithId` and revert the rest? Do we need an interface for object/metadata/deleted? Or should we just rely on us catching it during PRs? All tests access `Object` already anyway to verify deserialization so it would be hard to miss

ob-stripe:
> Should we do StripeEntity and StripeEntityWithId and revert the rest? 

To be clear, I was suggesting going one step further and get rid of `StripeEntityWithId` as well. All model classes would derive from `StripeEntity` and those that have an ID would implement `IHasId` and explicitly declare the `Id` property.

> Do we need an interface for object/metadata/deleted? 

Right now we have no need for the latter two. Having an `IHasObject` will help me with a clean implementation of #1313: a generic converter that needs to look at the `object` property to choose which class to instantiate could use such an interface.

remi-stripe:
I do like the distinction between `StripeEntity` and `StripeEntityWithId`. I think it's worth having it because it makes it clear what should have an Id so I guess we're mostly back to the old state right?

ob-stripe:
I'd argue that if the goal is to make it clear what resources should have an ID, then using the `IHasId` interface directly would be a better fit since telling you what should go into a class is exactly what interfaces are for :)

remi-stripe:
But `IHasId` requires adding the `Id` property everywhere and it's just repetitive. But yeah sorry I'm running out of opinion on that one to be honest :(
I don't really feel strongly so I trust you to do the right choice (easy to say I know!)

ob-stripe:
When it comes to model classes, I think explicitness is more valuable than DRY. Having each model class directly define all of its properties makes it easier to compare the model to a reference (e.g. the API ref or the OpenAPI spec). I also think it's cleaner to have both `IHasId` and `IHasObject` interfaces (and future `IHasFoo` interfaces, if we ever need them) be added to model classes in the same way.

ptal @remi-stripe

src/Stripe.net/Entities/Events/EventRequest.cs
```diff
@@ -2,8 +2,11 @@ namespace Stripe
 {
     using Newtonsoft.Json;
 
-    public class EventRequest : StripeEntityWithId
+    public class EventRequest : StripeEntity, IHasId
```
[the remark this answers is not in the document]
ob-stripe: 👍  Agreed.

src/Stripe.net/Entities/EphemeralKeys/EphemeralkeyAssociatedObject.cs
```diff
@@ -2,8 +2,11 @@ namespace Stripe
 {
     using Newtonsoft.Json;
 
-    public class EphemeralKeyAssociatedObject : StripeEntityWithId
+    public class EphemeralKeyAssociatedObject : StripeEntity, IHasId
```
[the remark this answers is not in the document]
ob-stripe: 👍  Agreed.

src/Stripe.net/Entities/Common/PaymentSource.cs
```diff
@@ -12,8 +12,11 @@ public enum PaymentSourceType
     }
 
     [JsonConverter(typeof(PaymentSourceConverter))]
-    public class PaymentSource : StripeEntityWithId
+    public class PaymentSource : StripeEntity, IHasId
     {
+        [JsonProperty("id")]
+        public string Id { get; set; }
```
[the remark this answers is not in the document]
ob-stripe: I'm planning on removing all these union classes in #1313, so this is kind of moot.

src/Stripe.net/Entities/ExternalAccounts/ExternalAccount.cs
```diff
@@ -10,8 +10,11 @@ public enum ExternalAccountType
     }
 
     [JsonConverter(typeof(ExternalAccountConverter))]
-    public class ExternalAccount : StripeEntityWithId
+    public class ExternalAccount : StripeEntity, IHasId
     {
+        [JsonProperty("id")]
+        public string Id { get; set; }
+
```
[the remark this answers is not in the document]
ob-stripe: Same, this will go away in #1313.

src/Stripe.net/Entities/Orders/ShippingMethod.cs
```diff
@@ -2,8 +2,11 @@ namespace Stripe
 {
     using Newtonsoft.Json;
 
-    public class ShippingMethod : StripeEntityWithId
+    public class ShippingMethod : StripeEntity, IHasId
     {
+        [JsonProperty("id")]
+        public string Id { get; set; }
```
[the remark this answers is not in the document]
ob-stripe: 👍 Agreed.

src/Stripe.net/Entities/Common/PaymentSource.cs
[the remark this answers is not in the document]
remi-stripe: no you're moot

ob-stripe:
ptal @remi-stripe

PR #1321, opened 2018-10-05
Replace BalanceTransactionSource class by IBalanceTransactionSource interface

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Like #1320, but for `BalanceTransactionSource`.

I also fixed an issue with `StringOrObject`. Since the interface deserializer can now return `null` (in case of an unknown object), we need to handle this case in `StringOrObject` otherwise an exception would be raised when trying to access the `Id` attribute on the `null` object.


src/Stripe.net/Infrastructure/StringOrObject.cs
```diff
@@ -16,15 +16,24 @@ public static void Map(object value, Action<string> updateId, Action<T> updateOb
...
+                    updateId(item.Id);
+                    updateObject(item);
+                }
+                else
+                {
+                    // We were unable to deserialize the object, but make a last attempt to grab
+                    // the ID from the raw JObject.
+                    var id = ((JObject)value).SelectToken("id")?.ToString();
+                    updateId(id);
+                    updateObject(default(T));
```
remi-stripe: I'm glad you found that, I was planning to ask you about it since I felt we "lost" the ability to handle "unknown" types.

Out of curiosity, how does that work? `IBalanceTransactionSource` is an interface. It implements `IHasId` but it does not have an `Id` property. Where does the property go? I see in the test that it seems to go in `SourceId` but I don't think it's correct. If you have `Source` expanded, you _expect_ `Source` to not be null and it would not make sense to write code that falls back to `SourceId` right?

Also we likely (absolutely?) want to deserialize `Object` too since it's the only way for someone to know what type is not handled properly?

ob-stripe: Previously, when encountering an unknown type:
- `SourceId` is set to the ID
- `Source` is set to a `BalanceTransactionSource` instance with the following properties:
    - `Type` is set to `BalanceTransactionSourceType.Unknown`
    - all other properties are `null`

Now, when encountering an unknown type:
- `SourceId` is set to the ID
- `Source` is set to `null`

So basically, instead of checking that `Source.Type != BalanceTransactionSourceType.Unknown`, now you just check that `Source != null`.

In both cases (before and after), you can check the unknown object type and the rest of the JSON via the `StripeResponse.ResponseJson` property on the top-level `BalanceTransactiopn` instance.

ob-stripe: No worries. I also think it'd be useful to make the unknown object's name explicit in some way, but I think we'd need to add some logging capabilities to the library first. I'll try and look into that.

PR #1329, opened 2018-10-09
Use an interface for PaymentIntentSourceAction

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Similar to #1320 but for `PaymentIntentSourceAction`. The logic is slightly different because we can't deserialize the polymorphic attribute (`value`) based on its contents (like we can for Stripe objects, by looking at the `object` attribute). Instead, we need to look at the value of the `type` attribute that's at the same nesting level as `value`, so the converter has to be associated with the class that encompasses both attributes.

I've also removed `None` as a possible value in the enum -- instead, the entire `NextSourceAction` attribute will be set to `null` (like in the API).


src/Stripe.net/Infrastructure/JsonConverters/PaymentIntentSourceActionConverter.cs
```diff
@@ -22,24 +22,24 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
-            if (incoming.SelectToken("type")?.ToString() == "authorize_with_url")
+            if (type.Equals("authorize_with_url"))
             {
                 sourceAction.Type = PaymentIntentSourceActionType.AuthorizeWithUrl;
-                sourceAction.AuthorizeWithUrl = Mapper<PaymentIntentSourceActionAuthorizeWithUrl>.MapFromJson(incoming.SelectToken("value")?.ToString());
-            }
-            else
-            {
-                sourceAction.Type = PaymentIntentSourceActionType.Unknown;
+                sourceAction.Value = Mapper<PaymentIntentSourceActionAuthorizeWithUrl>.MapFromJson(incoming.SelectToken("value")?.ToString());
```
remi-stripe: OOC can't you do the same logic as the other one where you map based on `type` but use `value` for the deserialization? Or is it just that we have one example so it's cleaner this way for now?

src/Stripe.net/Infrastructure/JsonConverters/PaymentIntentSourceActionConverter.cs
```diff
@@ -22,24 +22,24 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
-                sourceAction.Type = PaymentIntentSourceActionType.None;
-                return sourceAction;
+                return null;
             }
 
             var incoming = JObject.Load(reader);
+            var sourceAction = new PaymentIntentSourceAction()
+            {
+                Type = PaymentIntentSourceActionType.Unknown,
+                Value = default(IPaymentIntentSourceActionValue),
```
remi-stripe: What does `default` mean here? Shouldn't `Value` be `null`?

src/StripeTests/Entities/PaymentIntents/PaymentIntentTest.cs
```diff
@@ -61,7 +61,9 @@ public void DeserializeNextSourceAction()
             Assert.Equal("payment_intent", intent.Object);
 
             Assert.Equal(PaymentIntentSourceActionType.AuthorizeWithUrl, intent.NextSourceAction.Type);
-            Assert.Equal("https://stripe.com", intent.NextSourceAction.AuthorizeWithUrl.Url);
+            var authorizeWithUrl = intent.NextSourceAction.Value as PaymentIntentSourceActionAuthorizeWithUrl;
+            Assert.NotNull(authorizeWithUrl);
+            Assert.Equal("https://stripe.com", authorizeWithUrl.Url);
```
remi-stripe: Should we test with `None`?

Issue #1333, opened 2018-10-15
StripeSubscriptionUpdateOptions docs are wrong


kspearrin:
See here: https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/Subscriptions/StripeSubscriptionUpdateOptions.cs#L36

I think your docs summaries for `CancelAtPeriodEnd` and `ProrationDate` are flipped.

remi-stripe:
@kspearrin Thank you for catching! We'll fix it in the next release!

remi-stripe:
@kspearrin I just opened a PR for this. It will be part of the next major version which we plan to ship in the next week or two

PR #1334, opened 2018-10-15
Fix docs and move params to the shared class for Create and Update Subscription

remi-stripe:
This fixes https://github.com/stripe/stripe-dotnet/issues/1333 as the docs were inverted for 2 of the parameters. They are also moved to the shared class as those parameters are now supported on both Create and Update endpoints.

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
`BillingCycleAnchor` is already present in `SubscriptionCreateOptions`, just without the additional properties for `"now"` and `"unchanged"`. I think we should keep this as is (since you cannot use these special values in creation requests) and only move `CancelAtPeriodEnd` in the shared class.

remi-stripe:
@ob-stripe Yep you're right I mixed things between that one and billing when I read. PTAL

PR #1343, opened 2018-10-18
Add support for new methods on the Invoice resource

remi-stripe:
This is a breaking change as it will be linked to a new API version since we are removing some fields.

cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Services/Invoices/InvoiceCreateOptions.cs
+++ b/src/Stripe.net/Services/Invoices/InvoiceCreateOptions.cs
@@ -16,6 +16,9 @@ namespace Stripe
         [JsonProperty("application_fee")]
         public long? ApplicationFee { get; set; }
 
+        [JsonProperty("auto_advance")]
+        public bool? AutoAdvance { get; set; }
+
         /// <summary>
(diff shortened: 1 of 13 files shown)


remi-stripe:
Removing ob since this can not be released before we have the new API version ready.

src/Stripe.net/Services/Invoices/InvoiceCreateOptions.cs
```diff
@@ -38,6 +38,9 @@ public class InvoiceCreateOptions : BaseOptions
         [JsonProperty("days_until_due")]
         public long? DaysUntilDue { get; set; }
 
+        [JsonProperty("default_source")]
+        public string DefaultSource { get; set; }
+
```
brandur-stripe: If `AutoAdvance` belongs in `InvoiceParams` in Go, doesn't it belong in either create or update here too?

remi-stripe: Awesome, nice catch. I totally missed those 2 when I did the PR. Fixed, PTAL

Issue #1346, opened 2018-10-22
Shared Customers does not support providing specific payment method


anelder-stripe:
Currently you can’t select a specific payment method when using the Shared Customers approach to re-tokenizing an existing customer on a connected account with stripe-dotnet.  It will only tokenize the default payment method.

https://stripe.com/docs/connect/shared-customers#making-tokens

https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/Tokens/StripeTokenCreateOptions.cs

We should probably add a `CustomerSource`-argument or something that maps to the `source`-argument of the create token endpoint for shared customers.

remi-stripe:
@anelder-stripe When you creating a token for a shared customer and specific card you would pass the `card` parameter set to the card id `card_ABC` of that customer along with the customer id `cus_123` in the `customer` parameter. Similarly if you want to tokenize a specific bank account you would pass the `bank_account` parameter. Both of those are supported as expected in the class today and there is no parameter `source` on token creation for shared customers.

If you have a Source on the customer, `src_1B2C` then you have to use the Shared Card Source approach instead (which only works for card sources today) of passing the `original_source` parameter on source creation as documented here: https://stripe.com/docs/sources/connect#shared-card-sources

remi-stripe:
Turns out @anelder-stripe was right and we misunderstood the report. Today you can not pass a specific card id or bank account id with `TokenCreateOptions`. It only supports hashes with real data in it.

ob-stripe:
Fixed in 20.1.0.

PR #1348, opened 2018-10-22
Properly support shared customers for a specific source

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/1346 which is a bug I just introduced because I removed the ability to pass `TokenId` in `StripeCreditCardOptions`. Ultimately, the old way was confusing and did not make a lot of sense. Now we explicitly can pass a specific card or bank account id in `TokenCreateOptions`.

r? @ob-stripe 
cc @anelder-stripe @stripe/api-libraries


ob-stripe:
Released as 20.1.0.

PR #1355, opened 2018-10-24
Fix handling of Expand* properties in list requests

### LINKED ISSUE #1351: Expand product property when listing plans
Using Windows 10, .NET Core 2.1 and Stripe.net 20.1.0.

After upgrading to Stripe.net 20.1.0, an exception is thrown when retrieving plans and setting `PlanService.ExpandProduct` field to true.

The error message is: "This property cannot be expanded (product). You may want to try expanding 'data.product' instead".

Using the following code:
```c#
var service = new PlanService();
service.ExpandProduct = true;
service.ListAsync(planListOptions);
```

A workaround to this issue is by appending "data.product" to `PlanListOptions.Expand` list and leaving the `PlanService.ExpandProduct` property set to false.

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Fix handling of `Expand*` properties in list requests. After #1308, we no longer passed the `isListMethod` value to `ApplyAllParameters`, which caused the `data.` prefix to not be prepended in list requests.

Fixes #1351.

### DIFF
--- a/src/Stripe.net/Services/Account/AccountService.cs
+++ b/src/Stripe.net/Services/Account/AccountService.cs
@@ -58,12 +58,12 @@ namespace Stripe
 
         public virtual Account GetSelf(RequestOptions requestOptions = null)
         {
-            return this.GetRequest<Account>($"{Urls.BaseUrl}/account", null, requestOptions);
+            return this.GetRequest<Account>($"{Urls.BaseUrl}/account", null, requestOptions, false);
         }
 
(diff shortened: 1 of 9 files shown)


ob-stripe:
@remi-stripe I added a test by setting `ExpandDefaultSource` to `true` in the tests for the customer list methods. If the library fails to add the `data.` prefix, stripe-mock would return an error and the test would fail. It's not ideal, but better than nothing.

ob-stripe:
Released as 20.2.0.

PR #1358, opened 2018-10-25
Ignore culture when encoding parameters

### LINKED ISSUE #1357: Issue with post of decimal data when decimal separator is not dot (again)
I see this has been fixed before in #801 but it has re-appeared. Probably in #1276. This occurs in v20.1.0 and v20.0.0 but not in v19.10.0.

I'd make a pull request of this but I'm having some issues with the solution so here's just my guess at the fix in `RequestStringBuilder` in `private static List<Parameter> FlattenParamsValue(object value, string keyPrefix)`:
```
else if (value is decimal dec)
{
    flatParams = new List<Parameter>();
    flatParams.Add(new Parameter(keyPrefix, dec.ToString(CultureInfo.InvariantCulture)));
}
```

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Ignore culture when encoding parameters, to ensure they're encoded like the API expects.

Fixes #1357.

### DIFF
--- a/src/Stripe.net/Infrastructure/Middleware/RequestStringBuilder.cs
+++ b/src/Stripe.net/Infrastructure/Middleware/RequestStringBuilder.cs
@@ -100,7 +100,7 @@ namespace Stripe.Infrastructure.Middleware
                 DateTime? dateTime = (DateTime)value;
                 if (dateTime.HasValue)
                 {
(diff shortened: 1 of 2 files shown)


ob-stripe:
I disabled the test for `netcoreapp1.1` because I couldn't find to make it work. It's tested with `netcoreapp2.1` and `net452`, so that's good enough.

ob-stripe:
Released as 20.2.0.

PR #1368, opened 2018-11-05
Bring back converter for dates in resources to fix deserialization

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Okay, I pushed another commit that adds `JsonConverter` attributes for two other cases:
- `Event.Data.Object`
- `Charge.Source`

Both of these are always expanded, so there's no issue with `StringOrObject` here. This should allow users to deserialize events and charges.

Also, I realized that the `StripeList<IExternalAccount>` property on account objects is more or less in the same situation as expandable attributes: we can't apply the converter directly (because it's a `StripeList<IExternalAccount>`, not an `IExternalAccount`). We should ensure that we find a good fix for this case as well.

remi-stripe:
@ob-stripe Thanks for adding the extra tests! Approving!

PR #1373, opened 2018-11-07
Add support for `flat_amount` on Plan tiers.

### LINKED ISSUE #1375: Deserialization error on Pricing plans with null Price per Unit or Flat Fee
Building out a test "create customer with subscription and products" method I kept getting a deserialization error when I would attempt to create a subscription with plans that had either nulls in the _price per unit_ or _flat fee_. The UI allows for creating those kinds of plans and doesn't throw an error in that instance. However, if I change those values from null to 0 I can go through the process successfully.

It would appear that those fields need to be nullable.

image

Sample method:
```c#
// [...] abbreviated for brevity
// create subscription
var subscriptionItems = new List<SubscriptionItemOption>();
foreach (var product in configuration.DefaultSubscriptionProducts)
    subscriptionItems.Add(new SubscriptionItemOption {
        PlanId = product,
        Quantity = 0
    });
var subscription = subscriptionService.Create(new SubscriptionCreateOptions {
    CustomerId = customer.Id,
    Items = subscriptionItems,
    BillingCycleAnchor = beginningOfNextMonth.DateTime,
    Metadata = metadata,
});
// [...]
```

**OS Version**: Windows 10
**.net Version**: .net core 2.0
**Stripe Version**: 20.4.0

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 

Fixes #1375.

### DIFF
--- a/src/Stripe.net/Entities/Plans/PlanTier.cs
+++ b/src/Stripe.net/Entities/Plans/PlanTier.cs
@@ -4,8 +4,11 @@ namespace Stripe
 
     public class PlanTier : StripeEntity
     {
+        [JsonProperty("flat_amount")]
+        public long? FlatAmount { get; set; }
+
         [JsonProperty("unit_amount")]
-        public long UnitAmount { get; set; }
+        public long? UnitAmount { get; set; }
 
         [JsonProperty("up_to")]
         public long? UpTo { get; set; }
--- a/src/Stripe.net/Services/Plans/PlanTierOptions.cs
+++ b/src/Stripe.net/Services/Plans/PlanTierOptions.cs
@@ -6,6 +6,9 @@ namespace Stripe
 
     public class PlanTierOptions : INestedOptions
     {
+        [JsonProperty("flat_amount")]
+        public long? FlatAmount { get; set; }
+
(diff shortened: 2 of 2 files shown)


src/Stripe.net/Entities/Plans/PlanTier.cs
```diff
@@ -4,6 +4,9 @@ namespace Stripe
 
     public class PlanTier : StripeEntity
     {
+        [JsonProperty("flat_amount")]
+        public long FlatAmount { get; set; }
```
ob-stripe: This should be nullable.

(`UnitAmount` should be nullable too btw, but I guess that'd be a breaking change, so let's fix in the next major version.)

ob-stripe: Well, a user just reported a breakage because `UnitAmount` is not nullable (#1375) so let's try and fix that ASAP.

remi-stripe: Well this is going in a breaking release already so fixing now.

remi-stripe:
@ob-stripe Fixed, PTAL

PR #1377, opened 2018-11-08
URL-encode dictionary keys

### LINKED ISSUE #1376: Metadata encoding issue with certain characters
As noted in #1374, if you attempt to create an object in the API with metadata with certain characters(such as `#`), it results in a malformed request. The API itself accepts these characters so it appears to be an issue with the library. The following code reproduces the problem:

var chargeOptions = new ChargeCreateOptions()
{
   Amount = 2000,
   Currency = "usd",
   SourceId = "tok_visa",
   Metadata = new Dictionary<string, string>{
      {"Invoice #", "42"}
    }
};

ChargeService chargeService = new ChargeService();

Charge charge = chargeService.Create(chargeOptions);
Assert.NotNull(charge);

request ID : req_TSGAWLekqimvqv 

It looks like the raw POST data sent is :
`amount=2000&currency=usd&metadata[Invoice%20`

I'm not sure what the cause here is, maybe the [string interpolation](https://github.com/stripe/stripe-dotnet/blob/cd3a69a9e690587c4fa7fc12e2053d33eeedf958/src/Stripe.net/Infrastructure/Middleware/RequestStringBuilder.cs#L382)?

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries @karlr-stripe 

Ensures that dictionary keys (typically, metadata keys) are URL-encoded. Otherwise, if a metadata key contains a hash mark character `#`, the requestor interprets it as a fragment delimiter and doesn't send the character and the rest of the string to the server.

Fixes #1376.

### DIFF
--- a/src/Stripe.net/Infrastructure/Middleware/RequestStringBuilder.cs
+++ b/src/Stripe.net/Infrastructure/Middleware/RequestStringBuilder.cs
@@ -181,7 +181,7 @@ namespace Stripe.Infrastructure.Middleware
 
             foreach (KeyValuePair<string, object> entry in dictionary)
             {
-                string key = entry.Key;
+                string key = WebUtility.UrlEncode(entry.Key);
                 object value = entry.Value;
 
                 string newPrefix = NewPrefix(key, keyPrefix);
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released as 20.4.1.

PR #1387, opened 2018-11-12
Add XML comments on interfaces

### LINKED ISSUE #1369: v20 DefaultSource changed to interface.  Any advice on migrating?
Hi,

It's not clear how to solve this breaking change.  My v19 code creates a customer with:
Stripe.Customer stripeCustomer = new Stripe.CustomerService().Create(customerOptions);

Now DefaultSource properties error, e.g. stripeCustomer.DefaultSource.Card has on Card property because DefaultSource is now a IPaymentSource.

I guess I need to cast to the correct interface.  What is the recommended way to do that?

And for similar problems, where can we find out which interface we need to cast to?

Thanks for an awesome library!

Eamonn

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add XML comments on `IBalanceTransactionSource`, `IExternalAccount` and `IPaymentSource` to provide an explicit list of concrete classes that implement the interfaces, so users know which types they should test when receiving one of these interfaces.

Fixes #1369.

### DIFF
--- a/src/Stripe.net/Entities/_interfaces/IBalanceTransactionSource.cs
+++ b/src/Stripe.net/Entities/_interfaces/IBalanceTransactionSource.cs
@@ -2,6 +2,42 @@ namespace Stripe
 {
     /// <summary>
     /// Resources that implement this interface can appear as sources in balance transactions.
+    /// <para>Possible concrete classes:</para>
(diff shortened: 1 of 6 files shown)


ob-stripe:
Released as 21.2.0.

PR #1388, opened 2018-11-12
Add support for Terminal resources

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries @stevene-stripe


src/Stripe.net/Services/Terminal/ConnectionTokens/ConnectionTokenService.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        public ConnectionTokenService(string apiKey)
+            : base(apiKey)
+        {
+        }
+
+        public override string BasePath => "/terminal/connection_tokens";
+
+        public virtual ConnectionToken Create(ConnectionTokenCreateOptions options, RequestOptions requestOptions = null)
```
ob-stripe: I wonder if we should provide overloads with no options. It seems a little bit silly to force users to create a `ConnectionTokenCreateOptions` instance that contains nothing.

remi-stripe: I hear you but then it becomes a mess of overloads and I'm not a big fan, we saw what it did to stripe-java.

remi-stripe:
@ob-stripe Already did for the `StripeObjectConverter` part, likely while you reviewed

src/Stripe.net/Services/Terminal/ConnectionTokens/ConnectionTokenService.cs
ob-stripe: Just tested it, and we can simply set `null` as the default value for the `options` argument. It still satisfies the interface and makes the argument optional without requiring an overload.

remi-stripe: fixed, PTAL

ob-stripe:
Released as 21.2.0.

PR #1389, opened 2018-11-12
Add support for providing client_secret when retrieving sources and payment intents

### LINKED ISSUE #1378: RequestOptions missing ClientSecret properties
I've read in the documentation che Stripe.NET is intended to be used client side, and so any source retrieved with SourceService.Get method should have the private key provided. Calling a Source using the public key and source id returns a "missing client secret" error. 
This behaviour doesn't consider the use of Stripe.NET in Xamarin as a client side library for Stripe enabled mobile application. Here we need to use Public key to tokenize a credit card, and use a 3D Secure source to check if it's available. We cannot retrieve back the 3D Secure source to check if the authorization has  been provided or not. Right now we switched back to the REST API that allow passing the client secret. 

Suggested changes are

- Add a property ClientSecret to the RequestOptions Object
- In Requestor.GetRequestMessage (or in SourceInstance.InstanceUrl, or where applicable) read the ClientSecret value and add it to the  request

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add support for providing `client_secret` when retrieving sources and payment intents.

Fixes #1378.

### DIFF
--- a/src/Stripe.net/Services/PaymentIntents/PaymentIntentGetOptions.cs
+++ b/src/Stripe.net/Services/PaymentIntents/PaymentIntentGetOptions.cs
@@ -1,0 +1,15 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class PaymentIntentGetOptions : BaseOptions
+    {
+        /// <summary>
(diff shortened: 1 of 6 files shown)


ob-stripe:
Released as 21.2.0.

PR #1395, opened 2018-11-15
Fix `DateTime` deserialization

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Removes `DateTimeConverter` from the list of global converters and apply it directly to `DateTime` properties via the `JsonConverter` attribute.

Updates tests to use `JsonConvert.DeserializeObject<>` instead of `Mapper<>.MapFromJson` to deserialize objects wherever possible.


ob-stripe:
Released as 21.4.1.

PR #1396, opened 2018-11-16
Fix deserialization logic for polymorphic types

### LINKED ISSUE #1178: Metadata containing date is parsed differently when in expanded entity
Given a metadata on a customer containing the output of `DateTimeOffset.ToString("o")`, as in "2018-04-30T19:03:07.6062218+00:00", calling `StripeSubscriptionService.ListAsync` with `ExpandCustomer` set to true returns a `Customer.Metadata[key]` containing a different date (local and parsed), like "04/30/2018 15:03:07".

Sample code:

```c#
    // Given each customer has a "trialStartedDate" metadata
    var subService = new StripeSubscriptionService() { ExpandCustomer = true };
    var custService = new StripeCustomerService();

    var subscriptions = await subService.ListAsync();
    var firstSubscription = subscriptions.FirstOrDefault();
    var customer = await custService.GetAsync(firstSubscription.CustomerId);

    Console.WriteLine($"{firstSubscription.Customer.Metadata["trialStartedDate"]}");
    Console.WriteLine($"{customer.Metadata["trialStartedDate"]}");

```

The output is:

> 04/30/2018 15:03:07

> 2018-04-30T19:03:07.6062218+00:00

Tested with version 13.3.1 and 15.8.0.
Tested with a .NET Core app on Windows and Linux.
Confirmed the JSON payload is the same, so the conversion occurs in the SDK.

ob-stripe:
r? @remi-stripe @talljoe-stripe 
cc @stripe/api-libraries 

Fixes the deserialization logic to properly handle polymorphic types, i.e. properties that are declared with an interface as their type rather than a concrete class.

Summary of changes:
- new `Stripe.Infrastructure.Util` class that contains:
    - a mapping of `object` values to concrete types, e.g. `"charge" => typeof(Stripe.Charge)`
    - a `GetConcreteType()` method that accepts a "potential type" and an `object` value:
        - if the potential type is a concrete type (not an interface), return it
        - otherwise, use the `object` value to find the matching concrete type, and return it (if it's compatible with the interface)
- `AbstractStripeObjectConverter`, `BalanceTransactionSourceConverter`, `ExternalAccountConverter`, `PaymentSourceConverter` and `StripeObjectConverter` are all gone
- new `StripeObjectConverter` deserializer (same name, but completely different logic) that uses `Util.GetConcreteType()` to figure out which concrete type to instantiate and populate
    - the converter is directly applied to properties with interface types (e.g. `EventData.Object`), which ensures that the class can be deserialized with `JsonConvert.DeserializeObject()`
    - the converter is also used as the `ItemConverterType` for `StripeList.Data` so that `StripeList` can be used with interfaces (e.g. `StripeList<IPaymentSource>` on `Customer`)
- `StringOrObject` class used for expandable fields also uses `Util.GetConcreteType()` to decide which type to pass to `JObject.ToObject()` for deserialization
- `Stripe.Mapper` now exposes `SerializerSettings` property that can be customized by users. Default settings are:
    - Use the `StripeObjectConverter` deserializer everywhere (which ensures that interfaces can always be deserialized)
    - Don't parse dates (which fixes #1178)

Test changes:
- Use `JsonConvert.DeserializeObject<>` instead of `Mapper<>.MapFromJson` wherever possible (i.e. when the type is not an interface)
- Add a new wholesome test to ensure that all model classes that implement `IHasObject` have an entry in the `Util.ObjectsToTypes` dictionary
- Add a new test to ensure that strings that contain ISO dates are not changed
- Add a test to ensure that serializing/deserializing/reserializing produces consistent results


src/Stripe.net/Infrastructure/Public/Mapper.cs
```diff
@@ -73,5 +67,17 @@ private static void ApplyStripeResponse(string json, StripeResponse stripeRespon
...
+
+        private static JsonSerializerSettings InitSerializerSettings()
+        {
+            return new JsonSerializerSettings
+            {
+                Converters = new List<JsonConverter>
+                {
+                    new StripeObjectConverter(),
+                },
+                DateParseHandling = DateParseHandling.None,
```
remi-stripe: Didn't you need to set something related to `CultureInfo` here too?

src/Stripe.net/Infrastructure/Util.cs
```diff
@@ -0,0 +1,109 @@
...
+            Type concreteType = null;
+
+            if (!string.IsNullOrEmpty(objectValue) &&
+                ObjectsToTypes.TryGetValue(objectValue, out concreteType))
+            {
+                // Found a concrete type matching the value of the `object` key, check if it's
+                // compatible with the interface.
+                if (potentialType.GetTypeInfo().IsAssignableFrom(concreteType.GetTypeInfo()))
+                {
+                    return concreteType;
```
remi-stripe: how could this one really happen? Since to exist in your Dictionary above you have to have a mapping to the corresponding class

src/StripeTests/Infrastructure/SerializationTest.cs
```diff
@@ -38,5 +37,17 @@ public void HandleNull()
...
+        public void Serialize()
+        {
+            var json = GetResourceAsString("api_fixtures.events.customer_updated.json");
+            var evt = JsonConvert.DeserializeObject<Event>(json);
+            var serialized = JsonConvert.SerializeObject(evt, Formatting.Indented);
+            var reserialized = JsonConvert.SerializeObject(
+                JsonConvert.DeserializeObject<Event>(serialized),
+                Formatting.Indented);
+            Assert.Equal(serialized, reserialized);
+        }
```
remi-stripe: Can we try with something other than Event? Like I worry that both serialization results would have `data` null and we would not see it.

Also, serializing twice only proves a subset of the case right? Let's say I have a JSON and deserialize to Account. If some properties are not supported then they would never be deserialized, so when I serialize the first time the json has only a subset of what was originally in the fixture. When I deserialize again, I end up with what I had the first time and not what was in the original json.

So I know that we can't reliably use a fixture from stripe-mock since every time a new property would be added to the resource that unrelated would break and force you to update the resource. But we could use a local fixture that should never be broken.

It still has the problem of comparing 2 JSON strings that are formatted differently, but I think it'd give us more confidence.

src/Stripe.net/Infrastructure/Public/Mapper.cs
ob-stripe: I'm not sure how Newtonsoft.Json uses the culture exactly (except for formatting date strings, but that's irrelevant since we're specifically disabling that feature) but according to the [docs](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonSerializerSettings_Culture.htm), the default value is `InvariantCulture` which is probably what we want.

src/Stripe.net/Infrastructure/Util.cs
ob-stripe: Let's say we're trying to deserialize an `IExternalAccount`, but the API returned `"object": "source"`. The mapping would return `Source` as the concrete type, but `Source` does not implement `IExternalAccount`. We could instantiate the type and populate it, but then the deserializer would crash when trying to assign the value to the property, since the types are incompatible.

src/StripeTests/Infrastructure/SerializationTest.cs
ob-stripe: It's not really possible to retrieve the input JSON once it's been deserialized into a model class, save for really simple classes and/or specifically formatted input JSON files that don't really match actual API payloads.

For instance, all deletable resources have a `bool? Deleted` attribute. When serializing, this will add `"deleted": null` to the JSON string, which wouldn't be present in an actual API payload.

Also, serialization of expandable fields is basically broken right now (and I think has always been).

We can (and probably should) improve serialization by fixing all those special cases, but this is out of scope for this PR.

ob-stripe:
r? @brandur-stripe 

Hey Brandur, mind taking a look at this one? I know you haven't been too involved in stripe-dotnet recently, but I'd like to get as many eyes as possible on this PR. Let me know if you have any questions or need more context.

src/Stripe.net/Infrastructure/Util.cs
```diff
@@ -0,0 +1,111 @@
+namespace Stripe.Infrastructure
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
+
+    internal static class Util
```
brandur-stripe: Not really a serious comment, but if the library had avoided a general `Util` class this long, it might be worth avoiding introducing one. Like make this called `StripeTypeConverter` or `StripeTypeInstantiator` or something like that instead.

The reason is essentially that if a `Util` is available, it's a very tempting place for people to throw underdesigned one-offs if they don't want to try and think of a better place to put it.

ob-stripe: Very good point. Most of our other libraries have a `Util` equivalent, but you're right that it's kind of an antipattern.

ob-stripe:
I've renamed `Util` to `StripeTypeRegistry` and added some tests for `GetConcreteType`.

Pulling this in!

PR #1405, opened 2018-11-25
Add IssuingCardId param to EphemeralKeyCreateOptions

jack-stripe:
We now accept an issuing_card parameter in the API when making an ephemeral key; this allows doing so from the .Net bindings.

r? @remi-stripe


remi-stripe:
Released as 21.5.0, thanks for the work @jack-stripe

PR #1408, opened 2018-11-28
Add missing properties on the Refund resource

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries


remi-stripe:
ah yes sorry, I did both branches and forgot that I needed it on both!

PR #1414, opened 2018-12-07
Add support for auto-pagination

### LINKED ISSUE #1038: Feature Request: Add support for auto-pagination
Is there an easy way to List out all customers or subscriptions?  Right now I am having to write this code which handles the pagination.  

image

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Adds support for auto-pagination.

Example usages:

```c#
var service = new BalanceTransactionService();
var listOptions = new BalanceTransactionListOptions
{
  Created = new DateRangeOptions
  {
    GreatherThanOrEqual = DateTime.Parse("2018-11-01T00:00:00Z"),
    LessThan = DateTime.Parse("2018-12-01T00:00:00Z"),
  },
  Limit = 100,
};

foreach (var balanceTransaction in service.ListAutoPaging(listOptions))
{
  // Do something with balanceTransaction
}
```

```c#
var service = new InvoiceService();
var lineItems = service.ListLineItemsAutoPaging("in_123").ToList();
```

The second example uses the LINQ method [`ToList()`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.tolist?view=netframework-4.7.2) to directly convert the `IEnumerable<LineItem>` returned by `InvoiceService.ListLineItemsAutoPaging()` into a `List<LineItem>`.

Fixes #1038.

### DIFF
--- a/src/Stripe.net/Infrastructure/Requestor.cs
+++ b/src/Stripe.net/Infrastructure/Requestor.cs
@@ -239,9 +239,13 @@ namespace Stripe.Infrastructure
         {
             var result = new StripeResponse
             {
-                RequestId = response.Headers.Contains("Request-Id") ? response.Headers.GetValues("Request-Id").First() : "n/a",
-                RequestDate = Convert.ToDateTime(response.Headers.GetValues("Date").First(), CultureInfo.InvariantCulture),
-                ResponseJson = responseText
+                RequestId = response.Headers.Contains("Request-Id") ?
+                    response.Headers.GetValues("Request-Id").First() :
+                    "n/a",
+                RequestDate = response.Headers.Contains("Date") ?
+                    Convert.ToDateTime(response.Headers.GetValues("Date").First(), CultureInfo.InvariantCulture) :
+                    default(DateTime),
+                ResponseJson = responseText,
             };
 
             return result;
--- a/src/Stripe.net/Services/_interfaces/IListable.cs
+++ b/src/Stripe.net/Services/_interfaces/IListable.cs
@@ -1,14 +1,17 @@
 namespace Stripe
 {
+    using System.Collections.Generic;
     using System.Threading;
     using System.Threading.Tasks;
 
     public interface IListable<T, O>
-        where T : IStripeEntity
+        where T : IStripeEntity, IHasId
         where O : ListOptions
     {
         StripeList<T> List(O listOptions = null, RequestOptions requestOptions = null);
 
         Task<StripeList<T>> ListAsync(O listOptions = null, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken));
+
+        IEnumerable<T> ListAutoPaging(O listOptions = null, RequestOptions requestOptions = null);
     }
 }
--- a/src/StripeTests/Resources/pageable_models/0.json
+++ b/src/StripeTests/Resources/pageable_models/0.json
@@ -1,0 +1,15 @@
+{
+  "data": [
+    {
+      "id": "pm_123",
+      "object": "pageablemodel"
+    },
+    {
+      "id": "pm_124",
+      "object": "pageablemodel"
+    }
+  ],
+  "has_more": true,
+  "object": "list",
+  "url": "/v1/pageable_models"
+}
--- a/src/StripeTests/Services/AutoPagingTest.cs
+++ b/src/StripeTests/Services/AutoPagingTest.cs
@@ -1,0 +1,110 @@
+namespace StripeTests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Net;
+    using System.Net.Http;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Moq;
+    using Moq.Protected;
+    using Newtonsoft.Json;
+    using Stripe;
+    using Xunit;
+
+    public class AutoPagingTest : BaseStripeTest
+    {
+        public AutoPagingTest(MockHttpClientFixture mockHttpClientFixture)
+            : base(mockHttpClientFixture)
+        {
+        }
+
+        [Fact]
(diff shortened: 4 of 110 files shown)


src/Stripe.net/Infrastructure/Requestor.cs
```diff
@@ -239,9 +239,13 @@ private static StripeResponse BuildResponseData(HttpResponseMessage response, st
...
             {
-                RequestId = response.Headers.Contains("Request-Id") ? response.Headers.GetValues("Request-Id").First() : "n/a",
-                RequestDate = Convert.ToDateTime(response.Headers.GetValues("Date").First(), CultureInfo.InvariantCulture),
-                ResponseJson = responseText
+                RequestId = response.Headers.Contains("Request-Id") ?
+                    response.Headers.GetValues("Request-Id").First() :
+                    "n/a",
+                RequestDate = response.Headers.Contains("Date") ?
+                    Convert.ToDateTime(response.Headers.GetValues("Date").First(), CultureInfo.InvariantCulture) :
+                    default(DateTime),
```
ob-stripe: I couldn't find a way to mock response headers, so I made the `Date` header optional. I think that's better anyway.

src/Stripe.net/Services/_interfaces/IListable.cs
```diff
@@ -4,7 +4,7 @@ namespace Stripe
     using System.Threading.Tasks;
 
     public interface IListable<T, O>
-        where T : IStripeEntity
+        where T : IStripeEntity, IHasId
```
ob-stripe: We need the resource to have an ID in order to be able to paginate. I also added `IHasId` as a constraint on other interfaces where it makes sense to do so (i.e. everything except create).

src/StripeTests/BaseStripeTest.cs
```diff
@@ -37,9 +38,11 @@ public BaseStripeTest()
             var initialized = initializer.Value;
 
             // Reset the mock before each test
-            mockHandler.Invocations.Clear();
+            MockHandler.Reset();
         }
 
+        protected static Mock<HttpClientHandler> MockHandler { get => mockHandler; set => mockHandler = value; }
```
ob-stripe: I needed to expose the mock to subclasses because of advanced mocking needs for the auto-paginator test. Fields should be private, so I wrapped it into a protected property. That's apparently how you're supposed to do this in C# (i.e. StyleCop told me to).

src/StripeTests/Services/AutoPagingTest.cs
```diff
@@ -0,0 +1,107 @@
...
+    using System.Net.Http;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Moq;
+    using Moq.Protected;
+    using Newtonsoft.Json;
+    using Stripe;
+    using Xunit;
+
+    public class AutoPagingTest : BaseStripeTest
```
ob-stripe: This test seems pretty complex, but it's actually pretty straightforwards:
1. set up 3 stubbed requests with file fixtures
2. call `ListAutoPaging`
3. check invocations to make sure the correct parameters were sent

src/StripeTests/Resources/pageable_models/0.json
```diff
@@ -0,0 +1,15 @@
...
+      "object": "pageablemodel"
+    },
+    {
+      "id": "pm_124",
+      "object": "pageablemodel"
+    }
+  ],
+  "has_more": true,
+  "object": "list",
+  "url": "/v1/pageable_models"
```
remi-stripe: OOC aren't we supposed to have `?starting_after=pm_124` in this URL?

src/StripeTests/Services/AutoPagingTest.cs
```diff
@@ -0,0 +1,105 @@
...
+            };
+            var models = service.ListAutoPaging(options).ToList();
+
+            // Check results
+            Assert.Equal(5, models.Count);
+            Assert.Equal("pm_123", models[0].Id);
+            Assert.Equal("pm_124", models[1].Id);
+            Assert.Equal("pm_125", models[2].Id);
+            Assert.Equal("pm_126", models[3].Id);
+            Assert.Equal("pm_127", models[4].Id);
```
remi-stripe: So auto-paging basically builds the entire list in ram no matter what? You can't auto page while also getting items one by one and not filling your RAM with all your BTs history?

ob-stripe:
> Is there a way we don't have to add new prototypes for this? I assume not but was hoping you could do an autoPaginate = true that defaults to false?

No, because the types that are returned are different. When you use the existing `List` method, you get a `StripeList<Resource>` object. Auto-pagination lets you iterate directly over the `Resource` objects, bypassing the `StripeList` wrapper.

src/StripeTests/Resources/pageable_models/0.json
ob-stripe: No, the API doesn't include `starting_after` parameters in `url` attributes.

src/StripeTests/Services/AutoPagingTest.cs
ob-stripe: You can. In these examples, I used `ToList()` to directly convert the iterator into a list, but you can also iterate over the objects one by one:

```c#
foreach (var balanceTransaction in balanceTransactionService.ListAutoPaging(listOptions))
{
  // Do something with balanceTransaction
}
```

remi-stripe: ah neat! Can you put a test for this in our test suite? I know people refer to the test suite a lot to find example. I guess this is where your `yield` applies. Fun how similar it is in each language you've shown me this approach!

ob-stripe:
Okay, I think this is ready to be merged. ptal @remi-stripe @anelder-stripe

ob-stripe:
Timing out and self-approving this one :)

PR #1417, opened 2018-12-10
Add support for verifying sources

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add support for verifying sources.

Fixes one of the remaining tasks in #1066.

### DIFF
--- a/src/Stripe.net/Services/Sources/SourceVerifyOptions.cs
+++ b/src/Stripe.net/Services/Sources/SourceVerifyOptions.cs
@@ -1,0 +1,14 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class SourceVerifyOptions : BaseOptions
+    {
+        /// <summary>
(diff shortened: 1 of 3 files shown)


src/Stripe.net/Services/Sources/SourceVerifyOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class SourceVerifyOptions : BaseOptions
+    {
+        /// <summary>
+        /// The values needed to verify the source.
+        /// </summary>
+        [JsonProperty("values")]
+        public List<string> Values { get; set; }
```
remi-stripe: I'm assuming this is supposed to be a string and not an int right?

ob-stripe: Yep, it's specified as an array of strings, presumably because the values can be more or less anything depending on the source's type.

PR #1418, opened 2018-12-11
Replace RecipientActiveAccount with BankAccount

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Remove the `RecipientActiveAccount` class.

Changes the type of `Recipient.ActiveAccount` to `BankAccount` instead.

Per the OpenAPI spec, this is the correct type.


ob-stripe:
I'm fairly sure that the API doesn't return these anyway. They're replaced by `status`.

Thanks!

PR #1419, opened 2018-12-11
Replace public fields with properties

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Replace public fields by private fields encapsulated by public properties.

> Public fields in public classes do not respect the encapsulation principle and has three main disadvantages:
> 
> Additional behavior such as validation cannot be added.
> The internal representation is exposed, and cannot be changed afterwards.
> Member values are subject to change from anywhere in the code and may not meet the programmer's assumptions.
> By using private fields and public properties (set and get), unauthorized modifications are prevented. Properties also benefit from additional protection (security) features such as Link Demands.
> 
> Note that due to optimizations on simple properties, public fields provide only very little performance gain.

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
+++ b/src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
@@ -7,7 +7,7 @@ namespace Stripe
 
     public static class StripeConfiguration
     {
-        public static string StripeApiVersion = "2018-11-08";
+        private static string stripeApiVersion = "2018-11-08";
 
         private static string apiKey;
         private static string apiBase;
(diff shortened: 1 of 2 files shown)


src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
```diff
@@ -19,6 +19,12 @@ static StripeConfiguration()
             StripeNetVersion = new AssemblyName(typeof(Requestor).GetTypeInfo().Assembly.FullName).Version.ToString(3);
         }
 
+        public static string StripeApiVersion
+        {
+            get { return stripeApiVersion; }
+            set { stripeApiVersion = value; }
```
remi-stripe: Does that mean anyone can set this?

ob-stripe: Yep. I also have doubts whether this should be allowed, but wanted to maintain the current behavior.

PR #1423, opened 2018-12-12
Replace constants with static read-only properties

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

It is considered bad practice to define public constants in C#. Static read-only properties should be used instead. Cf. [this SO answer](https://stackoverflow.com/a/755693/5307473).

### DIFF
--- a/src/Stripe.net/Constants/AccountType.cs
+++ b/src/Stripe.net/Constants/AccountType.cs
@@ -2,7 +2,8 @@ namespace Stripe
 {
     public static class AccountType
     {
-        public const string Custom = "custom";
-        public const string Standard = "standard";
+        public static string Custom => "custom";
+
+        public static string Standard => "standard";
     }
 }
--- a/src/Stripe.net/Constants/BankAccountHolderType.cs
+++ b/src/Stripe.net/Constants/BankAccountHolderType.cs
@@ -2,7 +2,8 @@ namespace Stripe
 {
     public static class BankAccountHolderType
     {
-        public const string Individual = "individual";
-        public const string Company = "company";
+        public static string Individual => "individual";
+
+        public static string Company => "company";
     }
 }
--- a/src/Stripe.net/Constants/Events.cs
+++ b/src/Stripe.net/Constants/Events.cs
@@ -7,481 +7,481 @@ namespace Stripe
         /// <summary>
         /// Occurs whenever an account status or property has changed.
         /// </summary>
-        public const string AccountUpdated = "account.updated";
+        public static string AccountUpdated => "account.updated";
 
         /// <summary>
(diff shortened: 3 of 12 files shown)


remi-stripe:
@ob-stripe Do you have something other than that SO answer to justify that change? It's a lot more verbose and I don't really get the upside mentioned in that specific answer.

ob-stripe:
This article explains why const strings are a bad idea for a C# library: https://www.stum.de/2009/01/14/const-strings-a-very-convenient-way-to-shoot-yourself-in-the-foot/

remi-stripe:
This is quite an old article (9 years) and it speaks of a really really specific edge-case that does not seem problematic with stripe-dotnet itself here since we own that constant and no one would be changing it.

Ultimately, I'm fine with the change itself. It's more wordy for us but it's not like we're using constants actively like we do in stripe-go so deferring to you on that one

ob-stripe:
Did some research and it turns out the `=>` operator can be used to define property getters (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-operator#expression-body-definition).

I updated the PR to use that syntax, making it much less wordy.

Nicholi:
Now we can't use any of these magic strings in case statements because they aren't constant. Most notably the Events types...

Was it worth it?

ob-stripe:
Hey, thanks for bringing this to our attention.

> The strings are not readonly - only static. Please canthese be changed to:
`public static readonly string xxx => "value";`?

The former constants are now declared as properties, not fields, so we can't apply the `readonly` modifier. They are already read-only, since they don't have setters.

```c#
public static string Foo => "foo";
```
is the same thing as:
```c#
public static string Foo { get { return "foo"; } };
```

Unfortunately, C# only accepts `const` values in `switch` statements. We didn't realize this when we pushed this change, else we'd have tagged it as a breaking change. I will retroactively flag this change as breaking, and mention in the [migration guide](https://github.com/stripe/stripe-dotnet/wiki/Migration-guide-for-v22).

Despite the added toil of converting `switch` statements to a bunch of `if / else if`, I still think this change is desirable. With constants, in the (admittedly unlikely, but not impossible) event that we need to change one of the values, library users would be forced to recompile in order to use the new values. With static readonly strings, users can simply use the new version of `Stripe.net.dll`.

Using static readonly properties instead of constants is also recommended by static analysis tools such as Sonar (https://github.com/SonarSource/sonar-dotnet/blob/master/sonaranalyzer-dotnet/tests/SonarAnalyzer.UnitTest/TestCases/PublicConstantField.cs).

marshall76963:
Thanks for the reply

Understand why you have made the change and must confess that I was not aware that the `readonly` modifier could not be applied.

I will bite the bullet and refactor my code accordingly!

PR #1432, opened 2018-12-14
Upgrade stripe-mock to 0.39.0

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/StripeTests/Services/Terminal/Locations/LocationServiceTest.cs
+++ b/src/StripeTests/Services/Terminal/Locations/LocationServiceTest.cs
@@ -38,7 +38,6 @@ namespace StripeTests.Terminal
 
             this.listOptions = new LocationListOptions
             {
-                OperatorAccount = "acct_123",
             };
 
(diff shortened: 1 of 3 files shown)


src/StripeTests/Services/Terminal/Locations/LocationServiceTest.cs
```diff
@@ -36,10 +36,7 @@ public LocationServiceTest()
                 DisplayName = "my location",
             };
 
-            this.listOptions = new LocationListOptions
-            {
-                OperatorAccount = "acct_123",
-            };
+            this.listOptions = new LocationListOptions();
```
remi-stripe: ooc why do you prefer the () to the {}? I prefer the latter so that we don't forget to remove the () when we re-add another parameter

ob-stripe: StyleCop rules force us to have `{}` on different lines so I went for brevity, but I don't care strongly one way or ther other. Updated the PR.

PR #1433, opened 2018-12-14
Fix JSON serialization of Stripe objects

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

- Adds `ToString` and `ToJson` methods to `StripeEntity`
- Exclude `StripeResponse` from JSON serialization
- Fix serialization of expandable attributes
- Don't serialize `deleted` attribute when it's `null`
- Add a test for serialization of expandable fields

I realize that the new getters for expandable fields add a lot of boilerplate. In the future, I'll probably replace the `object` declarations with an `ExpandableField<>` generic class and a custom serializer/deserializer, similar to what we do in Java, but for now this does the job.

### DIFF
--- a/src/Stripe.net/Entities/Cards/Card.cs
+++ b/src/Stripe.net/Entities/Cards/Card.cs
@@ -13,6 +13,7 @@ namespace Stripe
         public string Object { get; set; }
 
         #region Expandable Account
+        [JsonIgnore]
         public string AccountId { get; set; }
 
         [JsonIgnore]
@@ -21,6 +22,11 @@ namespace Stripe
         [JsonProperty("account")]
         internal object InternalAccount
         {
+            get
+            {
+                return this.Account ?? (object)this.AccountId;
+            }
+
             set
             {
                 StringOrObject<Account>.Map(value, s => this.AccountId = s, o => this.Account = o);
@@ -65,6 +71,7 @@ namespace Stripe
         public string Currency { get; set; }
 
         #region Expandable Customer
+        [JsonIgnore]
         public string CustomerId { get; set; }
 
         [JsonIgnore]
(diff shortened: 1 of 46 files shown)


ob-stripe:
Aw, Windows outputs CRLF line endings. Will fix.

src/Stripe.net/Entities/Cards/Card.cs
```diff
@@ -89,7 +101,7 @@ internal object InternalCustomer
         /// <summary>
         /// Whether this object is deleted or not.
         /// </summary>
-        [JsonProperty("deleted")]
+        [JsonProperty("deleted", NullValueHandling=NullValueHandling.Ignore)]
         public bool? Deleted { get; set; }
```
remi-stripe: Could we have `IStripeDeleted` interface or similar to ensure the setup of `Deleted` is always correctly done? Just worried we'll miss that custom setting in the future

ob-stripe: I'll add a wholesome test for this. To be clear, this is purely cosmetic: there is no functional difference between having `"deleted": null` or not having `"deleted"` at all in the JSON serialization.

ob-stripe:
Added a wholesome test, ptal @remi-stripe

src/Stripe.net/Entities/Cards/Card.cs
remi-stripe: ohhhhh, wdyt of removing then?

ob-stripe: I think it's nice to keep the serialization fairly close to what the API would return, and the wholesome test should ensure we stay consistent as we add more deletable resources in the future.

remi-stripe: I disagree but I see your point so let's do it your way!

PR #1434, opened 2018-12-14
Make Deleted nullable on all Stripe objects

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

`Deleted` was nullable on `Customer` but not on other Stripe model classes.


ob-stripe:
Going to self-approve as we chatted about this OOB.

PR #1436, opened 2018-12-14
Make object instances in tests readonly

ob-stripe:
It's good practice to declare fields that are only assigned in constuctors as `readonly`.


ob-stripe:
Tests still pass, and only test files were modified in this PR, so going to self-approve.

PR #1440, opened 2018-12-20
Add support for the Checkout Session resource

remi-stripe:
This resource is now ready to be released. The PR has been updated to match the latest spec.

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/Checkout/Sessions/Session.cs
+++ b/src/Stripe.net/Entities/Checkout/Sessions/Session.cs
@@ -1,0 +1,162 @@
+namespace Stripe.Checkout
+{
+    using System;
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class Session : StripeEntity, IHasId, IHasObject
+    {
+        /// <summary>
+        /// Unique identifier for the object.
+        /// </summary>
+        [JsonProperty("id")]
+        public string Id { get; set; }
+
+        /// <summary>
+        /// String representing the object’s type. Objects of the same type share the same value.
+        /// </summary>
+        [JsonProperty("object")]
+        public string Object { get; set; }
+
+        /// <summary>
+        /// Specify whether Checkout should collect the customer’s billing address. If set to
+        /// <c>required</c>, Checkout will always collect the customer’s billing address. If left
+        /// blank or set to <c>auto</c> Checkout will only collect the billing address when
+        /// necessary.
+        /// </summary>
+        [JsonProperty("billing_address_collection")]
+        public string BillingAddressCollection { get; set; }
+
+        /// <summary>
+        /// The URL the customer will be directed to if they decide to go back to your website.
+        /// </summary>
+        [JsonProperty("cancel_url")]
+        public string CancelUrl { get; set; }
+
+        /// <summary>
+        /// A unique string to reference the Checkout Session. This can be a customer ID, a cart
+        /// ID, or similar. It is included in the <c>checkout.session.completed</c> webhook and can
+        /// be used to fulfill the purchase.
+        /// </summary>
+        [JsonProperty("client_reference_id")]
+        public string ClientReferenceId { get; set; }
+
+        #region Expandable Customer
+
+        /// <summary>
+        /// ID of the customer this Session is for if one exists.
+        /// </summary>
+        [JsonIgnore]
+        public string CustomerId { get; set; }
+
+        [JsonIgnore]
+        public Customer Customer { get; set; }
+
+        [JsonProperty("customer")]
+        internal object InternalCustomer
+        {
+            get
+            {
+                return this.Customer ?? (object)this.CustomerId;
+            }
+
+            set
+            {
+                StringOrObject<Customer>.Map(value, s => this.CustomerId = s, o => this.Customer = o);
+            }
+        }
+        #endregion
+
+        /// <summary>
+        /// The email address used to create the customer object.
+        /// </summary>
+        [JsonProperty("customer_email")]
+        public string CustomerEmail { get; set; }
+
+        /// <summary>
+        /// The line items, plans, or SKUs that were purchased by the customer.
+        /// </summary>
+        [JsonProperty("display_items")]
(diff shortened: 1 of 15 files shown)


remi-stripe:
Updated most of the PR to match the new resource shape. Still a few things to fix in tests

remi-stripe:
cc @matt-stripe for awareness

src/Stripe.net/Entities/Checkout/Sessions/Session.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// The line items, plans, or SKUs that were purchased by the customer.
+        /// </summary>
+        [JsonProperty("display_items")]
+        public List<SessionDisplayItem> DisplayItems { get; set; }
+
+        /// <summary>
+        /// The email address used to create the customer object.
+        /// </summary>
+        [JsonProperty("customer_email")]
+        public string CustomerEmail { get; set; }
```
ob-stripe: This should be before `display_items`.

src/Stripe.net/Entities/Checkout/Sessions/SessionDisplayItem.cs
```diff
@@ -0,0 +1,49 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+
+    public class SessionDisplayItem : StripeEntity
```
ob-stripe: The spec defines four different types `payment_pages_checkout_session_checkout_session_resource_custom_display_item`, `payment_pages_checkout_session_checkout_session_resource_plan_display_item`, `payment_pages_checkout_session_checkout_session_resource_sku_display_item` and `payment_pages_checkout_session_checkout_session_resource_subscription_data_display_item` (whew). In theory we should define four different classes that implement a common interface `ISessionDisplayItem`, rather than a single class that implements all possible fields.

src/Stripe.net/Infrastructure/StripeTypeRegistry.cs
```diff
@@ -23,6 +23,7 @@ internal static class StripeTypeRegistry
             { "bank_account", typeof(BankAccount) },
             { "card", typeof(Card) },
             { "charge", typeof(Charge) },
+            { "checkout/session", typeof(Checkout.Session) },
```
ob-stripe:
```suggestion
            { "checkout.session", typeof(Checkout.Session) },
```

src/Stripe.net/Services/Checkout/SessionCreateOptions.cs
```diff
@@ -0,0 +1,86 @@
...
+        public string CancelUrl { get; set; }
+
+        /// <summary>
+        /// Specify whether Checkout should collect the customer’s billing address. If set to
+        /// <c>required</c>, Checkout will always collect the customer’s billing address. If left
+        /// blank or set to <c>auto</c> Checkout will only collect the billing address when
+        /// necessary.
+        /// </summary>
+        [JsonProperty("billing_address_collection")]
+        public string BillingAddressCollection { get; set; }
```
ob-stripe: Should be before `cancel_url`.

src/Stripe.net/Entities/Checkout/Sessions/SessionDisplayItem.cs
remi-stripe: Hmmmm that's new? This is not the way we did `PaymentMethod`, why didn't you mention this at the time? Is this something we should have done in that case too?

ob-stripe: You'll have to be more specific, which polymorphic field are you referring to exactly?

remi-stripe:
@ob-stripe Fixed what you found. PTAL

src/Stripe.net/Entities/Checkout/Sessions/SessionDisplayItem.cs
remi-stripe: After discussing it internally, we do want this to be one resource after all and the current implementation should be correct.

src/Stripe.net/Services/Checkout/SessionCreateOptions.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// </summary>
+        [JsonProperty("client_reference_id")]
+        public string ClientReferenceId { get; set; }
+
+        /// <summary>
+        /// ID of the customer this Checkout Session is for if one exists. May only be used with
+        /// <c>LineItems</c>. Usage with <c>SubscriptionData</c> is not yet available.
+        /// </summary>
+        [JsonProperty("customer")]
+        public string CustomerId { get; set; }
```
ob-stripe: I'm not sure if we're super consistent w/ regards to ordering the attributes alphabetically per attribute name (`CustomerId`) or per JSON name (`customer`), but in `Session`, `CustomerId` appears after `CustomerEmail`. Can you quickly check if there's a common style we use for this and update the PR to be consistent?

src/Stripe.net/Entities/Checkout/Sessions/SessionDisplayItem.cs
ob-stripe: 👍

src/Stripe.net/Services/Checkout/SessionCreateOptions.cs
remi-stripe: There's no style beyond "remi does not know his alphabet". Will fix.

remi-stripe: Ha okay I know why. It's because if something is expandable then it ends up appearing first. But I'd rather re-alphabetize when it becomes expandable.

remi-stripe:
Okay re-assigning to @mickjermsurawong-stripe to triple check based on the work we will do on stripe-java

remi-stripe:
Okay so now `customer`, `payment_intent` and `subscription` are expandable. So I had to change the PR again.

Also we will need a new release of stripe-mock to test this properly.

remi-stripe:
@ob-stripe Mick released a new stripe-mock that allows for expanding the right resource and I updated tests.

Can you have one last deep look so that we can release?

PR #1444, opened 2018-12-21
Refactor test infrastructure

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Alright, some context first. When I initialized the new `StripeTests` stripe-mock based suite a few months back, I set up a hack using a [`Lazy<>`](https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1?view=netframework-4.7.2) object to ensure the initialization (e.g. checking stripe-mock's version) is only done once.

This worked, but it's definitely not how this sort of things should be done with xUnit. Also, while this worked for initializing stuff, it doesn't work for cleaning stuff up afterwards. We handled that by simply not cleaning anything up.

This PR reworks the test suite to do things the proper way. All stripe-mock based tests are now grouped in a ["collection"](https://xunit.github.io/docs/running-tests-in-parallel#parallelism-in-test-frameworks). Tests in the same collection are run sequentially, and can share collection fixtures.

Two collection fixtures are added:
- `MockHttpClientFixture` is used to manage the mock HTTP client, for things like asserting requests (and in the future, stubbing requests)
- `StripeMockFixture` is used to manage stripe-mock

The most annoying part of this PR is that test classes that rely on a particular fixture must define a constructor to receive the fixture instance and pass it to the base constructor. I've added helpful error messages to detect when a test class relies on a fixture but didn't define the needed constructor.

The big motivation behind this PR is that I want to port https://github.com/stripe/stripe-ruby/pull/715 to stripe-dotnet, but to do so I need to be able to set up a global teardown function (in order to stop the stripe-mock process if the test suite starts its own).

### DIFF
--- a/src/StripeTests/Entities/Discounts/DiscountTest.cs
+++ b/src/StripeTests/Entities/Discounts/DiscountTest.cs
@@ -6,6 +6,11 @@ namespace StripeTests
 
     public class DiscountTest : BaseStripeTest
     {
+        public DiscountTest(StripeMockFixture stripeMockFixture)
+            : base(stripeMockFixture)
+        {
+        }
+
         [Fact]
         public void Deserialize()
         {
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ image: Visual Studio 2017
 environment:
   COVERALLS_REPO_TOKEN:
     secure: T0PmP8uyzCseacBCDRBlti2y9Tz5DL6fknea0MKWvbPYrzADmLY2/5kOTfYIsPUk
-  # If you bump this, don't forget to bump `MinimumMockVersion` in `BaseStripeTest.cs` as well.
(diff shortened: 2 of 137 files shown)


src/StripeTests/Entities/Discounts/DiscountTest.cs
```diff
@@ -6,6 +6,11 @@ namespace StripeTests
 
     public class DiscountTest : BaseStripeTest
     {
+        public DiscountTest(StripeMockFixture stripeMockFixture)
+            : base(stripeMockFixture)
+        {
+        }
+
```
ob-stripe: This test (and some others) doesn't rely on `GetFixture`, so it doesn't need to receive the `StripeMockFixture` instance. My codemod wasn't smart enough to figure that out though, and it's probably best to be consistent anyway.

Issue #1402, opened 2018-11-23
Error parsing JSON returned from webhook


steve-tapley:
Hi,
I've recently moved to version 21.4.1 from 19.x, and now I'm receiving errors when parsing some webhook events.

Below is some JSON that caused an error in the parser.   It looks like it falls over trying to generate a request object - from what it looks like to be an Id only,

I've tried both the ParseEvent and ConstructEvent methods - both fail identically.

```
{ "id": "evt_1DZUr64WLhpec2eZ8YxDi4sm", "object": "event", "api_version": "2014-06-17", "created": 1542941840, "data": { "object": { "id": "card_1DZUqq4WLhpec2eZr2GUSqJW", "object": "card", "address_city": null, "address_country": null, "address_line1": null, "address_line1_check": null, "address_line2": null, "address_state": null, "address_zip": null, "address_zip_check": null, "brand": "Visa", "country": "US", "customer": "cus_4iIYzLmgQznYlK", "cvc_check": "pass", "dynamic_last4": null, "exp_month": 9, "exp_year": 2019, "fingerprint": "tXS6DZYDEDNFUeFL", "funding": "credit", "last4": "4242", "metadata": { }, "name": "Gabe Newell", "tokenization_method": null } }, "livemode": false, "pending_webhooks": 1, "request": "req_7hFEZuZn4RwSm6", "type": "customer.card.deleted" }
```

Newtonsoft.Json.JsonSerializationException
Message = Error converting value "req_7hFEZuZn4RwSm6" to type 'Stripe.EventRequest'. Path 'request', line 36, position 33.

Hack "fix" until this has been sorted out:
```c#
var jObject = Newtonsoft.Json.JsonConvert.DeserializeObject<Newtonsoft.Json.Linq.JObject>(json);`
if (jObject["request"] != null)`
{
    jObject["request"] = null;
    json = jObject.ToString();
}
```

Cheers,
Steve

steve-tapley:
Hi @ob-stripe ,
Thanks for the info.  Yes, a better error message would have saved me some time.

Theres a few other issues here too, I think.

It would be nice to know what API versions a particular Stripe.net version is compatible with (e.g 2018-08-23 thru to 1081-11-08 for example).

At the moment, its difficult to test upgrades of Stripe.Net, because I cant set my test environment to be a different version of API to the live version.  
[...]

PR #1450, opened 2018-12-26
Better handling of API version mismatch when deserializing events

ob-stripe:
r? @remi-stripe 

Opening the PR for early feedback, but totally open to changes here.

As it is, this PR introduces two changes:
- support for event objects formatted with older API versions (< 2017-05-25), by handling the `request` attribute as if it were expandable (it's not, but it's functionally the same thing: pre 2017-05-25, `request` was a string, post 2017-05-25, `request` is a hash with an `id`)
- when using `Stripe.EventUtility.ConstructEvent` or `ParseEvent` to deserialize event objects (presumably, in a webhook handler), by default the library raises an exception if the API version on the event object doesn't match the API version the library is pinned to. This is a breaking change in the sense the code that used to "work" will now throw an exception ("work" in quotes because when the API versions don't match, the deserialization may fail silently if the format of the nested object changed: older attributes returned by the API will be silently ignored, newer attributes declared in the library will be `null`). This behavior can be disabled by passing `throwOnApiVersionMismatch = false` to both methods

Tests fail right now because I haven't fixed them yet. I'll fix / add more tests if you think we should move forward with this change.

Tentative fix for #1402.


ob-stripe:
> I like this approach a lot, especially since the impact on the resource itself is quite minimal. The issue is what happens the day we change the event shape a lot (which could happen)?

Yeah, we're kind of lucky here that the change can be handled like an expandable attribute. In the future, we may need to write a custom deserializer.

> Crazy idea: Can we deserialize to an InternalEvent class instead that specifically just handles the request property and nothing else?

I'm not sure I follow. Do you mean we'd keep the existing `Event` class (with its fixed shape) to be used with `EventService` (i.e. when sending event retrieve or list requests), and use a different class for `EventUtility` (i.e. when receiving webhooks)? I don't really see an upside to this.

remi-stripe:
I mean having an `InternalEvent` class that you use to deserialize the JSON on first to get the API version and then decide if you should raise an exception or deserialize a second time on the right class. This avoids "leaking" the expandable logic on `request` just for the sake of supporting 2 API versions. And in the future if there are more changes, we'd break only the internal representation

ob-stripe:
Gotcha. I think a custom deserializer is a cleaner approach than an `InternalEvent` class. I'll update the PR.

ob-stripe:
@remi-stripe I've updated the PR to use a custom deserializer for event objects, so that the `Event` model class only exposes the attributes matching the current API version, and whatever mapping is necessary is done internally. I've also added some tests. ptal

ob-stripe:
@remi-stripe I made another change (which I should have added as a second commit but I didn't because I'm stupid): `EventData` now has an additional `RawObject` property containing an untyped copy of the data in `data.object`. This makes it easier to deal with resources for which the library does not have a concrete type (e.g. `checkout_beta`).

src/Stripe.net/Entities/Events/Event.cs
```diff
@@ -4,36 +4,73 @@ namespace Stripe
...
         public string Account { get; set; }
 
+        /// <summary>
+        /// The Stripe API version used to render <c>data</c>.
+        /// </summary>
+        /// <remarks>
+        /// Note: This property is populated only for events on or after October 31, 2014.
+        /// </remarks>
         [JsonProperty("api_version")]
         public string ApiVersion { get; set; }
```
remi-stripe: realizing this might affect your version mismatch error handling no?

src/Stripe.net/Entities/Events/Event.cs
```diff
@@ -4,36 +4,73 @@ namespace Stripe
...
 
+        /// <summary>
+        /// Has the value <c>true</c> if the object exists in live mode or the value <c>false</c> if
+        /// the object exists in test mode.
+        /// </summary>
         [JsonProperty("livemode")]
         public bool Livemode { get; set; }
 
+        /// <summary>
+        /// Number of webhooks that have yet to be successfully delivered (i.e., to return a 20x
```
remi-stripe: Ugh I don't like this wording. I know it's the official API docs one but we really need to automate the comments generation at least at some point in the future.

src/Stripe.net/Entities/Events/Event.cs
ob-stripe: Hm, good point. I don't think this is going to be a huge problem in practice since `EventUtility.ParseEvent` should only be used when receiving a webhook though.

src/Stripe.net/Entities/Events/Event.cs
ob-stripe: +1000. I started looking into this, it's not a trivial problem but it's definitely doable.

PR #1453, opened 2019-01-03
Fix exception when object key is missing in polymorphic resource deserializer

### LINKED ISSUE #1452: checkout_beta.session_succeeded is throwing a null pointer when parsed with EventUtility.ConstructEvent
Example code:

```c#
    [AllowAnonymous]
    [Route("api/[controller]/[action]")]
    public class StripeController : Controller
    {
        private readonly string SigningSecret;
        private readonly IMediator _mediator;

        public StripeController(IMediator mediator, ApplicationConfiguration configuration)
        {
            SigningSecret = configuration.STRIPE_SIGNING_SECRET;
            _mediator = mediator;
        }

        public async Task<IActionResult> Webhook()
        {
            var @event = await GetEventAsync();
            INotification notification = null;

            switch (@event.Type)
            {
                case "invoice.payment_succeeded":
                    notification = new PaymentSuccessfulNotification(@event);
                    break;
                case "invoice.payment_failed":
                    notification = new PaymentFailedNotification(@event);
                    break;
            }

            if(notification != null)
            {
                await _mediator.Publish(notification);
            }

            return Ok();
        }

        // Consider creating a custom model binder instead.
        private async Task<Event> GetEventAsync()
        {
            using (var reader = new StreamReader(Request.Body, Encoding.UTF8))
            {
                var json = await reader.ReadToEndAsync();
                return EventUtility.ConstructEvent(json, Request.Headers["Stripe-Signature"], SigningSecret);
            }
        }
    }
```

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries @da1rren

Gracefully handle the case where the `object` key is missing in the JSON for a polymorphic resource, by returning `null` instead of raising an exception.

Fixes #1452.

### DIFF
--- a/src/Stripe.net/Infrastructure/JsonConverters/AbstractStripeObjectConverter.cs
+++ b/src/Stripe.net/Infrastructure/JsonConverters/AbstractStripeObjectConverter.cs
@@ -39,7 +39,7 @@ namespace Stripe.Infrastructure
             var incoming = JObject.Load(reader);
             var obj = default(I);
             var objectName = incoming.SelectToken("object")?.ToString();
-            if (this.ObjectsToMapperFuncs.ContainsKey(objectName))
+            if (objectName != null && this.ObjectsToMapperFuncs.ContainsKey(objectName))
             {
                 var mapperFunc = this.ObjectsToMapperFuncs[objectName];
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released as 21.8.1.

PR #1458, opened 2019-01-07
Add support for `wallet_provider` on the Issuing Authorization

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries


remi-stripe:
> Can we add new constants for possible values here? apple_pay, google_pay, or samsung_pay

I don't think we should no. We don't really do that in stripe-dotnet except some historical exceptions or source types.

ob-stripe:
Released as 22.1.0.

PR #1469, opened 2019-01-11
Add transfer_data and application_fee_amount to Invoice and Subscriptions

remi-stripe:
This PRs add `transfer_data` and `application_fee_amount` support to `Invoice` and `Subscription`

r? @ob-stripe 
cc @stripe/api-libraries 

NOT asking for review yet as I want to double check that we don't need `on_behalf_of` for now.

### DIFF
--- a/src/Stripe.net/Entities/Invoices/InvoiceTransferData.cs
+++ b/src/Stripe.net/Entities/Invoices/InvoiceTransferData.cs
@@ -1,0 +1,30 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class InvoiceTransferData : StripeEntity
+    {
+        #region Expandable Destination (Account)
+        [JsonIgnore]
+        public string DestinationId { get; set; }
+
+        [JsonIgnore]
+        public Account Destination { get; set; }
+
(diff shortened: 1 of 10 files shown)


remi-stripe:
unassigning as there's some unknown about `on_behalf_of` right now

src/Stripe.net/Entities/Invoices/InvoiceTransferData.cs
```diff
@@ -0,0 +1,30 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class InvoiceTransferData : StripeEntity
```
remi-stripe: @ob-stripe Should I have made a common `TransferData` class you think? It's quite verbose. But in Java it's done the same way with internal classes and Go has classes per file

remi-stripe: Since we don't want to support `transfer_data[amount]` on all resources, just `Charge` I think it's better to duplicate for now.

remi-stripe:
@ob-stripe We originally wanted to wait until the release of `on_behalf_of` but this will take some time and we would like to unblock this feature for Billing ahead of the 2/19 launch for Charges.

PR #1472, opened 2019-01-17
Add support for billing thresholds

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 

This one is a bit complex as it's across multiple resources and no specific test was added for it so we need to make sure it matches OpenAPI.

cc @clintonb-stripe Could you double check that I added the right set of properties too and I'm not missing any?

Docs: https://stripe.com/docs/billing/subscriptions/metered-billing/thresholds
Internal PR with gate removal: #126799


remi-stripe:
Still deferring to @ob-stripe for full approval

PR #1478, opened 2019-01-18
Better serialization

ob-stripe:
Summary of changes:
- Renamed `ParameterBuilder` to `FormEncoder`
- Greatly simplified the form-encoding logic by replacing a bunch of reflection and unsafe casts with [pattern matching](https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching)
- New utility class `StringUtils` with one new method:
    - `ToSnakeCase()`: converts strings from `CamelCase` to `snake_case`. This is used for `Expand*` properties on services, to convert from the service's property name (e.g. `ExpandFooBar`) to the Stripe property name (e.g. `foo_bar`).
- New extension method on `BaseOptions`:
    - `ToQueryString()`: returns a form-encoded query string containing all of the parameters in the options class, including the ones set in the `ExtraParams` dictionary and the `Expand` list.
- New extension method on `Service`:
    -  `Expansions()`: returns a list of strings with the properties to expand. E.g. if the service has three properties `ExpandFoo`, `ExpandBar` and `ExpandAllTheThings` and the last two are set to `true`, the method will return a list containing `"bar"` and `"all_the_things"`.
- Greatly simplified the existing extension method `ApplyAllParameters()` on `Service`
- Moved the files that were at the root of `Services/` to either `Services/_base` (for base classes meant to be derived from) or `Services/_common` (for options classes that can be used as-is and that are not specific to a single service)
- Updated tests to be consistent with the changes above

Since we have pretty good test coverage for encoding (and I've added a few more tests) I'm fairly confident that I haven't broken anything, but there are a few changes:
- square brackets characters are no longer URL-encoded. This is consistent with most of our other clients and is meant to make the query strings easier to read by humans (since the server doesn't care one way or the other)
- expand parameters are now encoded with integer indices (e.g. `expand[0]=foo&expand[1]=bar`). Previously, they were encoded without indices (`expand[]=foo&expand[]=bar`). This was inconsistent with the way regular lists are encoded. Now all lists are encoded the same way (I've also checked that the server accepts this).

I'm targeting a new `integration-v23` branch despite the fact that there are no breaking changes in this PR, since I have a few more updates planned. The breaking changes should be fairly limited this time around though!


src/StripeTests/Infrastructure/EnumDeserializationTest.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        // TODO: find a better place for this test
+        [Fact]
+        public void EnumDecodeUnknownValue()
+        {
+            var json = "{\"enum\": \"TestTwo\"}";
+            TestOptions obj = JsonConvert.DeserializeObject<TestOptions>(json);
+
+            Assert.NotNull(obj);
+            Assert.Equal(TestOptions.TestEnum.TestTwo, obj.Enum);
```
remi-stripe: Isn't this value supposed to be unknown?

ob-stripe: The test is badly named -- the value is not "unknown", `TestTwo` merely doesn't define an explicit serialization value. I'll rename the test method.

ob-stripe:
cc @stripe/api-libraries

src/Stripe.net/Infrastructure/Extensions/ServiceExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            var expansions = service.Expansions(isListMethod);
+            if (options != null)
+            {
+                expansions.AddRange(options.Expand);
+            }
+
+            return FormEncoder.AppendQueries(
+                baseUrl,
+                options?.ToQueryString(true, false),
```
brandur-stripe: Just because at the callsite here it's not obvious what the parameters `(true, false)` mean, what do you think about explicitly naming them like `ToQueryString(includeExtraParams: true, includeExpandParams: false)`?

ob-stripe:
> but there is a clarity trade-off in that it's immediately clear where `StringUtils.ToSnakeCase` comes from, but not where `str.ToSnakeCase` comes from without going up to check the imports.

Great point, I did get a little carried away with the extension methods :) Looking at Newtonsoft.Json's source code, they do have a static method named exactly [`StringUtils.ToSnakeCase`](https://github.com/JamesNK/Newtonsoft.Json/blob/509643a8952ce731e0207710c429ad6e67dc43db/Src/Newtonsoft.Json/Utilities/StringUtils.cs#L208).

PR #1430, opened 2018-12-13
Rename some properties for consistency

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Renames some properties for consistency with the API:

| Old name | New name |
|---|---|
| `Account.BusinessLogoFileId` | `Account.BusinessLogoId` |
| `LegalEntityVerification.DocumentIdBack` | `LegalEntityVerification.DocumentBackId` |
| `Evidence.CancellationPolicyFileId` | `Evidence.CancellationPolicyId` |
| `Evidence.CancellationPolicyFile` | `Evidence.CancellationPolicy` |
| `Evidence.CustomerCommunicationFileId ` | `Evidence.CustomerCommunicationId ` |
| `Evidence.CustomerCommunicationFile` | `Evidence.CustomerCommunication` |
| `Evidence.CustomerSignatureFileId ` | `Evidence.CustomerSignatureId` |
| `Evidence.CustomerSignatureFile` | `Evidence.CustomerSignature` |
| `Evidence.CustomerSignatureFileId ` | `Evidence.CustomerSignatureId` |
| `Evidence.CustomerSignatureFile` | `Evidence.CustomerSignature` |
| `Evidence.DuplicateChargeDocumentationFileId ` | `Evidence.DuplicateChargeDocumentationId` |
| `Evidence.DuplicateChargeDocumentationFile` | `Evidence.DuplicateChargeDocumentation` |
| `Evidence.ReceiptFileId ` | `Evidence.ReceiptId` |
| `Evidence.ReceiptFile` | `Evidence.Receipt` |
| `Evidence.RefundPolicyFileId ` | `Evidence.RefundPolicyId` |
| `Evidence.RefundPolicyFile` | `Evidence.RefundPolicy` |
| `Evidence.ServiceDocumentationFileId ` | `Evidence.ServiceDocumentationId` |
| `Evidence.ServiceDocumentationFile` | `Evidence.ServiceDocumentation` |
| `Evidence.ShippingDocumentationFileId ` | `Evidence.ShippingDocumentationId` |
| `Evidence.ShippingDocumentationFile` | `Evidence.ShippingDocumentation` |

PR #1481, opened 2019-01-21
Rewrite expandable field handling

ob-stripe:
This PR adds a new `ExpandableField<>` generic class as well as an `ExpandableFieldConverter<>` JSON converter to serialize and deserialize expandable fields. This will replace the current `StringOrObject` implementation.

This change is largely backwards compatible with a small exception: I've removed the setters on the `FooId` and `Foo` properties of each `foo` expandable attribute. This shouldn't be very impactful since most users probably don't use setters at all on our entities classes.


src/Stripe.net/Infrastructure/JsonConverters/ExpandableFieldConverter.cs
```diff
@@ -0,0 +1,112 @@
...
+                case JsonToken.Null:
+                    // Do nothing
+                    break;
+
+                case JsonToken.String:
+                    value.Id = (string)reader.Value;
+                    break;
+
+                case JsonToken.StartObject:
+                    var jObject = JObject.Load(reader);
```
ob-stripe: `JsonReader` is a forward-only stream reader. If the deserialization fails (which can happen with polymorphic types, if the API returns a type that the library isn't aware of), we want to take a second look at the JSON to make a last attempt at extracting the ID. This is why we load the JSON into a `JObject`, then create another reader from the `JObject` instance.

src/Stripe.net/Infrastructure/JsonConverters/ExpandableFieldConverter.cs
```diff
@@ -0,0 +1,112 @@
...
+
+                    using (var subReader = jObject.CreateReader())
+                    {
+                        value.ExpandedObject = (T)base.ReadJson(
+                            subReader, typeof(T), value.ExpandedObject, serializer);
```
ob-stripe: `ExpandableFieldConverter` derives from `StripeObjectConverter`. By calling the parent method here, we ensure that `ExpandableFieldConverter` can be used with interfaces that represent polymorphic types. This is necessary because fields can be both expandable and polymorphic (e.g. `source` on `balance_transaction` objects).

src/Stripe.net/Entities/BalanceTransactions/BalanceTransaction.cs
```diff
@@ -44,24 +44,14 @@ public class BalanceTransaction : StripeEntity, IHasId, IHasObject
 
         #region Expandable Source
         [JsonIgnore]
-        public string SourceId { get; set; }
+        public string SourceId => this.InternalSource.Id;
```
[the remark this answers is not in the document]
ob-stripe: The Go and Java implementations are pretty different. The approach I took here is very similar to stripe-java's, with an `ExpandableField<>` generic and manually defined accessors that map to the ID string and typed expanded object within the generic instance.

We can't use stripe-go's way because we use interfaces for polymorphic fields. When an expandable polymorphic field is unexpanded, we need a concrete type to hold the ID string. That concrete type is the `ExpandableField<>` generic. stripe-go doesn't have that issue because it uses concrete intermediary structs for polymorphic fields.

I agree that the boilerplate for defining each expandable field is annoying but unfortunately I think we'll have to live with it. This PR at least reduces the boilerplate a bit.

ob-stripe:
> What happens if we ever have a property that is expandable but the expanded resource does not have an id? We seem to explicitly expect this to always be true but would it always be? Should we have a fallback when accessing the Id only if set?

Right now, the assumption is that expandable resources always have an ID. (What would an unexpanded resource without an ID even look like?) This assumption is enforced by the compiler, `ExpandableField<>` can only be used with types that implement `IHasId`: https://github.com/stripe/stripe-dotnet/pull/1481/files#diff-ddebcc86a5dc5afa535cf057155bc6f0R10. If it turns out that assumption becomes untrue in the future, we'll have to update the implementation.

> I'm realizing that this means we really need to keep appending Id to all properties that could be expandable in case they ever became expandable in the future to avoid a breaking change. This is something we had considered removing last time and this would have been a bad decision right?

Yep, probably!

> We likely need to decide how to handle comments in the code for the various properties.

Yeah. I made a vague attempt at it in a previous PR (#1430).

> We had some broken integrations last time we changed the serialization/deserialization. Did we make this easier to handle?

This should (🤞) have no effect on external serialization/deserialization. The expandable properties are explicitly annotated with `[JsonConverter(typeof(ExpandableFieldConverter<...>))]` which ensures that the right converter is always used.

src/Stripe.net/Entities/ExpandableField.cs
```diff
@@ -0,0 +1,50 @@
...
+            set => this.expandedObject = value;
+        }
+
+        object IExpandableField.ExpandedObject
+        {
+            get => this.ExpandedObject;
+            set => this.ExpandedObject = (T)value;
+        }
+
+        bool IExpandableField.IsExpanded => this.expandedObject != null;
```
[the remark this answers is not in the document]
ob-stripe: It's implicitly tested by the serialization tests, but I can add an explicit test for it, sure!

src/Stripe.net/Entities/Payouts/Payout.cs
```diff
@@ -85,21 +65,11 @@ internal object InternalDestination
         public string FailureBalanceTransactionId { get; set; }
 
         [JsonIgnore]
-        public BalanceTransaction FailureBalanceTransaction { get; set; }
+        public BalanceTransaction FailureBalanceTransaction => this.InternalFailureBalanceTransaction.ExpandedObject;
```
brandur-stripe: OOC, should we add a `FailureBalanceTransactionId` for this one?

ob-stripe: Oops, nice catch! Looks like my codemod missed this one.

brandur-stripe:
> Technically, those aren't lambdas but [expression-bodied members](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/expression-bodied-members). They do use the same `=>` arrow operator that lambdas do, but they're really only syntactic sugar to define concise getters. (I'm flaunting my C# knowledge but I only learned about expression-bodied members while working on this PR 😂)

Haha, I guess I should've known given the context. Still, I'm glad that I have the right name for this feature now. Thanks for the link!

PR #1483, opened 2019-01-23
Standardize signature of OAuthTokenService.Deauthorize

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Standardizes the signature of `Deauthorize` / `DeauthorizeAsync` methods on `OAuthTokenService` to accept an options class. (This is a breaking change.)

While I was at it, I also added basic support for stubbing requests in tests and wrote tests for `OAuthTokenService`.

### DIFF
--- a/src/StripeTests/Services/OAuth/OAuthTokenServiceTest.cs
+++ b/src/StripeTests/Services/OAuth/OAuthTokenServiceTest.cs
@@ -1,0 +1,92 @@
+namespace StripeTests
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Net;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+
+    using Stripe;
+    using Xunit;
+
+    public class OAuthTokenServiceTest : BaseStripeTest
+    {
+        private const string AccountId = "acct_123";
+        private const string AuthorizationCode = "ac_123";
+        private const string ClientId = "ca_123";
+
+        private readonly OAuthTokenService service;
+        private readonly OAuthTokenCreateOptions createOptions;
+        private readonly OAuthTokenDeauthorizeOptions deauthorizeOptions;
+
+        public OAuthTokenServiceTest(MockHttpClientFixture mockHttpClientFixture)
+            : base(mockHttpClientFixture)
+        {
+            this.service = new OAuthTokenService();
+
+            this.createOptions = new OAuthTokenCreateOptions
+            {
+                Code = AuthorizationCode,
+                GrantType = "authorization_code",
+            };
+
+            this.deauthorizeOptions = new OAuthTokenDeauthorizeOptions
+            {
+                ClientId = ClientId,
+                StripeUserId = AccountId,
+            };
+        }
+
+        [Fact]
+        public void Create()
+        {
+            // stripe-mock doesn't support OAuth endpoints, so stub the request
+            var json = GetResourceAsString("oauth_fixtures.token_response.json");
+            this.StubRequest(HttpMethod.Post, "/oauth/token", HttpStatusCode.OK, json);
+
+            var oauthToken = this.service.Create(this.createOptions);
+            this.AssertRequest(HttpMethod.Post, "/oauth/token");
+            Assert.NotNull(oauthToken);
+            Assert.Equal("sk_test_access_token", oauthToken.AccessToken);
+        }
+
+        [Fact]
+        public async Task CreateAsync()
+        {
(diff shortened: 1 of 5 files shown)


src/StripeTests/Services/OAuth/OAuthTokenServiceTest.cs
```diff
@@ -0,0 +1,89 @@
...
+    using System.Net;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+
+    using Stripe;
+    using Xunit;
+
+    public class OAuthTokenServiceTest : BaseStripeTest
+    {
+        private const string OrderId = "or_123";
```
remi-stripe: wrong constant

src/StripeTests/Services/OAuth/OAuthTokenServiceTest.cs
```diff
@@ -0,0 +1,89 @@
...
+        private readonly OAuthTokenDeauthorizeOptions deauthorizeOptions;
+
+        public OAuthTokenServiceTest(MockHttpClientFixture mockHttpClientFixture)
+            : base(mockHttpClientFixture)
+        {
+            this.service = new OAuthTokenService();
+
+            this.createOptions = new OAuthTokenCreateOptions
+            {
+                Code = "this_is_an_authorization_code",
```
remi-stripe: might be cleaner to use `ac_123` here

src/StripeTests/Services/OAuth/OAuthTokenServiceTest.cs
```diff
@@ -0,0 +1,89 @@
...
+                GrantType = "authorization_code",
+            };
+
+            this.deauthorizeOptions = new OAuthTokenDeauthorizeOptions
+            {
+                StripeUserId = "acct_test_deauth",
```
remi-stripe: This should have the `client_id=ca_123` per https://stripe.com/docs/connect/standard-accounts#revoked-access

src/StripeTests/Services/OAuth/OAuthTokenServiceTest.cs
```diff
@@ -0,0 +1,89 @@
...
+            };
+        }
+
+        [Fact]
+        public void Create()
+        {
+            // stripe=mock doesn't support OAuth endpoints, so stub the request
```
remi-stripe: Not the right place but: should we add support for it?

src/StripeTests/Services/OAuth/OAuthTokenServiceTest.cs
```diff
@@ -0,0 +1,89 @@
...
+            var json = GetResourceAsString("oauth_fixtures.token_response.json");
+            this.StubRequest(HttpMethod.Post, "/oauth/token", HttpStatusCode.OK, json);
+
+            var oauthToken = this.service.Create(this.createOptions);
+            this.AssertRequest(HttpMethod.Post, "/oauth/token");
```
remi-stripe: which parts double check that we use the right URL path?

src/StripeTests/Services/OAuth/OAuthTokenServiceTest.cs
remi-stripe: also `stripe-mock` not `stripe=mock`

ob-stripe: Probably? It's arguably a part of the API. I don't see it happening soon though, the backend implementation is completely independent from the rest of the API and is not ready at all to be encoded into the OpenAPI spec.

src/StripeTests/Services/OAuth/OAuthTokenServiceTest.cs
ob-stripe: `AssertRequest` checks the HTTP method and path, but nothing else (i.e. the request parameters). I'll try to add support for it in the future.

(Also, `StubRequest` implicitly checks the method and path as well -- if the request weren't stubbed with the correct method and path, the call to `Create` would send a real request to the API and fail.)

ob-stripe:
ptal @remi-stripe 

Some things I'll fix in different PRs:
- Add a new `AuthorizeUrl` utility method for forming the `https://connect.stripe.com/oauth/authorize?...` URL
- Add support for Express accounts (we need to do this in all of our libs -- right now I think only stripe-ruby has it)
- When calling `Create`, auto-fill `ClientSecret` with the secret API key if not already set

PR #1484, opened 2019-01-23
New `FromJson` method

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Adds a new `FromJson()` utility method to deserialize JSON strings into Stripe objects.

The method can be used in two ways:
- `var charge = StripeEntity.FromJson<Charge>(json);`
- `var charge = Charge.FromJson(json);`

(The latter is made possible by using the [curiously recurring template pattern](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern).)


ob-stripe:
> A bit sad we have to explicitly pass the class name to StripeEntity though

I could add a method that instantiates the correct type based on the `object` key in the JSON, but it would have to live on `IHasObject`, not `StripeEntity`, since it would only work for models that _do_ have an `object` key.

ob-stripe:
> Do you think it'd be better? It would prevent deserializing some hashes that are not real API resources right?

Such a method could only deserialize "real" API resources, i.e. resources with an `object` key, yes. Otherwise we simply don't know which type to instantiate.

To be clear, all of these are just convenience helper methods. We won't be relying on them internally, we're just exposing them to users to make it a bit simpler to manually deserialize resources.

bugejakurt:
Can we have an async equivalent of this method?

ob-stripe:
@bugejakurt Unfortunately, I don't think that'd be possible. The library uses Newtonsoft.Json to deserialize JSON, and Newtonsoft.Json's `DeserializeObject` method is synchronous.

PR #1485, opened 2019-01-23
Modernize StripeConfiguration

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Modernize `StripeConfiguration`:
- all configuration values are now represented by properties rather than `GetFoo()` / `SetFoo()` methods
- default values are exposed as read-only properties
- all values have XML doc comments
- `HttpTimeSpan` is renamed to `HttpTimeout`, becomes non-nullable, and defaults to 80 seconds (like our other libraries)
- setting the API key via [`ConfigurationManager.AppSettings`](https://docs.microsoft.com/en-us/dotnet/api/system.configuration.configurationmanager.appsettings?view=netframework-4.7.2) now works when targeting .NET Standard 2.0
- moved `/v1` prefix out of the base URLs and into each service's path
- got rid of the `Urls` internal class

No breaking changes: every existing method maps to the new property as appropriate. Existing methods are annotated with `Obsolete` which will generate warnings when used, with a message indicating what to use instead.


src/Stripe.net/Stripe.net.csproj
```diff
@@ -50,10 +50,6 @@
     <PackageReference Include="System.Collections.Immutable" Version="1.5.0" />
   </ItemGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net45' ">
-    <DefineConstants>$(DefineConstants);NET45</DefineConstants>
```
ob-stripe: Turns out this wasn't needed. Preprocessor constants are automatically defined for the target framework: https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-target-frameworks.

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
```diff
@@ -7,27 +7,76 @@ namespace Stripe
...
+
+        /// <summary>Gets or sets the base URL for Stripe's API.</summary>
+        public static string ApiBase { get; set; } = DefaultApiBase;
+
+#if NET45 || NETSTANDARD2_0
+        /// <summary>Gets or sets the API key.</summary>
+        /// <remarks>
+        /// You can also set the API key using the <c>StripeApiKey</c> key in
+        /// <see cref="System.Configuration.ConfigurationManager.AppSettings"/>.
+        /// </remarks>
```
remi-stripe: I feel weird about preprocessor commands + comments. Isn't it better to document it globally?

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
```diff
@@ -36,98 +85,95 @@ public static string StripeApiVersion
...
-            set { serializerSettings = value; }
-        }
-
-        public static string StripeNetVersion { get; }
+        public static JsonSerializerSettings SerializerSettings { get; set; } = DefaultSerializerSettings();
 
         /// <summary>
-        /// This option allows you to provide your own HttpMessageHandler. Useful with Android/iPhone.
+        /// Gets or sets the API version used for requests sent to Stripe's API. It is highly
+        /// recommended you do not change this value.
```
remi-stripe: I don't understand why we support this in the first place?

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
```diff
@@ -36,98 +85,95 @@ public static string StripeApiVersion
...
+        public static string StripeNetVersion { get; }
 
-        public static void SetApiKey(string newApiKey)
+        /// <summary>
+        /// Gets or sets the timespan to wait before the request times out.
+        /// This property is deprecated and will be removed in a future version, please use the
+        /// <see cref="HttpTimeout"/> property instead.
+        /// </summary>
+        // TODO: remove this property in a future major version
+        [Obsolete("Use StripeConfiguration.HttpTimeout instead.")]
```
remi-stripe: We're already merging in a major version. Why not just do it?

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
```diff
@@ -36,98 +85,95 @@ public static string StripeApiVersion
...
-            return apiBase;
+                Converters = new List<JsonConverter>
+                {
+                    new StripeObjectConverter(),
+                },
+                DateParseHandling = DateParseHandling.None,
+            };
         }
 
+        // TODO: remove everything below this in a future major version
```
remi-stripe: Same as above, you're already planning version 23, why not do it?

src/Stripe.net/Infrastructure/Requestor.cs
```diff
@@ -120,7 +117,7 @@ private static async Task<StripeResponse> ExecuteRequestAsync(HttpRequestMessage
 
         internal static HttpRequestMessage GetRequestMessage(string url, HttpMethod method, RequestOptions requestOptions)
         {
-            requestOptions.ApiKey = requestOptions.ApiKey ?? StripeConfiguration.GetApiKey();
+            requestOptions.ApiKey = requestOptions.ApiKey ?? StripeConfiguration.ApiKey;
```
remi-stripe: Okay minor but need to flag: there's no thread safety and such, which I think is okay, but removing a getter prevents from ever adding this right? Though I _guess_ it still has a getter/setter associated with it?

src/Stripe.net/Services/Account/AccountService.cs
```diff
@@ -58,12 +58,12 @@ public virtual Task<Account> GetAsync(string accountId, RequestOptions requestOp
...
         public virtual Account GetSelf(RequestOptions requestOptions = null)
         {
-            return this.GetRequest<Account>($"{Urls.BaseUrl}/account", null, requestOptions, false);
+            return this.GetRequest<Account>($"{StripeConfiguration.ApiBase}/v1/account", null, requestOptions, false);
         }
 
         public virtual Task<Account> GetSelfAsync(RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return this.GetRequestAsync<Account>($"{Urls.BaseUrl}/account", null, requestOptions, false, cancellationToken);
+            return this.GetRequestAsync<Account>($"{StripeConfiguration.ApiBase}/v1/account", null, requestOptions, false, cancellationToken);
```
remi-stripe: ooc can't we do this in `GetRequestAsync` instead?

readme.md
```diff
@@ -35,7 +35,7 @@ You can configure the Stripe.net package to use your secret API key in one of tw
 a) In your application initialization, set your API key (only once once during startup):
 
 ```csharp
-StripeConfiguration.SetApiKey("[your api key here]");
+StripeConfiguration.ApiKey = "[your api key here]";
```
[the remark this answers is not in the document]
ob-stripe: My intention was to avoid a breaking change by keeping the old methods around, but marking them as `Obsolete` to encourage users to upgrade to the new way of setting these values, and actually remove the methods in a future major version. This would make the upgrade a bit less painful for users.

I don't feel too strongly about it, if you'd rather we drop the old methods immediately that's fine by me.

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
```diff
@@ -7,27 +7,76 @@ namespace Stripe
...
+        public static string DefaultConnectBase => "https://connect.stripe.com";
+
+        /// <summary>Default base URL for Stripe's Files API.</summary>
+        public static string DefaultFilesBase => "https://files.stripe.com";
+
+        /// <summary>Default timespan before the request times out.</summary>
+        public static TimeSpan DefaultHttpTimeout => new TimeSpan(80 * TimeSpan.TicksPerSecond);
+
+        /// <summary>Default API version used by the library.</summary>
+        public static string DefaultStripeApiVersion => "2018-11-08";
```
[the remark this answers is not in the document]
ob-stripe: Good point, let's remove the setter and rename to `ApiVersion`.

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
ob-stripe: The issue is that we can't link to `System.Configuration.ConfigurationManager.AppSettings` with .NET Standard 1.2 because the class doesn't exist. I could just remove the link or the comment entirely, but it's kind of a shame.

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
ob-stripe: ¯\\\_(ツ)\_/¯ I'm not sure if it was intended or an oversight, but let's remove it.

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
ob-stripe: See above.

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
ob-stripe: See above.

src/Stripe.net/Infrastructure/Requestor.cs
ob-stripe: I'm not sure I follow. The only thing that changed with regards to the global API key is that `StripeConfiguration.GetApiKey()` is now `StripeConfiguration.ApiKey` and `StripeConfiguration.SetApiKey(newKey)` is now `Stripe.Configuration.ApiKey = newKey`.

Changing the global API was never thread-safe. Users that want to use different API keys in a multi-threaded environment must set the API key per request via `RequestOptions`.

src/Stripe.net/Services/Account/AccountService.cs
ob-stripe: Yup! I plan on doing exactly that in another PR.

ob-stripe:
ptal @remi-stripe

readme.md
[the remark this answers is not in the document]
remi-stripe: Okay I think that's fine and a nicer experience for sure

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
remi-stripe: Ah yeah I see what you mean. Ugh. Sounds good then!

PR #1487, opened 2019-01-28
Replace `Requestor` with `StripeClient` and `HttpClient`

ob-stripe:
Alright, this should be the last of the big changes to the internal requestor logic.

The `Stripe.Infrastructure.Requestor` static class is replaced by several new concepts:
- `IStripeClient` is the interface for a Stripe client class. Its contract defines a single method `MakeRequestAsync<T>`, which accepts a request's arguments (HTTP method, path, options class containing the request's parameters and request options class containing the request's special modifiers) and returns an instance of `T` (where `T` is a type that implements `IStripeEntity`)
- `StripeClient` is the concrete class that implements `IStripeClient`. It accepts an optional `HttpClient` in its constructor. `StripeClient` delegates HTTP communications to the `HttpClient` instance and only takes care of decoding the response from Stripe's API
- `IHttpClient` is the interface for HTTP clients. Its contract defines a single method `MakeRequestAsync` which accepts a `Request` and returns a `Response`
- `SystemNetHttpClient` is the default concrete implementation that implements `IHttpClient`. It accepts an optional `System.Net.HttpClient` instance, or initializes one with default parameters.

This might seem over-engineered but it offers a clear separation of concepts and makes the library easier to test and mock, both by ourselves and by users. (This design is very similar to Twilio's C# library.)

Also, I've renamed the new `Request` class I added in #1508 to `StripeRequest` to make it consistent with `StripeResponse`. I've also extended `StripeResponse` to contain ~all information about the response and added some helpers to access custom Stripe headers.

Breaking changes:
`HttpMessageHandler` and `Timeout` were removed from `StripeConfiguration` and replaced by `StripeClient`. Users who want to pass a custom message handler can still do so by passing a custom client. This is a bit more complex, but it's a lot more powerful:
```c#
StripeConfiguration.StripeClient = new StripeClient(new HttpClient(new System.Net.HttpClient(messageHandler)));
```


ob-stripe:

Whew! I think this is finally ready for review.

src/Stripe.net/Infrastructure/Public/HttpClient.cs
```diff
@@ -0,0 +1,32 @@
...
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// Abstract base class for HTTP clients used to make requests to Stripe's API.
+    /// </summary>
+    public abstract class HttpClient
+    {
+        /// <summary>Gets or sets the last request made by this client.</summary>
+        /// <value>The last request made by this client.</value>
+        public StripeRequest LastRequest { get; protected set; }
```
remi-stripe: How does the concept of `LastRequest` work with async requests where multiple requests are ongoing?

I'm asking because I know we want to add telemetry to .NET too but this one is fully tailored for async and sync requests so not sure how that will fit here.

src/Stripe.net/Infrastructure/Public/StripeResponse.cs
```diff
@@ -1,13 +1,90 @@
...
+
+        /// <summary>
+        /// Gets the body of the response.
+        /// This method is deprecated and will be removed in a future version, please use the
+        /// <see cref="Content"/> property getter instead.
+        /// </summary>
+        /// <value>The body of the response.</value>
+        // TODO: remove this in a future a major version
+        [Obsolete("Use Content instead")]
+        public string ResponseJson => this.Content;
```
remi-stripe: Don't we use this for EphemeralKeys today? https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Entities/EphemeralKeys/EphemeralKey.cs#L25 
We likely need to fix this too

src/Stripe.net/Infrastructure/Public/HttpClient.cs
ob-stripe: `LastRequest` contains the request that was sent last, and `LastResponse` contains the response that was received last. There is no guarantee that the two match, i.e. it's very possible that `LastResponse` contains the response to a different request than the one in `LastRequest`.

For telemetry, we will do something similar to stripe-java and use a [`ConcurrentQueue`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentqueue-1?view=netframework-4.7.2) to enqueue metrics each time a request completes, and dequeue metrics each time a request is sent.

src/Stripe.net/Infrastructure/Public/StripeResponse.cs
ob-stripe: Yep, I did fix `EphemeralKey` :) Our StyleCop rules prevent us from calling obsolete methods in the library itself.

src/StripeTests/Infrastructure/Public/StripeClientTest.cs
```diff
@@ -0,0 +1,112 @@
...
+            var charge = await this.stripeClient.RequestAsync<Charge>(
+                HttpMethod.Post,
+                "/v1/charges",
+                this.options,
+                this.requestOptions);
+
+            Assert.NotNull(charge);
+            Assert.Equal("ch_123", charge.Id);
+            Assert.Equal(response, charge.StripeResponse);
+        }
```
brandur-stripe: Just wanted to echo again that I love how short + succinct these sorts of tests come out! Very readable.

src/Stripe.net/Infrastructure/Public/HttpClient.cs
brandur-stripe: @ob-stripe Don't you think that by exposing `LastResponse` / `LastRequest` inline with an asynchronous method that we're exposing a bit of a misleading API? i.e., You can access these properties, but you won't be able to use them reliably for most things.

I might be missing something here, but couldn't we still get the telemetry we need by not exposing them and doing some management of requests and responses down in `StripeClient. MakeRequestAsync`?

ob-stripe: TBH I cargo-culted `LastRequest` and `LastResponse` from Twilio's SDK -- we don't need them at all for telemetry. I was thinking they could be useful in tests or more generally in contexts where you know what requests are being sent (typically because only one request is sent).

If you think the API could be confusing, I'm fine with making those properties `internal` or even removing them entirely.

brandur-stripe: Ah, I see. My main objection was just that the majority of things people might try to use the properties for wouldn't actually be safe, even if they might have initially seemed safe. I see your point though that it could be useful to have access to this kind of thing during testing (although I think using any form of parallel testing could be potential for trouble).

I'd probably lean toward removal or marking `internal` to avoid the possibility of misuse, but I don't feel super strongly about it. If we do keep them public though, I think we should comprehensively document them to indicate that they're not safe to use in most cases.

src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// <param name="request">The parameters of the request to send.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
+        /// <returns>The task object representing the asynchronous operation.</returns>
+        public override async Task<StripeResponse> MakeRequestAsync(
+            StripeRequest request,
+            CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var httpRequest = BuildRequestMessage(request);
+
+            this.LastRequest = request;
```
brandur-stripe: While we're on the subject: should you synchronize access when setting these properties?

Based on my very [surface-level reading on the task model](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/task-based-asynchronous-programming), it looks like tasks may or may not run on different threads. Is the idea here that there could be a conflict on set, but it doesn't matter?

src/Stripe.net/Infrastructure/Public/HttpClient.cs
ob-stripe: Ack. I've removed the properties and turned `HttpClient` from an abstract class to an interface (and renamed it to `IHttpClient`).

src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
ob-stripe: I've removed the properties so it doesn't really matter anymore, but I think it wouldn't have mattered, but only because nothing relied on these properties. With multiple async requests and/or concurrent requests from different threads, you'd have no guarantee about the values.

In retrospect, removing them was probably the right call :) Thanks Brandur!

ob-stripe:
I've removed the unsafe properties and changed `HttpClient` from an abstract class to an interface (and renamed it to `IHttpClient`).

ptal @brandur-stripe

brandur-stripe:
> I've removed the unsafe properties and changed `HttpClient` from an abstract class to an interface (and renamed it to `IHttpClient`).

👍 

LGTM.

PR #1488, opened 2019-01-28
Simplify Service request methods

ob-stripe:
cc @stripe/api-libraries 

This PR is in preparation of the upcoming refactor in #1487.

It removes the `GetRequest<T> / PostRequest<T> / DeleteRequest<T>` methods and their `Async` counterparts in `Service` and replaces them with just `Request<T> / RequestAsync<T>` methods that take the HTTP verb as their first parameter.

The two methods also have non-typed counterparts `Request / RequestAsync` that use the service's type (since in most cases, when you're sending a request from a service, you're expecting the same resource as the service's).

It also gets rid of the super annoying `isListMethod` flag by using a little introspection: if the return type is a `StripeList<>`, then it's a list request, otherwise it's not.

### DIFF
--- a/src/Stripe.net/Infrastructure/Requestor.cs
+++ b/src/Stripe.net/Infrastructure/Requestor.cs
@@ -100,7 +100,7 @@ namespace Stripe.Infrastructure
             return ExecuteRequestAsync(wr, cancellationToken);
         }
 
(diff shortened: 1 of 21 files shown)


brandur-stripe:
Love it.

LGTM.

ob-stripe:
Thanks Brandur!

PR #1490, opened 2019-01-28
Move base URLs out of resource services where possible

ob-stripe:
cc @stripe/api-libraries 

Simply base URL handling by centralizing it into `Service` base class.

Individual services can override the base URL in one of two ways:
- by overriding the `BaseUrl` property for the service itself, which will set the base URL for all methods in the service (e.g. `OAuthTokenService`)
- by setting the `BaseUrl` property in `RequestOptions`, which will set the base URL for the specific request (e.g. `FileService.Create / CreateAsync`)

### DIFF
--- a/src/Stripe.net/Services/Files/FileService.cs
+++ b/src/Stripe.net/Services/Files/FileService.cs
@@ -24,22 +24,26 @@ namespace Stripe
 
         public virtual File Create(FileCreateOptions options, RequestOptions requestOptions = null)
         {
+            requestOptions = this.SetupRequestOptions(requestOptions);
+            requestOptions.BaseUrl = StripeConfiguration.FilesBase;
             return Mapper<File>.MapFromJson(
                 Requestor.PostFile(
-                    this.ClassUrl(StripeConfiguration.FilesBase),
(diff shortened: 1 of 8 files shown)


src/Stripe.net/Services/Files/FileService.cs
```diff
@@ -24,22 +24,26 @@ public FileService(string apiKey)
 
         public virtual File Create(FileCreateOptions options, RequestOptions requestOptions = null)
         {
+            requestOptions = this.SetupRequestOptions(requestOptions);
+            requestOptions.BaseUrl = StripeConfiguration.FilesBase;
             return Mapper<File>.MapFromJson(
                 Requestor.PostFile(
-                    this.ClassUrl(StripeConfiguration.FilesBase),
+                    StripeConfiguration.FilesBase + this.ClassUrl(),
```
ob-stripe: Technically, setting `requestOptions.BaseUrl` is useless here because the method doesn't call `Service.Request`. That's only temporary because `Service.Request` doesn't support multipart encoding yet, but I'll fix that later. In the meantime, this ensure that `requestOptions.BaseUrl` is consistent with the URL that is passed to `Requestor.PostFile`.

PR #1491, opened 2019-01-28
Add support for Subscription Schedule

remi-stripe:
Add support for Subscription Schedule APIs

* Add Subscription Schedule
* Add Subscription Schedule Revisions
* Add `invoice_settings` on the Customer

cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/SubscriptionsScheduleRevisions/SubscriptionScheduleRevision.cs
+++ b/src/Stripe.net/Entities/SubscriptionsScheduleRevisions/SubscriptionScheduleRevision.cs
@@ -1,0 +1,74 @@
+namespace Stripe
+{
+    using System;
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class SubscriptionScheduleRevision : StripeEntity, IHasId, IHasObject
+    {
+        /// <summary>
+        /// Unique identifier for the object.
+        /// </summary>
+        [JsonProperty("id")]
+        public string Id { get; set; }
+
+        /// <summary>
+        /// String representing the object’s type. Objects of the same type share the same value.
+        /// </summary>
+        [JsonProperty("object")]
+        public string Object { get; set; }
+
+        /// <summary>
+        /// Time at which the object was created. Measured in seconds since the Unix epoch.
+        /// </summary>
+        [JsonProperty("created")]
+        [JsonConverter(typeof(DateTimeConverter))]
+        public DateTime? Created { get; set; }
+
+        /// <summary>
+        /// The schedule's default invoice settings.
+        /// </summary>
+        [JsonProperty("invoice_settings")]
+        public SubscriptionScheduleInvoiceSettings InvoiceSettings { get; set; }
+
+        /// <summary>
+        /// Has the value <code>true</code> if the object exists in live mode or the value
+        /// <code>false</code> if the object exists in test mode.
+        /// </summary>
+        [JsonProperty("livemode")]
+        public bool Livemode { get; set; }
+
+        /// <summary>
+        /// Configuration for the subscription schedule’s phases.
+        /// </summary>
+        [JsonProperty("phases")]
+        public List<SubscriptionSchedulePhase> Phases { get; set; }
+
+        /// <summary>
+        /// ID of the previous subscription schedule revision.
+        /// </summary>
+        [JsonProperty("previous_revision")]
+        public string PreviousRevisionId { get; set; }
+
+        /// <summary>
+        /// Interval and duration at which the subscription schedule renews for when it ends if
+        /// <code>renewal_behavior</code> is <code>renew</code>.
+        /// </summary>
+        [JsonProperty("renewal_behavior")]
+        public string RenewalBehavior { get; set; }
+
+        /// <summary>
+        /// Interval and duration at which the subscription schedule renews for when it ends if
+        /// <code>renewal_behavior</code> is <code>renew</code>.
+        /// </summary>
+        [JsonProperty("renewal_interval")]
+        public SubscriptionScheduleRenewalInterval RenewalInterval { get; set; }
+
+        /// <summary>
+        /// ID of the subscription schedule the revision points to.
+        /// </summary>
+        [JsonProperty("schedule")]
+        public string ScheduleId { get; set; }
+    }
+}
--- a/src/Stripe.net/Entities/SubscriptionsSchedules/SubscriptionSchedule.cs
+++ b/src/Stripe.net/Entities/SubscriptionsSchedules/SubscriptionSchedule.cs
@@ -1,0 +1,193 @@
+namespace Stripe
+{
+    using System;
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class SubscriptionSchedule : StripeEntity, IHasId, IHasMetadata, IHasObject
+    {
+        /// <summary>
+        /// Unique identifier for the object.
+        /// </summary>
+        [JsonProperty("id")]
+        public string Id { get; set; }
+
+        /// <summary>
+        /// String representing the object’s type. Objects of the same type share the same value.
+        /// </summary>
+        [JsonProperty("object")]
+        public string Object { get; set; }
+
+        /// <summary>
+        /// One of <see cref="Billing" />. When charging automatically, Stripe will attempt to pay
+        /// this subscription at the end of the cycle using the default source attached to the
+        /// customer. When sending an invoice, Stripe will email your customer an invoice with
+        /// payment instructions.
+        /// </summary>
+        [JsonProperty("billing")]
+        public Billing? Billing { get; set; }
+
+        /// <summary>
+        /// Define thresholds at which an invoice will be sent, and the subscription advanced to a
+        /// new billing period
+        /// </summary>
+        [JsonProperty("billing_thresholds")]
+        public SubscriptionBillingThresholds BillingThresholds { get; set; }
+
+        /// <summary>
+        /// Time at which the subscription schedule was canceled. Measured in seconds since the
+        /// Unix epoch.
+        /// </summary>
+        [JsonProperty("canceled_at")]
+        [JsonConverter(typeof(DateTimeConverter))]
+        public DateTime? CanceledAt { get; set; }
+
+        /// <summary>
+        /// Time at which the subscription schedule was completed. Measured in seconds since the
+        /// Unix epoch.
+        /// </summary>
+        [JsonProperty("completed_at")]
+        [JsonConverter(typeof(DateTimeConverter))]
+        public DateTime? CompletedAt { get; set; }
+
+        /// <summary>
+        /// Time at which the object was created. Measured in seconds since the Unix epoch.
+        /// </summary>
+        [JsonProperty("created")]
+        [JsonConverter(typeof(DateTimeConverter))]
+        public DateTime? Created { get; set; }
+
+        /// <summary>
+        /// Object representing the start and end dates for the current phase of the subscription
+        /// schedule, if it is <code>active</code>.
+        /// </summary>
+        [JsonProperty("current_phase")]
+        public SubscriptionScheduleCurrentPhase CurrentPhase { get; set; }
+
+        #region Expandable Customer
+
+        /// <summary>
+        /// ID of the <see cref="Customer"/> associated with the subscription schedule.
+        /// <para>Expandable.</para>
+        /// </summary>
+        [JsonIgnore]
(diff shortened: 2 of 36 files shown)


remi-stripe:
@ob-stripe Okay I think this is ready to review entirely. There is a *lot* of content here as this is a complex resource with multiple layers of nested hashes and there are "shared" properties where naming (in this PR) is not great.

Can I ask you to review thoroughly to ensure everything is fully correct? Please pay attention to all class names and feel free to ask me to rename everything. 

Relatedly, there's a deserialization error because I try to expand `phases.coupon` but I don't get a coupon back` and only get `null` which fails the test and shows either a failure in the fixture/openapi/something else.

src/Stripe.net/Entities/SubscriptionsSchedules/SubscriptionSchedulePhasePlan.cs
```diff
@@ -0,0 +1,45 @@
...
+    using Stripe.Infrastructure;
+
+    public class SubscriptionSchedulePhasePlan : StripeEntity
+    {
+        /// <summary>
+        /// Define thresholds at which an invoice will be sent, and the subscription advanced to a
+        /// new billing period.
+        /// </summary>
+        [JsonProperty("billing_thresholds")]
+        public SubscriptionSchedulePhasePlanBillingThresholds BillingThresholds { get; set; }
```
remi-stripe: Flagging that this is the same hash as the one on Subscription Item documented [here](https://stripe.com/docs/api/subscription_items/object#subscription_item_object-billing_thresholds). I did not want to re-use that class name as it's `SubscriptionItemBillingThresholds` and it's confusing.

@alexander-stripe suggested we unify though and rename both to `UsageBasedBillingThreshold` for example. Renaming would be a breaking change but since have a major revision shipping tomorrow or Friday maybe it's worth it? This also impacts the Options class(es).

Note that there is also `billing_thresholds` at the top level of this object and this one matches the one on Subscription and so I shared `SubscriptionBillingThresholds` because it felt like it made sense.

src/Stripe.net/Entities/SubscriptionsSchedules/SubscriptionSchedule.cs
```diff
@@ -0,0 +1,179 @@
...
+        public DateTime? Created { get; set; }
+
+        /// <summary>
+        /// Object representing the start and end dates for the current phase of the subscription
+        /// schedule, if it is <code>active</code>.
+        /// </summary>
+        [JsonProperty("current_phase")]
+        public SubscriptionScheduleCurrentPhase CurrentPhase { get; set; }
+
+        // TODO: figure out how we document those
```
remi-stripe: Should I bother documenting since this is going away in your next major version?

src/Stripe.net/Services/SubscriptionSchedules/SubscriptionScheduleInvoiceSettingsOptions.cs
```diff
@@ -0,0 +1,15 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class SubscriptionScheduleInvoiceSettingsOptions : StripeEntity
```
remi-stripe: Talking with @alexander-stripe There is a world in which all `invoice_settings` are merged (instead of being 2 separate hashes). This is not yet scheduled/planned though, just some thoughts. This means that when it happens it'd be a major/breaking change anyway.

Still flagging that I am not happy with all those classes names that are attached to the first object ever seeing them...

src/Stripe.net/Services/SubscriptionSchedules/SubscriptionScheduleListOptions.cs
```diff
@@ -0,0 +1,39 @@
+namespace Stripe
+{
+    using System;
+    using Newtonsoft.Json;
+
+    public class SubscriptionScheduleListOptions : ListOptionsWithCreated
+    {
+        // TODO: Figure out how to document those
```
remi-stripe: How do I document those? Do I ignore since you're fixing in the next major revision?

src/Stripe.net/Services/SubscriptionSchedules/SubscriptionScheduleRenewalIntervalOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class SubscriptionScheduleRenewalIntervalOptions : StripeEntity
+    {
+        /// <summary>
+        /// Interval at which to renew the subscription schedule for when it ends. Possible values
+        /// are <code>day</code>, <code>week</code>, <code>month</code>, or <code>year</code>.
+        /// </summary>
+        // TODO: Decide if using PlanInterval
```
remi-stripe: Should I be using `PlanInterval` here? It's the same list, but also feels like a different meaning (and maybe we would like this to be renamed to `BillingInterval`?)

ob-stripe: Yeah, it's annoying, and the OpenAPI spec is no help here either as the enums are not defined as a single schema that is referenced from different parameters, but rather duplicated for each parameter.

I'd leave as is for now, but it's something we should fix upstream.

src/StripeTests/Entities/SubscriptionScheduleRevisions/SubscriptionScheduleRevisionTest.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+        }
+
+        [Fact]
+        public void Deserialize()
+        {
+            string json = this.GetFixture("/v1/subscription_schedules/sub_sched_123/revisions/sub_sched_rev_123");
+            var scheduleRevision = JsonConvert.DeserializeObject<SubscriptionScheduleRevision>(json);
+            Assert.NotNull(scheduleRevision);
+            Assert.IsType<SubscriptionSchedule>(scheduleRevision);
```
ob-stripe:
```suggestion
            Assert.IsType<SubscriptionScheduleRevision>(scheduleRevision);
```

src/Stripe.net/Entities/SubscriptionsScheduleRevisions/SubscriptionScheduleRevision.cs
```diff
@@ -0,0 +1,74 @@
...
+        public bool Livemode { get; set; }
+
+        /// <summary>
+        /// Configuration for the subscription schedule’s phases.
+        /// </summary>
+        [JsonProperty("phases")]
+        public List<SubscriptionSchedulePhase> Phases { get; set; }
+
+        /// <summary>
+        /// TODO: Figure out docs
```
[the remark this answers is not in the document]
remi-stripe: Sure, I did this mostly because the compiler complained otherwise originally (it was expanded in my code). I'm not planning to keep as is, just missed it :)

src/Stripe.net/Entities/SubscriptionsSchedules/SubscriptionSchedule.cs
remi-stripe: @ob-stripe Sure but the entire "region" is going away though. So my todo/question holds: How should I document this? Can you show me an example with this region on where to put the comments so that your rebase is not too painful?

src/Stripe.net/Entities/SubscriptionsSchedules/SubscriptionSchedulePhasePlan.cs
remi-stripe: I have clear fud here because openapi is generated based on client-metadata in pay-server. This is a separate discussion though.
Also whatever is in openapi can change without being considered a breaking change by anyone but a few of us.

I am fine using the `subscription_item_billing_thresholds` if you want, but I think it's for the wrong reasons. We should rename openapi in that case really.

src/Stripe.net/Services/SubscriptionScheduleRevisions/SubscriptionScheduleRevisionService.cs
```diff
@@ -0,0 +1,48 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public class SubscriptionScheduleRevisionService : ServiceNested<SubscriptionScheduleRevision>,
+        INestedListable<SubscriptionScheduleRevision, SubscriptionScheduleRevisionListOptions>,
+        INestedRetrievable<SubscriptionScheduleRevision>
```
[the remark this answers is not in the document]
remi-stripe: Can you clarify? They are each on their own line already no?

src/Stripe.net/Services/SubscriptionSchedules/SubscriptionScheduleInvoiceSettingsOptions.cs
remi-stripe: I think last time we tried there was a stylecop rule against nested class.

src/StripeTests/Entities/SubscriptionScheduleRevisions/SubscriptionScheduleRevisionTest.cs
remi-stripe: fixed

src/Stripe.net/Services/SubscriptionScheduleRevisions/SubscriptionScheduleRevisionService.cs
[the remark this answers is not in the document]
ob-stripe: Yes, sorry, I can't read.

src/Stripe.net/Entities/SubscriptionsSchedules/SubscriptionSchedule.cs
ob-stripe: The region would stay there too, what would change is:
- `public string CustomerId { get; set; }` --> `public string CustomerId => InternalCustomer.Id;`
- `public Customer Customer { get; set; }` --> `public Customer Customer => InternalCustomer.ExpandedObject;`
- `internal object InternalCustomer { ... }` --> `internal ExpandableField<Customer> InternalCustomer;`

Here's an example for documenting the `public` accessors for the ID and object: https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Entities/Disputes/Evidence.cs#L16-L28. Not sure it's the best format, but it's something.

src/Stripe.net/Entities/SubscriptionsSchedules/SubscriptionSchedulePhasePlan.cs
ob-stripe: My point was that both resources use the same schema in OpenAPI, which means that in our backend code `subscription_item_billing_thresholds` is defined once and referenced in both resources. So if it ever changes, it would change for both resources at once and we'd need to reflect that in the bindings.

I agree the name is bad because it specifically mentions `subscription_item`, and should be changed.

src/Stripe.net/Services/SubscriptionSchedules/SubscriptionScheduleInvoiceSettingsOptions.cs
ob-stripe: Ah right, thanks. I'm not against changing that rule, but we'd have to be sure it's the right move. stripe-java does use inner classes, so it's worth experimenting with them at least.

ob-stripe:
> Relatedly, there's a deserialization error because I try to expand `phases.coupon` but I don't get a coupon back`and only get`null` which fails the test and shows either a failure in the fixture/openapi/something else.

Yes, `phases.coupon` is `null` in the fixtures file for both the schedule and schedule revision objects. Mind commenting out that assertion in the test for now?

remi-stripe:
@alexander-stripe Assigning to you for the final review to ensure all the resources are described properly and that I did not miss anything.

The main "meta" question left is whether we want to "share" `BillingThresholds` classes between the one on a subscription item and the one on the `phases[][plans][]`. If you don't feel strongly we will go with what the PR did (separate classes) for now.

alexander-stripe:
I glanced over this, it looks like it has the right shape to me. Thanks for this :-)  I don't feel strongly to share the classes, although I can't see them diverge in the future so it may be worth unifying. Other than that

LGTM

remi-stripe:
After spending a while on the stripe-java version (manual and autogenerated) I decided that I agreed and that `Phase` and `PhaseItem` were better names and that I should re-use the `SubscriptionItemBillingThresholds*` classes so I did just that.

remi-stripe:
@ob-stripe now that stripe-mock has been released, we should be able to merge. The feature is stable (but gated for a bit)

PR #1492, opened 2019-01-29
Add changes to the Account and Person resources

remi-stripe:
This PR contains the changes related to the new API version `2019-02-19`:
* Move stripe-go to API version `2019-02-19`
* Change the shape of `Account`
  * The `legal_entity` property on the Account API resource has been replaced with `individual`, `company`, and `business_type`.
  * The `verification` hash has been replaced with a `requirements` hash.
  * Multiple top-level properties were moved to the `settings` hash.
  * The `keys` property on `Account` has been removed. Platforms should authenticate as their connected accounts with their own key via the `Stripe-Account` [header](https://stripe.com/docs/connect/authentication#authentication-via-the-stripe-account-header).
* The `requested_capabilities` property on `Account` creation is now required for accounts in the US.
* Removed deprecated parameter `save_source_to_customer` on `PaymentIntent`. Use `save_payment_method` instead.
cc @stripe/api-libraries


remi-stripe:
@ob-stripe Could you have a look at the first draft? This won't ship for ~2/3 weeks but I want to get some eyes early on.

src/Stripe.net/Entities/Persons/BirthDay.cs
```diff
@@ -2,6 +2,7 @@ namespace Stripe
 {
     using Newtonsoft.Json;
 
+    // TODO: Decide if we want to rename that class
```
ob-stripe: We probably should. Birthday != date of birth.

src/Stripe.net/Services/Account/AccountSharedOptions.cs
```diff
@@ -43,32 +34,15 @@ public abstract class AccountSharedOptions : BaseOptions
         [JsonProperty("external_account")]
         public AccountBankAccountOptions ExternalBankAccount { get; set; }
 
-        [JsonProperty("legal_entity")]
-        public AccountLegalEntityOptions LegalEntity { get; set; }
-
+        // TODO: Figure out how to handle Individual
```
ob-stripe: What's the issue exactly? Can't we declare it as `public Person Individual { get; set; }`?

src/Stripe.net/Entities/Accounts/Account.cs
```diff
@@ -89,52 +56,29 @@ internal object InternalBusinessLogo
...
-
-        [JsonProperty("payout_statement_descriptor")]
-        public string PayoutStatementDescriptor { get; set; }
-
         [JsonProperty("payouts_enabled")]
         public bool PayoutsEnabled { get; set; }
 
-        [JsonProperty("product_description")]
-        public string ProductDescription { get; set; }
+        // TODO: Figure out if we can use Requirements as there's `current_deadline`
```
ob-stripe: I think we need to rename `Requirements` to `PersonRequirements` and have a separate `AccountRequirements`.

ob-stripe:
I haven't checked that every single change here is inline with the embedded spec, but overall this looks good to me. I've left comments for the TODOs.

src/Stripe.net/Services/Account/AccountSharedOptions.cs
remi-stripe: Well that works on the resource, not on the parameters. We could use `PersonCreateOptions` maybe but one is a `BaseOptions` while here we'd want an `INestedOptions`

src/Stripe.net/Entities/Accounts/Account.cs
remi-stripe: Yep, I was just hoping not to have to do that. It's annoying to have so many of those resources with similar names and one property that changes.

src/Stripe.net/Services/Account/AccountSharedOptions.cs
ob-stripe: I don't think that's a huge problem. The only practical difference between `BaseOptions` and `INestedOptions` is that classes that inherit the former get `AddExpand()` and `AddExtraParam()` methods. It's not a big deal if we use a `BaseOptions` class as a nested options class.

remi-stripe: I had no idea! Fixed.

PR #1498, opened 2019-02-02
Handle OAuth error message in StripeException

### LINKED ISSUE #1497: Better StripeException Message?
Right now when a StripeException is thrown, the Message property is:

`Exception of type 'Stripe.StripeException' was thrown.`

And then there's a StripeError property that has more info. We use Sentry for logging exceptions, and so the only thing we're seeing is the stack trace and the top level message, which is not very useful. (We already know StripeException was thrown, since we're looking at it. :trollface:)

Would it be possible for you to copy the StripeException.StripeError.ErrorDescription and Error into the top level .Message property? 

For the record, this is what I'm seeing now for an example exception in our debugger:

screen shot 2019-02-01 at 11 02 22 am

So in this scenario, perhaps the top level .Message would be:

`Authorization code expired (invalid_grant).`

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

stripe-dotnet uses the same `StripeError` object for both regular and OAuth errors, but depending on which it is, only certain properties will be filled.

Despite this, when building a `StripeException`, the `Message` property is always passed as the exception message. This property is only filled for regular API errors. For OAuth errors, `ErrorDescription` should be used as the exception message instead.

Fixes #1497.

### DIFF
--- a/src/Stripe.net/Infrastructure/Requestor.cs
+++ b/src/Stripe.net/Infrastructure/Requestor.cs
@@ -229,7 +229,11 @@ namespace Stripe.Infrastructure
                 ? Mapper<StripeError>.MapFromJson(responseContent, null, response)
                 : Mapper<StripeError>.MapFromJson(responseContent, "error", response);
 
-            return new StripeException(statusCode, stripeError, stripeError.Message)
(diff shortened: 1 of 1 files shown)


ob-stripe:
Released as 22.8.1.

PR #1499, opened 2019-02-02
Simplify handling of Expand and ExtraParams

ob-stripe:
r? @remi-stripe 

Simplify handling of `Expand` and `ExtraParams` in `BaseOptions`:
- `Expand` is now annotated with `[JsonProperty("expand")]`
- `ExtraParams` is now annotated with `[JsonExtensionData]`

`JsonExtensionData` is a special attribute in Newtonsoft.Json that instructs the JSON serializer to write the values in the collection during serialization and to deserialize values with no matching class member during deserialization.

I added a few lines of code to `FormEncoder` to do something similar in our own serialization process. I also had to change the type of `ExtraParams` from `Dictionary<string, string>` to `Dictionary<string, object>`, but I think that's better anyway, as it lets uses use non-string extra params (e.g. a list or a hash).

Also, `Expand` is now `null` by default instead of an empty list, because otherwise it would be serialized as `expand=` when the list is empty. Using `AddExpand()` will automatically create the list if necessary.

### DIFF
--- a/src/Stripe.net/Infrastructure/Extensions/BaseOptionsExtensions.cs
+++ b/src/Stripe.net/Infrastructure/Extensions/BaseOptionsExtensions.cs
@@ -7,15 +7,10 @@ namespace Stripe.Infrastructure.Extensions
     {
         /// <summary>Serializes the parameters with form encoding.</summary>
         /// <param name="options">The options class containing the parameters.</param>
-        /// <param name="includeExtraParams">Whether to include extra parameters in the query string or not.</param>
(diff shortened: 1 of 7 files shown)


remi-stripe:
This looks great to me. There are some things I don't fully grasp so re-assign to @brandur-stripe to double check but feel free to self-approve and merge!

brandur-stripe:
Seems way cleaner to me!

LGTM.

ob-stripe:
Thanks both!

PR #1504, opened 2019-02-07
More request encoding refactoring

ob-stripe:

More refactoring. There will be a couple more PRs, but I swear I'm getting somewhere :)

- Two new classes:
    - `Stripe.Infrastructure.FormEncoding.FormUrlEncodedContent` is used to represent POST payloads encoded with `application/x-www-form-urlencoded` (i.e. every POST request except file creation requests)
    - `Stripe.Infrastructure.FormEncoding.MultipartFormDataContent` is used to represent POST payloads encoded with `multipart/form-data` (i.e. file creation requests)
- `Stripe.Infrastructure.FormEncoding.FormEncoder` changes:
    - `EncodeOptions()` is renamed to `CreateQueryString()`
    - The type for flattened parameters is changed from `List<KeyValuePair<string, string>>` to `List<KeyValuePair<string, object>>`. The value can now be either a string or a `Stream`.
    - New method: `CreateHttpContent()`, which accepts a `BaseOptions` like `CreateQueryString()` but returns a `HttpContent` to be used as a POST payload instead of a string. It will automatically returns the correct type of payload: if all of the values in the flattened parameters are strings, then it returns a `FormUrlEncodedContent`, otherwise it returns a `MultipartFormDataContent`
- `Stripe.Infrastructure.Requestor` changes:
    - Previously, `Service` would _always_ append the parameters to the URL as a query string, then `Requestor` would remove the query string and turn it into a `HttpContent` for POST requests
    - Now, `Requestor.GetRequestMessage()` receives both the URL (without a query string) and the `BaseOptions` request parameters, and directly turns the parameter into either a query string (for GET and DELETE requests) or a `HttpContent` (for POST requests)

Let me know if anything's unclear!


src/Stripe.net/Infrastructure/Extensions/BaseOptionsExtensions.cs
```diff
@@ -1,16 +1,17 @@
...
     internal static class BaseOptionsExtensions
     {
-        /// <summary>Serializes the parameters with form encoding.</summary>
+        /// <summary>Serializes the parameters as a query string.</summary>
         /// <param name="options">The options class containing the parameters.</param>
         /// <returns>A form-encoded string represensation of the parameters.</returns>
         public static string ToQueryString(this BaseOptions options)
         {
-            return FormEncoder.EncodeOptions(options);
+            return FormEncoder.CreateQueryString(options);
```
remi-stripe: I swear this was the name months ago and we changed it because we did not like it. Likely misremembering lol

src/Stripe.net/Infrastructure/FormEncoding/FormEncoder.cs
```diff
@@ -16,60 +18,53 @@ namespace Stripe.Infrastructure.FormEncoding
...
-        /// <param name="url">The base URL.</param>
-        /// <param name="queries">One or more query strings to be appended to the URL.</param>
-        /// <returns>The full URL with all the query strings.</returns>
-        public static string AppendQueries(string url, params string[] queries)
-        {
-            var fullQuery = JoinQueries(queries);
+            var flatParams = FlattenParamsValue(options, null);
 
-            if (!string.IsNullOrEmpty(fullQuery))
+            if (flatParams.All(kvp => kvp.Value is string))
```
brandur-stripe: Is the idea here that we know that by this point all parameters will have for sure been converted to a string except for those intended for multipart encoding? If so, mind adding a comment to that effect?

src/Stripe.net/Infrastructure/FormEncoding/MimeTypes.cs
```diff
@@ -0,0 +1,45 @@
...
+            { ".csv", "text/csv" },
+            { ".docx", "application/vnd.openxmlformats-officedocument.wordprocessingml.document" },
+            { ".gif", "image/gif" },
+            { ".jpeg", "image/jpeg" },
+            { ".jpg", "image/jpeg" },
+            { ".pdf", "application/pdf" },
+            { ".png", "image/png" },
+            { ".xls", "application/vnd.ms-excel" },
+            { ".xlsx", "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet" },
+        }.ToImmutableDictionary();
```
brandur-stripe: Have never seen `ToImmutableDictionary` before, but I like it! Nice.

src/Stripe.net/Infrastructure/FormEncoding/MimeTypes.cs
```diff
@@ -0,0 +1,45 @@
+namespace Stripe.Infrastructure.FormEncoding
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Collections.Immutable;
+
+    internal static class MimeTypes
+    {
+        private static readonly IDictionary<string, string> MimeTypeMap
```
brandur-stripe: I realize this is basically refactored from another piece of code, but mind writing up some justification for why these particular file extensions are handled and why they're deemed to be a sufficient set for what's likely to be uploaded?

As someone not too familiar with the code, it'd be nice to know when it's appropriate to add a new supported file.

src/Stripe.net/Infrastructure/FormEncoding/FormEncoder.cs
ob-stripe: Yes, that's it. The values in the flattened parameters map are either strings or streams. If all of them are strings, then the content can be encoded as `application/x-www-form-urlencoded`, otherwise it needs to be encoded as `multipart/form-data`. I'll add a comment 👍

src/Stripe.net/Infrastructure/FormEncoding/MimeTypes.cs
ob-stripe: The new dictionary contains all the mimetypes that can be used when uploading a file. We don't need to be very diligent about keeping the list up-to-date: the backend doesn't trust the `Content-Type` and determines the type from the actual file contents anyway. I considered removing this entirely and always using `application/octet-stream` for the `Content-Type` (some of the other clients do this IIRC), but I think this is a bit nicer and potentially helpful when debugging.

I'll add a comment to explain the above :)

ob-stripe:
> I especially love the new tests — very thorough and very succinct.

Thanks! I can't take too much credit here, those tests are heavily based on the tests for the existing `System.Net.Http.XContent` classes.

PR #1508, opened 2019-02-08
Introduce new Request class to represent requests to Stripe's API

ob-stripe:

This one is fairly straightforward. It adds a new `Stripe.Infrastructure.Request` class whose job is to turn the request parameters provided by services:
- `HttpMethod`: HTTP method (...)
- `path`: relative path, e.g. `/v1/charges`
- `options`: `BaseOptions` instance containing the request parameters, or `null`
- `requestOptions`: `RequestOptions` instance containing the request's "modifiers" to e.g override the global API key or set an idempotency key

into .NET's HTTP library classes:
- `HttpMethod`: this one is untouched
- `Uri`: `System.Net.Uri` instance containing the full URL, including the base path, and for non-POST requests, the parameters encoded in the query string
- `AuthorizationHeader`: `System.Net.Http.Headers.AuthenticationHeaderValue` representing the `Authorization` header with our preferred "Bearer" scheme and the API key
- `StripeHeaders`: a dictionary containing Stripe-specific HTTP headers. All requests include `Stripe-Version`, and may include `Stripe-Account` and/or `Idempotency-Key`
- `Content`: for POST requests, a `System.Net.Http.HttpContent` instance representing the request's body. For non-POST requests, `null`.

### DIFF
--- a/src/Stripe.net/Infrastructure/Request.cs
+++ b/src/Stripe.net/Infrastructure/Request.cs
@@ -1,0 +1,119 @@
+namespace Stripe.Infrastructure
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Net.Http;
+    using System.Net.Http.Headers;
+    using System.Text;
+    using Stripe.Infrastructure.Extensions;
+    using Stripe.Infrastructure.FormEncoding;
+
+    /// <summary>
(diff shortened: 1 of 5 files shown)


brandur-stripe:
Nice!

LGTM.

PR #1509, opened 2019-02-10
Fixes to FileService and OAuthTokenService

ob-stripe:
r? @remi-stripe 

In #1504, I inadvertently removed the part of the code that sends file creation requests to `files.stripe.com` rather than `api.stripe.com` 😱  Tests were still passing because in the test suite, we replace both hostnames with `localhost` since stripe-mock handles both.

It should be easier to test this once I'm done refactoring the requestor code.

The changes to `OAuthTokenService` isn't a fix, it's just to make the code a bit cleaner by using standard creation methods.

### DIFF
--- a/src/Stripe.net/Services/Files/FileService.cs
+++ b/src/Stripe.net/Services/Files/FileService.cs
@@ -1,11 +1,8 @@
 namespace Stripe
 {
     using System.Collections.Generic;
-    using System.IO;
-    using System.Net.Http;
     using System.Threading;
     using System.Threading.Tasks;
-    using Stripe.Infrastructure;
 
     public class FileService : Service<File>,
         ICreatable<File, FileCreateOptions>,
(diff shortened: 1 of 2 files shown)


src/Stripe.net/Services/Files/FileService.cs
```diff
@@ -26,11 +23,15 @@ public FileService(string apiKey)
 
         public virtual File Create(FileCreateOptions options, RequestOptions requestOptions = null)
         {
+            requestOptions = this.SetupRequestOptions(requestOptions);
+            requestOptions.BaseUrl = requestOptions.BaseUrl ?? StripeConfiguration.FilesBase;
```
remi-stripe: I don't fully grasp that line and why it's needed. Since we always know we want FilesBase here why change it? Is it just for tests?

ob-stripe: Do you mean why it's
```c#
requestOptions.BaseUrl = requestOptions.BaseUrl ?? StripeConfiguration.FilesBase;
```
instead of just
```c#
requestOptions.BaseUrl = StripeConfiguration.FilesBase;
```
? If so, it's just because I think it's bad practice to silently overwrite an existing value. In practice, this doesn't change anything: `BaseUrl` is `internal`, so it cannot be set by library users, and there is no internal code that calls `FileService.Create / CreateAsync`.

remi-stripe: Yeah that's mostly what I meant. The dichotomy between what you know you want and what the other property could be! But all good!

PR #1511, opened 2019-02-12
Remove extension methods

ob-stripe:

Remember those fancy extension methods I added in #1478? Well, they're gone already.

- `BaseOptionsExtensions.ToQueryString()` was no longer used in the library after all the requestor code refactoring, only in tests
- `ServiceExtensions.Expansions()` was only used in the `Service` class itself, so it didn't make sense for it to be an extension method. It's now a regular class method.

This PR also adds the ability to pass a custom Stripe client to service instances, and uses this for testing the now private `Expansions()` method.


brandur-stripe:
Haha, nice. LGTM!

remi-stripe:
This looks good but deferring to Brandur as usual for those core changes.

PR #1513, opened 2019-02-12
Add support for telemetry

ob-stripe:
r? @brandur-stripe @jameshageman-stripe 

Add support for telemetry. The implementation is fairly similar to stripe-java's: https://github.com/stripe/stripe-java/pull/661.

### DIFF
--- a/src/StripeTests/Functional/TelemetryTest.cs
+++ b/src/StripeTests/Functional/TelemetryTest.cs
@@ -1,0 +1,213 @@
+namespace StripeTests
+{
+    using System;
+    using System.Linq;
+    using System.Net;
+    using System.Net.Http;
+    using System.Net.Http.Headers;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Moq;
+    using Moq.Protected;
+    using Newtonsoft.Json.Linq;
+    using Stripe;
+    using Xunit;
+
+    public class TelemetryTest : BaseStripeTest, IDisposable
+    {
+        public TelemetryTest(MockHttpClientFixture mockHttpClientFixture)
+            : base(mockHttpClientFixture)
+        {
+        }
+
+        public void Dispose()
+        {
+            this.ResetStripeClient();
+            StripeConfiguration.EnableTelemetry = false;
+        }
+
+        [Fact]
+        public void TelemetryDisabled()
+        {
+            this.ResetStripeClient();
+            var fakeServer = FakeServer.ForMockHandler(this.MockHttpClientFixture.MockHandler);
+
+            StripeConfiguration.EnableTelemetry = false;
+            var service = new BalanceService();
+            service.Get();
+            service.Get();
+            service.Get();
+
+            this.MockHttpClientFixture.MockHandler.Protected()
+                .Verify(
+                    "SendAsync",
+                    Times.Exactly(3),
+                    ItExpr.Is<HttpRequestMessage>(m =>
+                        !m.Headers.Contains("X-Stripe-Client-Telemetry")),
+                    ItExpr.IsAny<CancellationToken>());
+        }
+
+        [Fact]
+        public void TelemetryEnabled()
+        {
+            this.ResetStripeClient();
+            var fakeServer = FakeServer.ForMockHandler(this.MockHttpClientFixture.MockHandler);
+            fakeServer.DelayMilliseconds = 10;
+
+            StripeConfiguration.EnableTelemetry = true;
+            var service = new BalanceService();
+            service.Get();
+            fakeServer.DelayMilliseconds = 20;
+            service.Get();
+            service.Get();
+
+            this.MockHttpClientFixture.MockHandler.Protected()
+                .Verify(
+                    "SendAsync",
+                    Times.Once(),
+                    ItExpr.Is<HttpRequestMessage>(m =>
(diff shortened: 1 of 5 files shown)


src/StripeTests/Functional/TelemetryTest.cs
```diff
@@ -0,0 +1,216 @@
...
+            Balance b1 = service.Get();
+            Balance b2 = service.Get();
+            Balance b3 = service.Get();
+
+            this.MockHttpClientFixture.MockHandler.Protected()
+                .Verify(
+                    "SendAsync",
+                    Times.Exactly(3),
+                    ItExpr.Is<HttpRequestMessage>(m =>
+                        !m.Headers.Contains("X-Stripe-Client-Telemetry")),
```
jameshageman-stripe: These assertions are awesome!

src/StripeTests/Functional/TelemetryTest.cs
```diff
@@ -0,0 +1,216 @@
...
+                    ItExpr.IsAny<CancellationToken>());
+
+            this.MockHttpClientFixture.MockHandler.Protected()
+                .Verify(
+                    "SendAsync",
+                    Times.Exactly(2),
+                    ItExpr.Is<HttpRequestMessage>(m =>
+                        TelemetryHeaderMatcher(
+                            m.Headers,
+                            (s) => s == "req_1" || s == "req_2",
```
jameshageman-stripe: It would be nice if could add all of the `request_id`s that the MockHandler sees to a set, and then asserting that the size of the set is 2 here. That way, instead of asserting that the sent metrics were "req_1" or "req_2", we assert that _both_ "req_1" and "req_2" were sent in some order.

src/StripeTests/Functional/TelemetryTest.cs
```diff
@@ -0,0 +1,216 @@
...
+            if (!headers.Contains("X-Stripe-Client-Telemetry"))
+            {
+                return false;
+            }
+
+            var payload = headers.GetValues("X-Stripe-Client-Telemetry").First();
+
+            var deserialized = JToken.Parse(payload);
+            var requestId = (string)deserialized["last_request_metrics"]["request_id"];
+            var duration = (long)deserialized["last_request_metrics"]["request_duration_ms"];
```
jameshageman-stripe: I like that you're testing against the raw json instead of the `ClientTelemetryPayload`, as this ensures that we send the exact json that Stripe expects.

src/StripeTests/Functional/TelemetryTest.cs
ob-stripe: Ack. I can't think of an easy way to do that, but given that there are only 2 requests, I'll just duplicate the `Verify()` to check that there was one invocation with "req_1" and another with "req_2". Not the most scalable solution, but I think it's sufficient for the purpose of this test.

src/StripeTests/Functional/TelemetryTest.cs
ob-stripe: I did it this way because `ClientTelemetryPayload` is `private` and I didn't want to expose it just for tests, but you're right that it actually makes the test better! :)

ob-stripe:
ptal @jameshageman-stripe

jameshageman-stripe:
The pr tests indicate that both of the two requests sent in the concurrent test that have telemetry are both reporting `req_1`:

```
HttpClientHandler.SendAsync(Method: GET, RequestUri: 'http://localhost:12111/v1/balance', Version: 1.1, Content: <null>, Headers:
{
  User-Agent: Stripe/v1
  User-Agent: .NetBindings/22.9.0
  Authorization: Bearer sk_test_123
  X-Stripe-Client-User-Agent: {"bindings_version":"22.9.0","lang":".net","publisher":"stripe","lang_version":".NET Core 4.6.25211.01","os_version":"Microsoft Windows 10.0.14393 "}
  Stripe-Version: 2018-11-08
  X-Stripe-Client-Telemetry: {"last_request_metrics":{"request_id":"req_1","request_duration_ms":21}}
}, CancellationToken)
HttpClientHandler.SendAsync(Method: GET, RequestUri: 'http://localhost:12111/v1/balance', Version: 1.1, Content: <null>, Headers:
{
  User-Agent: Stripe/v1
  User-Agent: .NetBindings/22.9.0
  Authorization: Bearer sk_test_123
  X-Stripe-Client-User-Agent: {"bindings_version":"22.9.0","lang":".net","publisher":"stripe","lang_version":".NET Core 4.6.25211.01","os_version":"Microsoft Windows 10.0.14393 "}
  Stripe-Version: 2018-11-08
  X-Stripe-Client-Telemetry: {"last_request_metrics":{"request_id":"req_1","request_duration_ms":22}}
}, CancellationToken)
```

Don't suspect that `this.prevRequestMetrics.TryDequeue(out requestMetrics)` is the issue, so maybe theres another subtler problem with the test?

ptal @ob-stripe

ob-stripe:
Ah yes, I think there's a potential race condition in the test itself. Incrementing `invocationCount` and generating the request ID string isn't an atomic operation, so two concurrent requests may end up with the same request ID. I'll try to fix this.

ob-stripe:
Okay, I've factorized the code for setting up the requests in a `FakeServer` class and added a lock when incrementing the counter and generating the request ID string to fix the race condition.

ptal @jameshageman-stripe

jameshageman-stripe:
@ob-stripe LGTM, thanks!

Issue #1515, opened 2019-02-14
Feature Request: Add models/entities for working with shipping+tax callbacks


HopesOblivion:
Looking at this: https://stripe.com/docs/orders/dynamic-shipping-taxes

I thought - OK cool neat I can just take the order request posted to my server and that'll deserialize right into the Order object from the library. It doesn't : /

It would be helpful if an 'order' from the above mentioned doc was the same as an order in the dotnet lib.

Alt: let us pass in shipping options + taxes on order creation instead of requiring the callback.

remi-stripe:
@HopesOblivion What part specifically doesn't work with stripe-dotnet exactly? Would you be able to provide a sample JSON that breaks your integration? As far as I know we would send the normal Order payload to your endpoint.

HopesOblivion:
@remi-stripe wow that was fast. The problem is that the order payload sent doesn't deserialize into a stripe-dotnet Order.

You can find a sample json here: https://stripe.com/docs/orders/dynamic-shipping-taxes

Ideally I'd like to just setup an API Controller with a method signature like ` public IActionResult(Order orderThatYoullPostMeInTheCallback) { // do stuff; return Ok(new CallBackResponseYouWant{ ...} );} `

Specific Issues:

- The payload is actually ` { order: {} } ` so you have to make a dummy class with a single order property or use JObject or dynamic (not a big deal really, just kind of weak)
- Stripe.OrderItem.Parent is a string, in the order payload the Parent is a fully expanded Sku
- The response that the callback is looking for isn't modeled in stripe-dotnet anywhere

So my feature request is - add Entities that accurately represent the Request and Response for the callback shipping & tax calculations so we can just plug them into our api.

I hope that's clear.

remi-stripe:
> The payload is actually `{ order: {} }` so you have to make a dummy class with a single order property or use JObject or dynamic (not a big deal really, just kind of weak)

Gotcha. Could you just parse the JSON and the deserialize what it's this onto `Order` instead?

> Stripe.OrderItem.Parent is a string, in the order payload the Parent is a fully expanded Sku

This seems like something we should support in stripe-dotnet since `parent` is expandable. It isa bit of an exception in our API though as it expands only if it relates to a SKU. I started a PR to get this discussed asI: https://github.com/stripe/stripe-dotnet/pull/1516

> The response that the callback is looking for isn't modeled in stripe-dotnet anywhere

[...]

PR #1516, opened 2019-02-14
Make `parent` on `OrderItem` expandable

remi-stripe:
Make `parent` on `OrderItem` expandable. This helps fix part of https://github.com/stripe/stripe-dotnet/issues/1515

This is a tricky one though. First because it's a breaking change as we original chose `Parent` and not `ParentId` as the name. Second, because `parent` is only expandable when it refers to a SKU and not otherwise. In stripe-go this is implemented [here](https://github.com/stripe/stripe-go/blob/master/order.go#L228-L257). In stripe-java, we avoided the issue by simply assuming that the expanded object is `HasId`.

Here, I used `Sku` but the alternative is likely to build an Interface just for it as we do for other polymorphic properties. Wanted to get your opinion first though.

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/Orders/OrderItem.cs
+++ b/src/Stripe.net/Entities/Orders/OrderItem.cs
@@ -1,6 +1,7 @@
 namespace Stripe
 {
     using Newtonsoft.Json;
+    using Stripe.Infrastructure;
 
     public class OrderItem : StripeEntity<OrderItem>, IHasObject
     {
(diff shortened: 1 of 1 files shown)


ob-stripe:
That's fine, `parent` can only be a `Sku`. (Initially it could be either a `Sku` or a `Discount`, but that never worked properly because discounts don't have IDs, which is a prerequisite for expansion. This was fixed a few months back.)

remi-stripe:
@ob-stripe Can you make that part of your next major release? Avoids a major version just for this.

PR #1518, opened 2019-02-15
Automatic request retries

ob-stripe:
cc @stripe/api-libraries 

Add support for automatic request retries.

The implementation is based on stripe-go, so the retry conditions are the same: requests are retried on connection errors (timeouts) and 409's. Given the recent discussions in e.g. https://github.com/stripe/stripe-node/pull/559, let me know if you want me to change the conditions (e.g. to retry on 500's).

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/RequestTelemetry.cs
+++ b/src/Stripe.net/Infrastructure/Public/RequestTelemetry.cs
@@ -1,9 +1,10 @@
 namespace Stripe
 {
+    using System;
     using System.Collections.Concurrent;
+    using System.Collections.Generic;
     using System.Linq;
     using System.Net.Http;
-    using System.Net.Http.Headers;
     using Newtonsoft.Json;
 
     /// <summary>
@@ -21,9 +22,14 @@ namespace Stripe
         /// a <c>X-Stripe-Client-Telemetry</c> header with the item; otherwise, do nothing.
         /// </summary>
         /// <param name="headers">The request headers.</param>
-        public void MaybeAddTelemetryHeader(HttpHeaders headers)
+        public void MaybeAddTelemetryHeader(IDictionary<string, string> headers)
         {
             if (!StripeConfiguration.EnableTelemetry)
+            {
+                return;
+            }
+
+            if (headers.ContainsKey("X-Stripe-Client-Telemetry"))
             {
                 return;
             }
@@ -44,8 +50,8 @@ namespace Stripe
         /// otherwise, do nothing.
         /// </summary>
         /// <param name="response">The HTTP response message.</param>
-        /// <param name="durationMs">The request duration, in milliseconds.</param>
-        public void MaybeEnqueueMetrics(HttpResponseMessage response, long durationMs)
+        /// <param name="duration">The request duration.</param>
+        public void MaybeEnqueueMetrics(HttpResponseMessage response, TimeSpan duration)
         {
             if (!StripeConfiguration.EnableTelemetry)
             {
@@ -67,7 +73,7 @@ namespace Stripe
             var metrics = new RequestMetrics
             {
                 RequestId = requestId,
-                RequestDurationMs = durationMs,
+                RequestDurationMs = (long)duration.TotalMilliseconds,
             };
 
             this.prevRequestMetrics.Enqueue(metrics);
--- a/src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
+++ b/src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
@@ -33,7 +33,17 @@ namespace Stripe
         public static string DefaultFilesBase => "https://files.stripe.com";
 
         /// <summary>Default timespan before the request times out.</summary>
-        public static TimeSpan DefaultHttpTimeout => new TimeSpan(80 * TimeSpan.TicksPerSecond);
+        public static TimeSpan DefaultHttpTimeout => TimeSpan.FromSeconds(80);
+
+        /// <summary>
+        /// Maximum sleep time between tries to send HTTP requests after network failure.
+        /// </summary>
+        public static TimeSpan MaxNetworkRetriesDelay => TimeSpan.FromSeconds(5);
+
+        /// <summary>
+        /// Minimum sleep time between tries to send HTTP requests after network failure.
+        /// </summary>
+        public static TimeSpan MinNetworkRetriesDelay => TimeSpan.FromMilliseconds(500);
 
         /// <summary>Gets or sets the base URL for Stripe's API.</summary>
         public static string ApiBase { get; set; } = DefaultApiBase;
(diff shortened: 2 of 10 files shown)


src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
```diff
@@ -24,6 +27,10 @@ public class SystemNetHttpClient : IHttpClient
 
         private readonly RequestTelemetry requestTelemetry = new RequestTelemetry();
 
+        private readonly object randLock = new object();
+
+        private readonly Random rand = new Random();
```
ob-stripe: This random number generator is used to generate the jitter in retry delays. Microsoft recommends creating a single generator instance rather than creating a new instance every time, but also warns that `Random` is not thread-safe, so a lock is needed. Cf. https://docs.microsoft.com/en-us/dotnet/api/system.random?view=netframework-4.7.2#Multiple.

src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
```diff
@@ -60,24 +67,62 @@ public static System.Net.Http.HttpClient BuildDefaultSystemNetHttpClient()
             StripeRequest request,
             CancellationToken cancellationToken = default(CancellationToken))
         {
-            var httpRequest = BuildRequestMessage(request);
+            TimeSpan duration;
+            HttpResponseMessage response = null;
+            int retry = 0;
+            OperationCanceledException timeoutException = null;
+
+            this.requestTelemetry.MaybeAddTelemetryHeader(request.StripeHeaders);
```
ob-stripe: I've changed `RequestTelemetry.MaybeAddTelemetryHeader()` to add the header to the `StripeRequest` instance rather than the `HttpRequestMessage` instance. This is because the `HttpRequestMessage` instance is consumed by every request attempt, so we need to generate a new one from the `StripeRequest` instance for each retry. This way, we don't lose metrics when a request fails.

src/StripeTests/Functional/TelemetryTest.cs
```diff
@@ -53,12 +53,12 @@ public void TelemetryEnabled()
         {
             this.ResetStripeClient();
             var fakeServer = FakeServer.ForMockHandler(this.MockHttpClientFixture.MockHandler);
-            fakeServer.DelayMilliseconds = 10;
+            fakeServer.Delay = TimeSpan.FromMilliseconds(20);
```
ob-stripe: This test was flaky and randomly failed. I think it's because the resolution of the timer used internally by `Task.Delay` is only around ~15 milliseconds, so even when the fake server was supposed to sleep for 10 milliseconds, we sometimes measured a duration of 8-9 milliseconds and the call to `Verify` would fail because we asserted that `d >= 10`.

I've relaxed the test by sleeping for a bit longer (20 milliseconds) and asserting that the duration is >= 15.

src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
```diff
@@ -60,24 +67,62 @@ public static System.Net.Http.HttpClient BuildDefaultSystemNetHttpClient()
...
-            var response = await this.httpClient.SendAsync(httpRequest, cancellationToken)
-                .ConfigureAwait(false);
-            stopwatch.Stop();
+                try
+                {
+                    response = await this.httpClient.SendAsync(httpRequest, cancellationToken)
+                        .ConfigureAwait(false);
+                }
+                catch (OperationCanceledException exception)
+                    when (!cancellationToken.IsCancellationRequested)
```
ob-stripe: Internally, `HttpClient` uses a `CancelationToken` to handle timeouts, so we need to use a `when` filter to distinguish between timeouts and user-requested cancelations.

src/Stripe.net/Infrastructure/Public/RequestTelemetry.cs
```diff
@@ -44,8 +50,8 @@ public void MaybeAddTelemetryHeader(HttpHeaders headers)
         /// otherwise, do nothing.
         /// </summary>
         /// <param name="response">The HTTP response message.</param>
-        /// <param name="durationMs">The request duration, in milliseconds.</param>
-        public void MaybeEnqueueMetrics(HttpResponseMessage response, long durationMs)
+        /// <param name="duration">The request duration.</param>
+        public void MaybeEnqueueMetrics(HttpResponseMessage response, TimeSpan duration)
```
brandur-stripe: _Way_ nicer.

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
```diff
@@ -85,6 +95,12 @@ public static string ApiKey
         /// </summary>
         public static JsonSerializerSettings SerializerSettings { get; set; } = DefaultSerializerSettings();
 
+        /// <summary>
+        /// Gets or sets the maximum number of times that the library will retry requests that
+        /// appear to have failed due to an intermittent problem.
+        /// </summary>
+        public static int MaxNetworkRetries { get; set; } = 0;
```
brandur-stripe: Man, they let you set defaults for auto-properties now too?! Love how much boilerplate they've stripped out of this language.

src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
brandur-stripe: Nice research. (And wow, a thread-safe convenient wrapper for `Rand` seems like a pretty obvious improvement for a future version of .NET.)

src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
```diff
@@ -122,5 +167,62 @@ private static string BuildStripeClientUserAgentString()
...
+            {
+                return false;
+            }
+
+            if (timeout == true)
+            {
+                return true;
+            }
+
+            if (statusCode == HttpStatusCode.Conflict)
```
brandur-stripe: It might be worth adding 503s here and non-`POST` 500s as recently came into stripe-node. I think that's the direction we're going, and we just haven't changed stripe-go yet.

ob-stripe:
I've updated `ShouldRetry` to also retry on 503's and non-POST 5xx's, similar to stripe-node.

I also realized there were a few important bits missing, which I also updated:
- we now set an idempotency key for all POST requests, if one is not already provided by the user (it would have been pretty dangerous to ship automatic retries without this!)
- non-timeout connection errors are reported as `HttpRequestException`s. These are now handled like timeout errors.
- I've also improved error handling to be a bit more resilient with unexpected bodies (I was getting `NullPointerException`s in some of my tests because I was returning `{}` as the body and the error handling code tries to access the `error` key in the JSON hash)

ptal @brandur-stripe

brandur-stripe:
> * we now set an idempotency key for all POST requests, if one is not already provided by the user (it would have been pretty dangerous to ship automatic retries without this!)

Phew! Glad you caught that one before shipping, haha.

LGTM!

Issue #1522, opened 2019-02-18
PaymentIntentService.Create = error on SaveSourceToCustomer


mnemanov:
Re: commit 206a44ca49e6cc6787732094ca771399001ed316

You changed the taget api version to the latest but forgot to implement this breaking change:

`save_source_to_customer has been renamed to save_payment_method`

remi-stripe:
@mnemanov Thanks for the heads up and sorry for the issue here. I'll work on a fix now.

mnemanov:
Thank you.

I just realized I can't even override the version on this call since you made RequestOptions.ApiVersion internal.

remi-stripe:
That's expected. stripe-dotnet is pinned to an API version (and always has been). You can technically use `AddExtraParam` to pass `save_payment_method` in the meantime while I work on the fix.

remi-stripe:
I'm working on the release and will update once it's done.

mnemanov:
Thanks, that's a quick response.

(I should that this is the reason we selected stripe for this project).

(Now if you can get me a stripe terminal it would be great).

remi-stripe:
@mnemanov I just released 23.1.0. It should appear on NuGet in a bit!

mnemanov:
Thanks again.

PR #1523, opened 2019-02-18
Add save_payment_method and deprecate save_source_to_customer

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/1522

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
We're about to release a new major, can we just get rid of the old attribute?

remi-stripe:
@ob-stripe I need to release the other PR as a non major so I thought it'd be nice to do both and unblock @mnemanov at the same time.

PR #1527, opened 2019-02-19
Fix a couple of options classes

ob-stripe:
r? @remi-stripe 

Just noticed that there are a couple of options classes that inherit `StripeEntity` instead of `INestedOptions`.


ob-stripe:
No worries, I should have caught it during the review too.

ob-stripe:
Fixed in 24.0.1.

Issue #1528, opened 2019-02-19
System.Collections.Immutable Dependancy


DanWM:
Version 24.0.0 dependency on System.Collections.Immutable.dll is triggering build errors in our ASP.NET Webforms project related to other System.Collection assemblies.  Is it necessary?  I was using version 21.7.0 prior.

ob-stripe:
@DanWM The dependency was added back in version 22 (#1428). We could technically remove it, but it is good practice to use immutable types where possible, and [System.Collections.Immutable](https://www.nuget.org/packages/System.Collections.Immutable/) should be compatible with all environments where Stripe.net is usable (it is compatible with .NET Framework 4.5 and .NET Standard 1.0, while Stripe.net's own minimum requirements are .NET Framework 4.5.1 and .NET Standard 1.2).

Could you tell me more about the build errors you're observing?

DanWM:
The problem appears to be with System.Collections.Immutable and System.Runtime.
It appears to require a web.config assembly reference to System.Runtime.   I'm not sure why.   The rest of the project has not needed it.   This has allowed my application to run now.

Also, based on the release notes since version 22.0.0 was released, I adjusted the PaymentIntentCreateoptions.  I'm successfully use the following parameters:
Amount, Currency,  Description,  PaymentMethodTypes, CaptureMethod, ApplicationFeeAmount,                 OnBehalfOf, TransferData,  PaymentIntentTransferDataOptions  { Destination }

However, I'm now suddenly getting an error when I later attempt to update just the Description of the  payment intent. I get the error below:
Newtonsoft.Json.JsonSerializationException
  HResult=0x80131500
  Message=Error converting value {null} to type 'System.Int64'. Path 'transfer_data.amount', line 121, position 24.
  Source=Newtonsoft.Json

ob-stripe:
System.Collections.Immutable does depend on NETStandard.Library if you're using .NET Standard 1.x, and NETStandard.Library itself depends on System.Runtime. I'm not sure why System.Runtime would cause any issues though.

The JSON deserialization error is unrelated. I've found the cause and will release a fix shortly, thanks for reporting!

[...]

PR #1531, opened 2019-02-20
Make ChargeTransferData.Amount nullable

ob-stripe:
cc @remi-stripe 

The API can return `null` for this attribute, so the Stripe.net property needs to be nullable, otherwise trying to deserialize `null` will cause an exception.

Fixes the issue described in the second part of this comment: https://github.com/stripe/stripe-dotnet/issues/1528#issuecomment-465342165.

(We should start thinking about a larger solution to this problem since it's so easy to get this wrong. The sledgehammer solution would be to make all value types nullable, like we already in options classes.)


ob-stripe:
Fixed in 24.0.2.

Issue #1532, opened 2019-02-20
PaymentIntentService.Confirm - ClientSecret is missing


bnserig:
Hi,
I'm using Stripe.NET in a Xamarin Forms application.
I try to confirm a PaymentIntent with the publishable key but the property ClientSecret is missing in PaymentIntentConfirmOptions.
Can you add this property to this object ?

Thank you.

remi-stripe:
@bnserig Thanks for the report! I'll work on a fix!

remi-stripe:
@bnserig This fix was released as 24.1.0. Really sorry about the delay on that one, I forgot to release last week.

bnserig:
Thanks for the release.

PR #1533, opened 2019-02-20
Multiple fixes for PaymentIntent

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/1532 and other issues

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/PaymentIntents/PaymentIntentTransferData.cs
+++ b/src/Stripe.net/Entities/PaymentIntents/PaymentIntentTransferData.cs
@@ -1,13 +1,11 @@
 namespace Stripe
 {
+    using System;
     using Newtonsoft.Json;
     using Stripe.Infrastructure;
 
     public class PaymentIntentTransferData : StripeEntity
     {
-        [JsonProperty("amount")]
-        public long Amount { get; set; }
-
         #region Expandable Destination (Account)
         [JsonIgnore]
         public string DestinationId { get; set; }
@@ -29,5 +27,9 @@ namespace Stripe
             }
         }
         #endregion
+
+        [Obsolete("Use PaymentIntent.ApplicationFeeAmount instead")]
+        [JsonProperty("amount")]
+        public long? Amount { get; set; }
     }
 }
--- a/src/Stripe.net/Services/PaymentIntents/PaymentIntentSharedOptions.cs
+++ b/src/Stripe.net/Services/PaymentIntents/PaymentIntentSharedOptions.cs
@@ -36,6 +36,9 @@ namespace Stripe
         [JsonProperty("save_payment_method")]
         public bool? SavePaymentMethod { get; set; }
 
+        [JsonProperty("shipping")]
+        public ChargeShippingOptions Shipping { get; set; }
+
         [JsonProperty("source")]
         public string SourceId { get; set; }
 
--- a/src/Stripe.net/Services/PaymentIntents/PaymentIntentTransferDataOptions.cs
+++ b/src/Stripe.net/Services/PaymentIntents/PaymentIntentTransferDataOptions.cs
@@ -1,14 +1,16 @@
 namespace Stripe
 {
+    using System;
     using Newtonsoft.Json;
 
     public class PaymentIntentTransferDataOptions : INestedOptions
     {
-        [JsonProperty("amount")]
-        public long? Amount { get; set; }
-
         // This is only available on creation and not update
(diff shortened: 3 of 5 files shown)


src/Stripe.net/Entities/PaymentIntents/PaymentIntentTransferData.cs
```diff
@@ -6,7 +6,7 @@ namespace Stripe
     public class PaymentIntentTransferData : StripeEntity
     {
         [JsonProperty("amount")]
-        public long Amount { get; set; }
+        public long? Amount { get; set; }
```
remi-stripe: Is this a breaking change?

ob-stripe: In the strictest sense, yes: code that uses this property will probably need to be updated to handle the nullability. That said, given that very few people are using this, and anyone who does is probably seeing deserialization errors, I think it's fine to consider this a bug fix and release as a patch version (though the rest of the changes in this PR qualify this as a minor, not a patch).

remi-stripe:
@ob-stripe I made some extra changes, can you re-review and if it's all good release a version?

src/Stripe.net/Services/PaymentIntents/PaymentIntentSharedOptions.cs
```diff
@@ -27,15 +27,18 @@ public class PaymentIntentSharedOptions : BaseOptions
...
         public string OnBehalfOf { get; set; }
 
-        [JsonProperty("payment_method_types")]
-        public List<string> PaymentMethodTypes { get; set; }
-
         [JsonProperty("receipt_email")]
         public string ReceiptEmail { get; set; }
 
+        [JsonProperty("payment_method_types")]
+        public List<string> PaymentMethodTypes { get; set; }
```
ob-stripe: Looks like this one was already in the correct alphabetical order, can you move it back?

src/Stripe.net/Entities/PaymentIntents/PaymentIntentTransferData.cs
```diff
@@ -29,5 +27,9 @@ internal object InternalDestination
             }
         }
         #endregion
+
+        [Obsolete("Use ApplicationFeeAmount instead")]
```
ob-stripe: The class doesn't have an `ApplicationFeeAmount` property. Can you change the message to `"Use PaymentIntent.ApplicationFeeAmount"`?

src/Stripe.net/Services/PaymentIntents/PaymentIntentTransferDataOptions.cs
```diff
@@ -1,14 +1,16 @@
...
     public class PaymentIntentTransferDataOptions : INestedOptions
     {
-        [JsonProperty("amount")]
-        public long? Amount { get; set; }
-
         // This is only available on creation and not update
         [JsonProperty("destination")]
         public string Destination { get; set; }
+
+        [Obsolete("Use ApplicationFeeAmount instead")]
```
ob-stripe: Same here, can you change the message to `"Use PaymentIntent*Options.ApplicationFeeAmount"`?

remi-stripe:
@ob-stripe thanks for catching all the issues. PTAL.

src/Stripe.net/Services/PaymentIntents/PaymentIntentTransferDataOptions.cs
```diff
@@ -1,14 +1,16 @@
...
     public class PaymentIntentTransferDataOptions : INestedOptions
     {
-        [JsonProperty("amount")]
-        public long? Amount { get; set; }
-
         // This is only available on creation and not update
         [JsonProperty("destination")]
         public string Destination { get; set; }
+
+        [Obsolete("Use PaymentIntent.ApplicationFeeAmount instead")]
```
ob-stripe: Sorry, nitpicking, but can you change this to something like `"Use PaymentIntent*Options.ApplicationFeeAmount"`, or maybe `"Use ApplicationFeeAmount on the parent options class"`?

`PaymentIntent.ApplicationFeeAmount` refers to the property on the entity class, not the options class, so it's a bit misleading.

remi-stripe: yep that's fair. PTAL

PR #1538, opened 2019-02-22
Enforce that all properties have a Json attribute

ob-stripe:
r? @remi-stripe 

This PR adds a wholesome test that checks that all properties on entities and options classes have either a `[JsonProperty("...")]` or `[JsonIgnore]` attribute.

I also plan on adding a second test to ensure that properties with types that require a custom converter do declare and use the correct converter (e.g. `DateTime` properties should be annotated with `[JsonConverter(typeof(DateTimeConverter))]`, but it's going to be a little more convoluted, so I'll do that in a followup PR.

### DIFF
--- a/src/Stripe.net/Entities/_base/StripeEntity.cs
+++ b/src/Stripe.net/Entities/_base/StripeEntity.cs
@@ -5,6 +5,7 @@ namespace Stripe
     using System.Runtime.CompilerServices;
     using Newtonsoft.Json;
 
+    [JsonObject(MemberSerialization.OptIn)]
     public abstract class StripeEntity : IStripeEntity
     {
         [JsonIgnore]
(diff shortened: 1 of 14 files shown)


src/Stripe.net/Entities/_base/StripeEntity.cs
```diff
@@ -5,6 +5,7 @@ namespace Stripe
     using System.Runtime.CompilerServices;
     using Newtonsoft.Json;
 
+    [JsonObject(MemberSerialization.OptIn)]
```
remi-stripe: usual "I don't know what this means, are you sure we won't break anyone?" comment

ob-stripe: Right, sorry. [`MemberSerialization.OptIn`](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_MemberSerialization.htm) changes the behavior of Newtonsoft.Json like this:
> Only members marked with `JsonPropertyAttribute` [...] are serialized.

In other words, this causes members that _don't_ have an explicit Json attribute to be treated as if they were annotated with `JsonIgnore`. Since this PR also forces us to _not_ have any members without an explicit Json attribute, this is not really needed anyway, it's just an additional safety measure.

PR #1539, opened 2019-02-22
Add `created` and `status_transitions` on `Invoice`

remi-stripe:
Technically the List Invoice API does not yet take `created` but will tomorrow or the day after. I want the PR to be ready/approved and I'll only merge once it works.

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Will the API return both `created` and `date` with the same value?

remi-stripe:
> Will the API return both `created` and `date` with the same value?

Correct yes and a new API version will remove `date` in the future.

PR #1541, opened 2019-02-25
Add support for the PaymentMethod resource

remi-stripe:
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails.cs
+++ b/src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails.cs
@@ -1,0 +1,56 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class ChargePaymentMethodDetails : StripeEntity
+    {
+        [JsonProperty("ach_credit_transfer")]
+        public ChargePaymentMethodDetailsAchCreditTransfer AchCreditTransfer { get; set; }
+
+        [JsonProperty("ach_debit")]
+        public ChargePaymentMethodDetailsAchDebit AchDebit { get; set; }
+
+        [JsonProperty("alipay")]
+        public ChargePaymentMethodDetailsAlipay Alipay { get; set; }
+
+        [JsonProperty("bancontact")]
+        public ChargePaymentMethodDetailsBancontact Bancontact { get; set; }
+
+        [JsonProperty("bitcoin")]
+        public ChargePaymentMethodDetailsBitcoin Bitcoin { get; set; }
+
+        [JsonProperty("card")]
+        public ChargePaymentMethodDetailsCard Card { get; set; }
+
+        [JsonProperty("card_present")]
+        public ChargePaymentMethodDetailsCardPresent CardPresent { get; set; }
+
+        [JsonProperty("eps")]
+        public ChargePaymentMethodDetailsEps Eps { get; set; }
+
+        [JsonProperty("giropay")]
+        public ChargePaymentMethodDetailsGiropay Giropay { get; set; }
+
+        [JsonProperty("ideal")]
+        public ChargePaymentMethodDetailsIdeal Ideal { get; set; }
+
+        [JsonProperty("multibanco")]
+        public ChargePaymentMethodDetailsMultibanco Multibanco { get; set; }
+
+        [JsonProperty("p24")]
+        public ChargePaymentMethodDetailsP24 P24 { get; set; }
+
+        [JsonProperty("sepa_debit")]
+        public ChargePaymentMethodDetailsSepaDebit SepaDebit { get; set; }
+
+        [JsonProperty("stripe_account")]
+        public ChargePaymentMethodDetailsStripeAccount StripeAccount { get; set; }
+
+        [JsonProperty("type")]
+        public string Type { get; set; }
+
+        [JsonProperty("wechat")]
+        public ChargePaymentMethodDetailsWechat Wechat { get; set; }
+    }
+}
--- a/src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails/ChargePaymentMethodDetailsCardPresent.cs
+++ b/src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails/ChargePaymentMethodDetailsCardPresent.cs
@@ -1,0 +1,41 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class ChargePaymentMethodDetailsCardPresent : StripeEntity
+    {
+        [JsonProperty("brand")]
+        public string Brand { get; set; }
+
+        [JsonProperty("country")]
+        public string Country { get; set; }
+
+        [JsonProperty("emv_auth_data")]
+        public string EmvAuthData { get; set; }
+
+        [JsonProperty("exp_month")]
+        public long ExpMonth { get; set; }
+
+        [JsonProperty("exp_year")]
+        public long ExpYear { get; set; }
+
+        [JsonProperty("fingerprint")]
+        public string Fingerprint { get; set; }
+
+        [JsonProperty("funding")]
+        public string Funding { get; set; }
+
+        [JsonProperty("generated_card")]
+        public string GeneratedCardId { get; set; }
+
+        [JsonProperty("last4")]
+        public string Last4 { get; set; }
+
+        [JsonProperty("read_method")]
+        public string ReadMethod { get; set; }
+
+        [JsonProperty("receipt")]
+        public ChargePaymentMethodDetailsCardPresentReceipt Receipt { get; set; }
+    }
+}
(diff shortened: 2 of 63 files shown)


src/Stripe.net/Entities/PaymentMethods/PaymentMethodCardWallet.cs
```diff
@@ -0,0 +1,34 @@
...
+        [JsonProperty("dynamic_last4")]
+        public string DynamicLast4 { get; set; }
+
+        [JsonProperty("google_pay")]
+        public PaymentMethodCardWalletGooglePay GooglePay { get; set; }
+
+        [JsonProperty("masterpass")]
+        public PaymentMethodCardWalletMasterpass Masterpass { get; set; }
+
+        [JsonProperty("microsoft")]
```
brandur-stripe: Should this be `microsoft_pay`?

remi-stripe: In theory it should. It wasn't when I did the PR and flagged it internally :p 
Also it's possible it will not be in the final PR!

src/Stripe.net/Entities/PaymentIntents/PaymentIntentLastPaymentError.cs
```diff
@@ -23,6 +23,12 @@ public class PaymentIntentLastPaymentError : StripeEntity
         [JsonProperty("param")]
         public string Param { get; set; }
 
+        [JsonProperty("payment_intent")]
```
remi-stripe: Should we fix this to be a StripeError instead and release as a major version?

src/Stripe.net/Services/PaymentMethods/PaymentMethodCreateOptions.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        [JsonProperty("billing_details")]
+        public BillingDetailsOptions BillingDetails { get; set; }
+
+        /// <summary>
+        /// If this is a <c>card</c> PaymentMethod, for backwards compatibility, you can  provide a
+        /// Stripe token (e.g., for apple pay, amex checkout, or legacy checkout).
+        /// </summary>
+        [JsonProperty("card")]
+        public string CardTokenId { get; set; }
```
remi-stripe: This API takes a token id for legacy reasons, or a hash (for PCI compliant users + .net mobile apps). Are we still okay taking that approach?
The alternative could be to not support this in .NET so not sure what's best.

I also put my own comments there since the ones in the docs would not work.

src/StripeTests/Services/BankAccounts/BankAccountServiceTest.cs
```diff
@@ -26,19 +26,7 @@ public BankAccountServiceTest(MockHttpClientFixture mockHttpClientFixture)
...
-                    Currency = "usd",
-                    AccountHolderName = "John Doe",
-                    AccountHolderType = BankAccountHolderType.Company,
-                    RoutingNumber = "110000000",
-                    Metadata = new Dictionary<string, string>
-                    {
-                        { "key", "value" },
-                    },
-                }
+                SourceToken = "btok_123",
```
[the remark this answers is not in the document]
mickjermsurawong-stripe: Ah sorry, I didn't foresee it would fail current library tests. Yup, it is my work on typed params to make public spec for these params only taking String as an attempt to limit non-PCI compliant users from accidentally passing in raw PAN data. So as I understand, we want to make this enforcement on in the client lib too, so passing String here would be okay.

Just a side note: on live request OpenAPI validation, we skipped validation on these undocumented polymorphic param so we won't see false rate of validation failure. (Additionally, there's no sever side effect because its' OpenAPI)

src/StripeTests/Services/ExternalAccounts/ExternalAccountServiceTest.cs
```diff
@@ -25,13 +25,7 @@ public ExternalAccountServiceTest(MockHttpClientFixture mockHttpClientFixture)
...
             this.createOptions = new ExternalAccountCreateOptions
             {
-                ExternalAccountBankAccount = new AccountBankAccountOptions
-                {
-                    AccountNumber = "000123456789",
-                    Country = "US",
-                    Currency = "usd",
-                    RoutingNumber = "110000000",
-                }
+                ExternalAccountTokenId = "btok_123",
```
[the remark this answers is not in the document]
mickjermsurawong-stripe: Ack, yup it applies to external account to. The only old endpoint that takes full hash is the token creation params that takes full card/bank account

src/Stripe.net/Services/PaymentMethods/PaymentMethodCreateOptions.cs
mickjermsurawong-stripe: Ah I see. Hmm so there's a valid use case of passing raw PANS specifically at this new endpoint (mobile apps)? On previous decision we assume that merchants passing raw PANS will already be doing custom integration anyways right? But on the .net mobile apps, can we make exception only for this new PaymentMethod, i.e can we enforce to passing token in old endpoints with params  `source` and `external_account`, and not this?

remi-stripe: I think it's fine to keep the existing code for now. We will do a major version in the future that removes this ability on endpoints that should be called server-side. But PaymentMethod creation can happen client-side in a mobile application and since we don't have a .NET mobile SDK, users rely on stripe-dotnet for this.

src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails/ChargePaymentMethodDetailsIdeal.cs
```diff
@@ -0,0 +1,20 @@
...
+        public string BankCode { get; set; }
+
+        [JsonProperty("bank_name")]
+        public string BankName { get; set; }
+
+        [JsonProperty("bic")]
+        public string Bic { get; set; }
+
+        [JsonProperty("verified_name")]
+        public string VerifiedName { get; set; }
```
mickjermsurawong-stripe: I think we got the schema from Giropay above instead. Ideal should have `iban_last4` and  `bank`

src/Stripe.net/Services/PaymentMethods/PaymentMethodUpdateOptions.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class PaymentMethodUpdateOptions : BaseOptions
+    {
+        [JsonProperty("metadata")]
+        public Dictionary<string, string> Metadata { get; set; }
```
mickjermsurawong-stripe: I compare with autogen params, and we seem to allow updating card and billing details too..
Maybe we don't want this?

remi-stripe: Nice catch, forgot to fix that one. cc @ob-stripe Are we okay having a nested option have names such as `PaymentMethodCardCreateOptions` and `PaymentMethodCardUpdateOptions`?

remi-stripe:
@mickjermsurawong-stripe Thanks for the thorough reviews and catching multiple mistakes here!

mickjermsurawong-stripe:
Thanks for the fix remi. The schema looks great to me!
I defer to @ob-stripe for the pending issue on the payment method update options.

src/Stripe.net/Entities/PaymentIntents/PaymentIntentLastPaymentError.cs
ob-stripe: I'd say yes.

src/Stripe.net/Services/PaymentMethods/PaymentMethodUpdateOptions.cs
ob-stripe: I think that's fine. I assume most users rely on Intellisense to discover classes anyway.

src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        [JsonProperty("p24")]
+        public ChargePaymentMethodDetailsP24 P24 { get; set; }
+
+        [JsonProperty("sepa_debit")]
+        public ChargePaymentMethodDetailsSepaDebit SepaDebit { get; set; }
+
+        [JsonProperty("stripe_account")]
+        public ChargePaymentMethodDetailsStripeAccount StripeAccount { get; set; }
+
```
ob-stripe: `type` is missing.

src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails/ChargePaymentMethodDetailsCardPresent.cs
```diff
@@ -0,0 +1,41 @@
...
+        public long ExpYear { get; set; }
+
+        [JsonProperty("fingerprint")]
+        public string Fingerprint { get; set; }
+
+        [JsonProperty("funding")]
+        public string Funding { get; set; }
+
+        [JsonProperty("generated_card")]
+        public string GeneratedCard { get; set; }
```
ob-stripe: If this is an ID, this should be `GeneratedCardId`.

src/Stripe.net/Entities/PaymentIntents/PaymentIntentLastPaymentError.cs
ob-stripe: (Although maybe just bundle this with the next release for the new API version?)

PR #1542, opened 2019-02-25
Improved OAuth support

### LINKED ISSUE #1543: assert_capabilities missing from OAuthTokenCreateOptions
The OAuth API supports asserting that certain requested capabilities have been applied to the relationship, but the SDK doesn't currently accept those parameters. Detailed here: 

https://stripe.com/docs/connect/express-accounts#verify-the-accounts-capability

In the meantime, I'm using `AddExtraParam` to work around the issue. Thanks!

ob-stripe:
r? @remi-stripe 

Improved support for OAuth:
- new `OAuthTokenService.AuthorizeUrl()` method for creating the authorize URLs
- added XML comments for every OAuth parameter and attribute
- removed `Error` and `ErrorDescription` on `OAuthToken` and `OAuthDeauthorize`. These were useless as when the API returns an error, a `StripeError` is deserialized and attached to the `StripeException`.
- renamed `OAuthTokenDeauthorizeOptions` to `OAuthDeauthorizeOptions` (not a breaking change, since `OAuthTokenDeauthorizeOptions` was added earlier in the integration branch)

Fixes #1543.


src/Stripe.net/Entities/OAuth/OAuthToken.cs
```diff
@@ -4,31 +4,62 @@ namespace Stripe
...
+        /// as you would any Stripe secret API key.
+        /// </para>
+        /// <para>
+        /// This key does not expire, but may be revoked by the user at any time (you'll get a
+        /// <c>account.application.deauthorized</c>
+        /// <see href="https://stripe.com/docs/api/events/types#event_types-account.application.authorized">webhook event</see>
+        /// when this happens).
+        /// </para>
+        /// </summary>
+        [JsonProperty("access_token")]
```
remi-stripe: Can we mark this as deprecated in favour of the `Stripe-Account` header? We really should not recommend that anyone uses this property.

src/Stripe.net/Entities/OAuth/OAuthToken.cs
```diff
@@ -4,31 +4,62 @@ namespace Stripe
...
+        public string AccessToken { get; set; }
 
+        /// <summary>
+        /// <para>
+        /// The live mode indicator for the token. If <c>true</c>, the <see cref="AccessToken"/>
+        /// can be used as a live secret key. If <c>false</c>, the <see cref="AccessToken"/> can be
+        /// used as a test secret key.
+        /// </para>
+        /// <para>Depends on the mode of the secret API key used to make the request.</para>
+        /// </summary>
```
remi-stripe: Same as a above, we could say that requests in test-only or both live/test would work instead of mentioning the access token.

src/Stripe.net/Entities/OAuth/OAuthToken.cs
```diff
@@ -4,31 +4,62 @@ namespace Stripe
...
-        [JsonProperty("stripe_user_id")]
-        public string StripeUserId { get; set; }
-
-        [JsonProperty("stripe_publishable_key")]
-        public string StripePublishableKey { get; set; }
-
+        /// <summary>
+        /// Can be used to get a new access token of an equal or lesser scope, or of a different
+        /// live mode (where
+        /// <see href="https://stripe.com/docs/connect/testing#test-keys-for-livemode-applications">applicable</see>).
```
remi-stripe: IMO that's a bad link about getting test mode access and not really about how to refresh (though the doc you're looking for does not exist anymore.

src/Stripe.net/Entities/OAuth/OAuthToken.cs
```diff
@@ -4,31 +4,62 @@ namespace Stripe
...
+        /// The scope granted to the access token, depending on the scope of the authorization code
+        /// and <c>scope</c> parameter.
+        /// </summary>
+        [JsonProperty("scope")]
+        public string Scope { get; set; }
+
+        /// <summary>
+        /// A publishable key that can be used with this account. Matches the mode—live or test—of
+        /// the token.
+        /// </summary>
```
remi-stripe: same.

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlOptions.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    using Newtonsoft.Json;
+
+    public class OAuthAuthorizeUrlOptions : BaseOptions
+    {
+        /// <summary>
+        /// Boolean to indicate that the user should always be asked to connect, even if they're
+        /// already connected. Defaults to <c>false</c>.
+        /// </summary>
+        [JsonProperty("always_prompt")]
```
remi-stripe: Is that a real thing? I tried quickly and could not see what it would do.

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlOptions.cs
```diff
@@ -0,0 +1,76 @@
...
+        public bool? AlwaysPrompt { get; set; }
+
+        /// <summary>
+        /// The unique identifier provided to your application, found in your
+        /// <see href="https://dashboard.stripe.com/account/applications/settings">application settings</see>.
+        /// </summary>
+        [JsonProperty("client_id")]
+        public string ClientId { get; set; } = StripeConfiguration.ClientId;
+
+        /// <summary>The only option at the moment is <c>code</c>.</summary>
```
remi-stripe: We usually do the comment on 3 lines.
also is `<c>....</c>` new?

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlOptions.cs
```diff
@@ -0,0 +1,76 @@
...
+        [JsonProperty("response_type")]
+        public string ResponseType { get; set; } = "code";
+
+        /// <summary>
+        /// <para>
+        /// The URL for the authorize
+        /// <see href="https://stripe.com/docs/connect/oauth-reference#get-authorize-response">response</see>
```
remi-stripe: Why mention this link instead of just calling it a redirect url?

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlOptions.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </para>
+        /// <para>
+        /// Defaults to <c>login</c> for scope <c>read_only</c> and <c>register</c> for scope
+        /// <c>read_write</c>.
+        /// </para>
+        /// </summary>
+        [JsonProperty("stripe_landing")]
+        public string StripeLanding { get; set; }
+
+        /// <summary>Prefilled details in the account form for new users.</summary>
```
remi-stripe: same, 3 lines.

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlStripeUserOptions.cs
```diff
@@ -0,0 +1,267 @@
...
+        /// This parameter is only relevant for Japan. You must prefill <see cref="Country"/> with
+        /// <c>JP</c> and <see cref="Zip"/> with a valid Japanese postal code to use this parameter.
+        /// </para>
+        /// </summary>
+        [JsonProperty("building_kanji")]
+        public string BuildingKanji { get; set; }
+
+        /// <summary>The legal name of the business, also used for the statement descriptor.</summary>
+        [JsonProperty("business_name")]
+        public string BusinessName { get; set; }
```
remi-stripe: Hum. I have quite a lot of FUD here. We just changed the shape of the Account API and it should now be `business_profile[name]`. We likely made things backwards compatible but I am not a fan of adding old/deprecated options to the library like this.

Sadly, it seems we forgot to add support for new properties in the OAuth flow with the new shape.

src/Stripe.net/Entities/OAuth/OAuthToken.cs
ob-stripe: 👍

src/Stripe.net/Entities/OAuth/OAuthToken.cs
ob-stripe: All the XML comments are copied straight from our docs and I'd rather not diverge from them, as I think having conflicting information would be even more confusing.

src/Stripe.net/Entities/OAuth/OAuthToken.cs
ob-stripe: 👍

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlOptions.cs
ob-stripe: ¯\\\_(ツ)\_/¯ It's documented here: https://stripe.com/docs/connect/oauth-reference#get-authorize

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlOptions.cs
ob-stripe: I sometimes put the comment on a single line if it fits under the 100 characters. Not a big deal one way or the other, it's rendered the same way.

`<c></c>` is used to inline code, like backquotes in Markdown: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/code-inline. It's already used in plenty of other comments.

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlOptions.cs
ob-stripe: Copied straight from the docs.

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlStripeUserOptions.cs
ob-stripe: It's unlikely we'll change the shape of the OAuth API anytime soon.

ob-stripe:
@remi-stripe I've marked `AccessToken` and `StripePublishableKey` as deprecated, and added the missing Express properties mentioned in #1543. ptal

src/Stripe.net/Entities/OAuth/OAuthDeauthorize.cs
```diff
@@ -4,13 +4,12 @@ namespace Stripe
...
+        /// returned, the revocation was successful.
+        /// </summary>
         [JsonProperty("stripe_user_id")]
         public string StripeUserId { get; set; }
-
-        [JsonProperty("error")]
-        public string Error { get; set; }
-
-        [JsonProperty("error_description")]
-        public string ErrorDescription { get; set; }
```
remi-stripe: is this a breaking change?

src/Stripe.net/Entities/OAuth/OAuthToken.cs
```diff
@@ -1,34 +1,78 @@
...
-
-        [JsonProperty("stripe_publishable_key")]
-        public string StripePublishableKey { get; set; }
-
+        /// <summary>
+        /// Can be used to get a new access token of an equal or lesser scope, or of a different
+        /// live mode (where
+        /// <see href="https://stripe.com/docs/connect/testing#test-keys-for-livemode-applications">applicable</see>).
+        /// </summary>
         [JsonProperty("refresh_token")]
```
remi-stripe: Can you add a comment about actively deprecating this?

src/Stripe.net/Entities/OAuth/OAuthToken.cs
remi-stripe: I could say we should edit the API ref or docs then, but fair.

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlStripeUserOptions.cs
remi-stripe: I disagree. Like we should _at least_ support all the new fields explicitly with the new layout and then have only those in the libs IMO. Have you asked Connect first?

src/Stripe.net/Entities/OAuth/OAuthDeauthorize.cs
ob-stripe: It isn't, as those attributes were never used. When an OAuth error occurs, the library raises a `StripeError` with the `Error` and `ErrorDescription` filled.

src/Stripe.net/Entities/OAuth/OAuthToken.cs
ob-stripe: 👍 Done.

src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlStripeUserOptions.cs
ob-stripe: Connect has confirmed that there are no plans to change the shape of the OAuth API.

PR #1546, opened 2019-02-28
Add support for `api_version` on the `WebhookEndpoint` resource

remi-stripe:
r? @mickjermsurawong-stripe 
cc @stripe/api-libraries 

The broken test is about charge expansion which is fixed in the other PR.

### DIFF
--- a/src/Stripe.net/Services/WebhookEndpoints/WebhookEndpointCreateOptions.cs
+++ b/src/Stripe.net/Services/WebhookEndpoints/WebhookEndpointCreateOptions.cs
@@ -7,6 +7,17 @@ namespace Stripe
 
     public class WebhookEndpointCreateOptions : WebhookEndpointSharedOptions
     {
+        /// <summary>
+        /// Events sent to this endpoint will be generated with this API version instead of your
+        /// account's default API version. We recommend that you set this to the API version that
+        /// the library is pinned to in <code>StripeConfiguration.stripeApiVersion</code>
+        /// </summary>
+        [JsonProperty("api_version")]
+        public string ApiVersion { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


src/Stripe.net/Services/WebhookEndpoints/WebhookEndpointCreateOptions.cs
```diff
@@ -7,6 +7,17 @@ namespace Stripe
 
     public class WebhookEndpointCreateOptions : WebhookEndpointSharedOptions
     {
+        /// <summary>
+        /// Events sent to this endpoint will be generated with this API version instead of your
+        /// account's default API version. We recommend that you set this to the API version that
+        /// the library is pinned to in <code>StripeConfiguration.stripeApiVersion</code>
+        /// </summary>
+        [JsonProperty("api_version")]
```
mickjermsurawong-stripe: Thank you for the comments! 
Given that in .NET we have this check which only gets triggered during real integration maybe we should add a comment here: "If set with different from this pinned version, 
set the `throwOnApiVersionMismatch` to false in `<code>EventUtility.ParseEvent</code>`
https://github.com/stripe/stripe-dotnet/blob/4b316d90ce0e481362a65969709b6a1e52e07faf/src/Stripe.net/Services/Events/EventUtility.cs#L21-L31

mickjermsurawong-stripe:
@remi-stripe just a nit on comments please
ptal @remi-stripe

remi-stripe:
@mickjermsurawong-stripe fixed the comment (and the failing test so that it's done). PTAL

mickjermsurawong-stripe:
The remove assertions on charge looks correct! Thank you remi
LGTM

PR #1559, opened 2019-03-25
Added Incomplete and Incomplete Expired statuses

Turnerj:
These statuses are part of the [documented API](https://stripe.com/docs/api/subscriptions/object#subscription_object-status) and were missing from this helper class.


ob-stripe:
Thanks @Turnerj!

ob-stripe:
The build failed because this is an external PR and we use secure variables for coverage reports in CI. This is safe to merge.

ob-stripe:
Released as 25.2.0.

PR #1565, opened 2019-03-31
[WIP]  Non-consecutive lists in update requests

ob-stripe:
r? @brandur-stripe @mickjermsurawong-stripe @remi-stripe 

A sort of silly experiment to add support for updating only certain elements in lists, e.g. updating only certain subscription items in a subscription update request.

This PR adds a `NonConsecutiveList<T>` generic class, that behaves just like a normal `List<T>` except that new values may be inserted at any index. This allows a user to do something like this:
```c#
var updatedItems = new NonConsecutiveList<SubscriptionItemUpdateOption>();
updatedItems[1] = new SubscriptionItemUpdateOption { Id = "si_123", ClearUsage = true };
subscriptionService.Update("sub_123", new SubscriptionUpdateOptions { Items = updatedItems });
```
which would send the following form-encoded body to Stripe:
```
items[1][id]=si_123&items[1][clear_usage]=true
```
thus letting the user update the second subscription item without affecting the first.

Tests don't pass, but I'm fairly sure the request is encoded properly and it's stripe-mock that doesn't accept the non-zero starting list because `items` is typed as `array` in the OpenAPI spec. It should ideally be typed as an integer-indexed dictionary, but I'm not sure that we have a way of representing that in the spec.

Even if we can make it work, I'm not certain we should move forward with this because a "non-consecutive list" is a bit of a weird/tricky concept. With a normal list, you can iterate like this:
```c#
for (var i = 0; i < list.Count; i++) {
  Console.WriteLine(list[i]);
}
```

This wouldn't work with a non-consecutive list as it as `Count` returns the maximum index but there may be gaps between `0` and `Count` and `list[i]` would raise an exception when trying to access an index that doesn't exist.

Using `foreach` would work fine though, and you can also use the `Dictionary` property to get a read-only dictionary containing the existing indices and their respective values (this is what is used internally to encode the parameters).

Anyway, opening this PR to get your thoughts and feedback.

### DIFF
--- a/src/Stripe.net/Entities/NonConsecutiveList.cs
+++ b/src/Stripe.net/Entities/NonConsecutiveList.cs
@@ -1,0 +1,252 @@
+namespace Stripe
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Collections.Immutable;
+    using System.Linq;
+
+    /// <summary>
+    /// <para>
+    /// <see cref="NonConsecutiveList{T}" /> behaves just like a normal <see cref="List{T}"/>,
+    /// except that the indices may be non-consecutive. Use <see cref="Insert" /> or
+    /// <see cref="this[int]" /> to insert items at any index.
+    /// </para>
+    /// <para>
+    /// This is useful for update requests, when an existing resource contains a list of elements
+    /// and you only want to update certain elements.
+    /// </para>
+    /// </summary>
+    /// <typeparam name="T">The type of elements in the list.</typeparam>
+    public class NonConsecutiveList<T> : INonConsecutiveList<T>
+    {
+        /// <summary>Store for the list indices and values.</summary>
+        private Dictionary<int, T> contents = new Dictionary<int, T>();
+
+        /// <summary>
+        /// Value of the next index when adding an element in the list without an explicit index.
+        /// </summary>
+        private int nextIndex;
+
+        /// <summary>
+        /// Gets the number of elements contained in the <see cref="NonConsecutiveList{T}" />.
+        /// </summary>
+        /// <value>
+        /// The number of elements contained in the <see cref="NonConsecutiveList{T}" />.
+        /// </value>
+        /// <remarks>
+        /// Because the indices may be non-consecutive, this value may be less than the value of the
+        /// max index. Do not use this value to iterate over the contents of the list; instead, use
+        /// <code>foreach</code> or iterate over the <see cref="Dictionary" /> containing the
+        /// indices.
+        /// </remarks>
+        public int Count => this.contents.Count;
+
+        /// <summary>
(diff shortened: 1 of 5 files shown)


remi-stripe:
@ob-stripe Just flagging that SubscriptionItems is likely the wrong example for your PR. Since the order here does not matter and you don't update a specific index in a list of items. When you do the update, you affect a given item by passing `{id: "si_123", deleted: true}` for example and it's not based on the index. I would even assume that the API errors out if you send an index 1 and not an index 0, at least for that endpoint.

I guess another example that would have been a fit was additional owners but it also was removed.

mickjermsurawong-stripe:
Nice! I didn't think of this inconvenience that users have to copy the exact list element in order to do  element-specific update.

I think there could be a problem with backend though. At least from looking at the test now in `ArrayParam`, it is expected to receive sequential numerical key. 
But supposed we can convert this into backend representation like `[nil, <item_object>, nil]`, it would be quite inconvenient to downstream consumer having to filter this list. So in that case maybe the actual map param with integer key might be a better representation for the param, which is also closer to what's implemented in this PR. 

But to avoid this polymorphic param of map and array, perhaps having a different param like `items_element_update` taking the integer-keyed map might be better? But having this additional param seems like the product decision too. Like @remi-stripe said, maybe many of our endpoints does actually have array update that is not index-sensitive (but ID-sensitive) instead

brandur-stripe:
Haha, wow @ob-stripe, this is slick!! Nice.

So this comment has nothing to do with your PR specifically, but: I think in general encouraging intenger-indexing on arrays in the API is a bad pattern because those indexes are not guaranteed to be stable. So for example, if you get back three elements, and want to update the element at index [1], you could transmit the update back as `[1]` only to realize that another API client has updated the array and the object in `[1]` is no longer the same object you thought it was, and so you clobber an entire object and end up with a duplicate!

In general I'd really like to see us move toward array updates that are indexed by an ID like `si_123` so your update is guaranteed to target the right object. However, as long as we have lots of index updates though, I guess it seems okay to offer some sugar to help do them.

ob-stripe:
@remi-stripe Oops, yeah, obviously subscription items were not the right target for this. I originally started working on this to make it easier to deal with `additional_owners`, but that's no longer an issue in recent API versions.

As far as I can tell, these are the only "list of hashes" parameters (i.e. lists of nested objects without explicit IDs) in update requests in the current API version:
- https://stripe.com/docs/api/customers/update#update_customer-invoice_settings-custom_fields
- https://stripe.com/docs/api/invoices/update#update_invoice-custom_fields

I'll close this PR for now. It was an interesting experiment, but I don't think the added complexity and maintenance burden is worth it. As @brandur-stripe mentioned, we should try not to add any more of these parameters in the API.

PR #1575, opened 2019-04-10
Fix single-line <code> tags in XML comments

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

For inline code, we should use [`<c>`](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/code-inline), not [`<code>`](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/code).


src/Stripe.net/Entities/Accounts/AccountRequirements.cs
```diff
@@ -18,7 +18,7 @@ public class AccountRequirements : StripeEntity
 
         /// <summary>
         /// The fields that need to be collected to keep the account enabled. If not collected by
-        /// the <code>current_deadline</code>, these fields are moved to <code>past_due</code> and
+        /// the <c>current_deadline</c>, these fields are moved to <c>past_due</c> and
         /// the account is disabled.
```
remi-stripe: The `the` should be on the previous line now. This is unacceptable

ob-stripe: Since it doesn't change the actual formatting of the comment, I think we're fine, but feel free to reformat manually yourself 😛

PR #1589, opened 2019-04-19
Switch from static strings to const strings

Franklin89:


Franklin89:
Unsure why this broke the build :-( something with the `coveralls` token :-(

ob-stripe:
Yes, PRs from forked repositories will result in failed builds because of the secure environment variable containing the coveralls token. It's a known issue (technically a feature to protect the secure variable).

I'm going to close this PR for now, since I don't think we will reach a decision on whether to enact this change or not right away. There are also a few other files that would need to be modified. In any case, recreating the PR should be quite trivial since it's basically just a big search and replace.

Regardless, thanks for the contribution! :)

PR #1593, opened 2019-04-25
Add a bunch of missing event types

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Constants/Events.cs
+++ b/src/Stripe.net/Constants/Events.cs
@@ -10,6 +10,11 @@ namespace Stripe
         public static string AccountUpdated => "account.updated";
 
         /// <summary>
+        /// Occurs whenever a user authorizes an application. Sent to the related application only.
+        /// </summary>
+        public static string AccountApplicationAuthorized => "account.application.authorized";
+
+        /// <summary>
         /// Occurs whenever a user deauthorizes an application. Sent to the related application only.
         /// </summary>
         public static string AccountApplicationDeauthorized => "account.application.deauthorized";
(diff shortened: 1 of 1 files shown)


src/Stripe.net/Constants/Events.cs
```diff
@@ -324,6 +379,16 @@ public static class Events
...
 
+        /// <summary>
+        /// Occurs whenever an issuing settlement is created.
+        /// </summary>
+        public static string IssuingSettlementCreated => "issuing_settlement.created";
+
+        /// <summary>
+        /// Occurs whenever an issuing settlement is updated.
+        /// </summary>
+        public static string IssuingSettlementUpdated => "issuing_settlement.updated";
```
remi-stripe: Minor but it's weird to add those when we don't support this resource today

src/Stripe.net/Constants/Events.cs
```diff
@@ -504,11 +639,47 @@ public static class Events
...
+
         /// <summary>
         /// Occurs whenever a new transfer is created.
         /// </summary>
         public static string TransferCreated => "transfer.created";
 
+        /// <summary>
+        /// Occurs whenever a transfer failed.
+        /// </summary>
+        public static string TransferFailed => "transfer.failed";
```
remi-stripe: This is incorrect, it's a legacy event and should not be here since we're pinned to the most recent API version

src/Stripe.net/Constants/Events.cs
```diff
@@ -504,11 +639,47 @@ public static class Events
...
+
+        /// <summary>
+        /// Occurs after a transfer is paid. For Instant Payouts, the event will be sent on the next
+        /// business day, although the funds should be received well beforehand.
+        /// </summary>
+        public static string TransferPaid => "transfer.paid";
```
remi-stripe: same as above

ob-stripe:
ptal @remi-stripe

PR #1601, opened 2019-05-01
Add support for replacing Issuing Card

remi-stripe:
r? @mickjermsurawong-stripe 
cc @ob-stripe just in case (expandable always weird me out)
cc @stripe/api-libraries


mickjermsurawong-stripe:
LGTM
Just a minorrr flag that his `Card` class has an expandable field with reference to itself, but I assume deserialization logic should be fine.

mickjermsurawong-stripe:
thank you for the test!
LGTM

PR #1603, opened 2019-05-02
Rename StripeConnectAccountId to StripeAccount

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Renames `StripeConnectAccountId` to `StripeAccount` in `RequestOptions`. This is consistent with the header's name and our other libraries.

The old `StripeConnectAccountId` property is kept around as an alias for now, to avoid making this a breaking change.


ob-stripe:
> Approving but make sure to update the readme too https://github.com/stripe/stripe-dotnet#request-options

Yup! I'm already using the new syntax in #1602 :)

Thanks!

PR #1608, opened 2019-05-05
Add missing PaymentMethod and Sigma events

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries


remi-stripe:
Correct they are gated but all new accounts have access to those!

PR #1612, opened 2019-05-09
Add support for the Capability resource and APIs

remi-stripe:
cc @stripe/api-libraries


src/Stripe.net/Services/Capabilities/CapabilityListOptions.cs
```diff
@@ -0,0 +1,8 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+
+    public class CapabilityListOptions : ListOptions
```
remi-stripe: Technically this endpoint is special and does not support Limit/pagination. But I'd rather stay consistent since this is "temporary"

ob-stripe:
Released as 25.18.0.

PR #1622, opened 2019-05-21
Remove StripeConfiguration.EnableTelemetry flag

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Remove the `StripeConfiguration.EnableTelemetry` flag. Telemetry is now always enabled. There's really no good reason to have a flag for this: request telemetry carries no PII and the performance impact should be negligible.

This isn't a breaking change because the flag was added earlier in the integration branch for the unreleased future major version.


brandur-stripe:
> This isn't a breaking change because the flag was added earlier in the integration branch for the unreleased future major version.

Phew!

LGTM.

PR #1631, opened 2019-05-27
Refactor StripeClient setup in tests

ob-stripe:
cc @stripe/api-libraries 

Changes the way `StripeClient` is set up in tests.

To clarify what's happening here:
- `MockHttpClientFixture` is only created once for the entire test suite (because it's set up as a ["collection fixture"](https://xunit.net/docs/shared-context#collection-fixture) [here](https://github.com/stripe/stripe-dotnet/blob/136c8b39afd8bf97e770a86eea8dbdce8ab5aedc/src/StripeTests/StripeMockTestCollection.cs#L7))
- `BaseStripeTest` is the parent class for all tests, so its constructor is called for each test class

The problem with the previous implementation is that since we set up the global `StripeClient` (via `StripeConfiguration.StripeClient`) in `MockHttpClientFixture`, the client was applied for all tests (even those that don't need mocking capabilities). More importantly, if a test needs to modify `StripeConfiguration.StripeClient` for its own purposes, it needs to save the original client and restore it afterwards because other test classes might need the `StripeClient` set up by `MockHttpClientFixture`.

With this change, we create a new `StripeClient` instance for each test class, and use the mock client from `MockHttpClientFixture` if needed. This provides better test isolation, and enables some followup changes I want to make to `StripeClient`.

### DIFF
--- a/src/StripeTests/BaseStripeTest.cs
+++ b/src/StripeTests/BaseStripeTest.cs
@@ -5,35 +5,85 @@ namespace StripeTests
     using System.Net.Http;
     using System.Reflection;
     using System.Text;
+    using Stripe;
     using Xunit;
 
     [Collection("stripe-mock tests")]
     public class BaseStripeTest
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BaseStripeTest"/> with no fixtures.
+        /// </summary>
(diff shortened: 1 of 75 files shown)


ob-stripe:
Thanks Brandur!

PR #1633, opened 2019-05-29
Add support for setting API key and client ID in StripeClient

ob-stripe:
cc @stripe/api-libraries 

This PR adds support for setting the API key and the client ID directly in the `StripeClient` instance instead of relying on the global `StripeConfiguration` object.

Setting these values via `StripeConfiguration` is still supported. Whenever `StripeConfiguration.ApiKey` or `StripeConfiguration.ClientId` is set, `StripeConfiguration.StripeClient` is reset to `null` so that on the next request, a new `StripeClient` instance will be created with the updated values.

I will update the documentation to clarify that when setting a custom `StripeClient` instance directly via `StripeConfiguration.StripeClient`, you must 1. pass the API key (and optionally the client ID) in the `StripeClient`'s constructor and 2. _not_ set `StripeConfiguration.ApiKey` or `StripeConfiguration.ClientId` afterwards, otherwise your custom `StripeClient` will be lost.

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/StripeClient.cs
+++ b/src/Stripe.net/Infrastructure/Public/StripeClient.cs
@@ -1,10 +1,12 @@
 namespace Stripe
 {
+    using System;
     using System.Net;
     using System.Net.Http;
     using System.Threading;
     using System.Threading.Tasks;
     using Newtonsoft.Json.Linq;
+    using Stripe.Infrastructure;
 
     /// <summary>
     /// A Stripe client, used to issue requests to Stripe's API and deserialize responses.
@@ -12,12 +14,38 @@ namespace Stripe
     public class StripeClient : IStripeClient
     {
         /// <summary>Initializes a new instance of the <see cref="StripeClient"/> class.</summary>
+        /// <param name="apiKey">The API key used by the client to make requests.</param>
+        /// <param name="clientId">The client ID used by the client in OAuth requests.</param>
         /// <param name="httpClient">
         /// The <see cref="IHttpClient"/> client to use. If <c>null</c>, an HTTP client will be
         /// created with default parameters.
         /// </param>
-        public StripeClient(IHttpClient httpClient = null)
+        /// <exception cref="ArgumentNullException">if <c>apiKey</c> is <c>null</c></exception>
+        /// <exception cref="ArgumentException">
+        /// if <c>apiKey</c> is empty or contains whitespace
+        /// </exception>
+        public StripeClient(
+            string apiKey,
+            string clientId = null,
+            IHttpClient httpClient = null)
         {
+            if (apiKey == null)
+            {
+                throw new ArgumentNullException(nameof(apiKey));
+            }
+
+            if (apiKey.Length == 0)
+            {
+                throw new ArgumentException("API key cannot be the empty string.", nameof(apiKey));
+            }
+
+            if (StringUtils.ContainsWhitespace(apiKey))
+            {
+                throw new ArgumentException("API key cannot contain whitespace.", nameof(apiKey));
+            }
+
+            this.ApiKey = apiKey;
+            this.ClientId = clientId;
             this.HttpClient = httpClient ?? BuildDefaultHttpClient();
         }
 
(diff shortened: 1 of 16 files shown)


src/Stripe.net/Infrastructure/Public/StripeClient.cs
```diff
@@ -1,23 +1,51 @@
...
+            if (apiKey.Length == 0)
+            {
+                throw new ArgumentException("API key cannot be the empty string.", nameof(apiKey));
+            }
+
+            if (StringUtils.ContainsWhitespace(apiKey))
+            {
+                throw new ArgumentException("API key cannot contain whitespace.", nameof(apiKey));
+            }
+
```
remi-stripe: Should this logic be put into an API Key validator or something since I'm assuming you have the same in the config and potentially the same on the request options?

src/Stripe.net/Infrastructure/Public/StripeClient.cs
```diff
@@ -1,23 +1,51 @@
...
+    using Stripe.Infrastructure;
 
     /// <summary>
     /// A Stripe client, used to issue requests to Stripe's API and deserialize responses.
     /// </summary>
     public class StripeClient : IStripeClient
     {
         /// <summary>Initializes a new instance of the <see cref="StripeClient"/> class.</summary>
+        /// <param name="apiKey">The API key used by the client to make requests.</param>
+        /// <param name="clientId">The client ID used by the client in OAuth requests.</param>
```
remi-stripe: I don't fully grasp this `clientId`. Reading the rest of the PR it seems to be the ca_123. But naming that variable `clientId` in the StripeClient class is quite confusing (at least to me). I kind of assumed you meant some unique client identifiers for that instance for some reason until I read more and it started being clear it did not mean that.

Food for thought as I don't feel strongly but wanted to mention this.

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
```diff
@@ -49,7 +49,15 @@ public static string ApiKey
                 return apiKey;
             }
 
-            set => apiKey = value;
+            set
+            {
+                if (value != apiKey)
+                {
+                    StripeClient = null;
```
remi-stripe: Could that affect ongoing requests that rely on this variable? Do we locally cache those or would that suddenly destroy the previously set client?

src/Stripe.net/Infrastructure/Public/StripeClient.cs
ob-stripe: We don't validate API keys in request options right now. We could, but the logic would be slightly different since `null` is allowed there.

The validation logic is essentially the same for the `StripeConfiguration. BuildDefaultStripeClient()` helper and the `StripeClient` constructor, but the exceptions that are thrown are different. The former throws `StripeException` with a long/helpful error message, the latter throws `ArgumentNullException` or `ArgumentException` with terse messages (which is more idiomatic, and I think fine since only "advanced" users creating their own client would see those).

src/Stripe.net/Infrastructure/Public/StripeClient.cs
ob-stripe: Yeah, the word "client" is a bit overloaded here. I think the documentation clarifies that this is the OAuth client ID, but I could rename the argument to `oauthClientId` or something similar to help disambiguate.

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
ob-stripe: It should be fine:
1. Set `StripeConfiguration.ApiKey = "sk_foo"` --> `StripeConfiguration.StripeClient` is reset to `null`
2. Create a service and send a request. The service will grab the client from `StripeConfiguration.StripeClient`, and the getter will automatically build a new client with `StripeConfiguration.ApiKey`
3. Set `StripeConfiguration.ApiKey = "sk_bar"` --> `StripeConfiguration.StripeClient` is reset to `null`
4. Send another request. The service will grab the client from `StripeConfiguration.StripeClient`, and the getter will automatically build a new client with `StripeConfiguration.ApiKey`

Even in a multi-threaded environment, it should still be fine: the client is not immediately destroyed when `StripeConfiguration.StripeClient` is reset to `null`. If any service has an ongoing request, it will have a reference to the `StripeClient` instance, and the client won't be GC'd until the service is done with the request.

(Really though the proper answer should be "set the global API key _once_ before sending any requests and don't change it afterwards", and we might want to enforce that in the future.)

PR #1637, opened 2019-05-30
Add support for `location` on `ConnectionToken`

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/Terminal/ConnectionTokens/ConnectionToken.cs
+++ b/src/Stripe.net/Entities/Terminal/ConnectionTokens/ConnectionToken.cs
@@ -10,6 +10,9 @@ namespace Stripe.Terminal
         [JsonProperty("object")]
         public string Object { get; set; }
 
+        [JsonProperty("location")]
+        public string LocationId { get; set; }
+
         [JsonProperty("secret")]
         public string Secret { get; set; }
     }
--- a/src/Stripe.net/Services/Terminal/ConnectionTokens/ConnectionTokenCreateOptions.cs
+++ b/src/Stripe.net/Services/Terminal/ConnectionTokens/ConnectionTokenCreateOptions.cs
@@ -5,6 +5,9 @@ namespace Stripe.Terminal
 
     public class ConnectionTokenCreateOptions : BaseOptions
     {
+        [JsonProperty("location")]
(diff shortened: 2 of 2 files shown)


src/Stripe.net/Entities/Terminal/ConnectionTokens/ConnectionToken.cs
```diff
@@ -10,6 +10,9 @@ public class ConnectionToken : StripeEntity, IHasObject
         [JsonProperty("object")]
         public string Object { get; set; }
 
+        [JsonProperty("location")]
+        public string Locaation { get; set; }
```
ob-stripe:
```suggestion
        public string Location { get; set; }
```

remi-stripe:
duh my bad, PTAL @ob-stripe

src/Stripe.net/Entities/Terminal/ConnectionTokens/ConnectionToken.cs
```diff
@@ -10,6 +10,9 @@ public class ConnectionToken : StripeEntity, IHasObject
         [JsonProperty("object")]
         public string Object { get; set; }
 
+        [JsonProperty("location")]
+        public string Location { get; set; }
```
ob-stripe: This is an ID, correct? Should we rename it to `LocationId`?

src/Stripe.net/Services/Terminal/ConnectionTokens/ConnectionTokenCreateOptions.cs
```diff
@@ -5,6 +5,9 @@ namespace Stripe.Terminal
 
     public class ConnectionTokenCreateOptions : BaseOptions
     {
+        [JsonProperty("location")]
+        public string Location { get; set; }
```
ob-stripe: And same here.

src/Stripe.net/Entities/Terminal/ConnectionTokens/ConnectionToken.cs
remi-stripe: ah yes you're right. I forgot because it's not expandable and likely won't be, but consistency is important!

remi-stripe:
@ob-stripe PTAL

PR #1639, opened 2019-05-31
Make client in services readonly

ob-stripe:
cc @stripe/api-libraries 

Make the `IStripeClient` property in `Service` readonly, i.e. it can only be set when the service is created.

Also remove the ability to pass an API key when creating a service. Users who need this can now pass a client instead, i.e. they'd replace:
```c#
var fooService = new FooService(apiKey);
```
with:
```c#
var fooService = new FooService(new StripeClient(apiKey));
```

### DIFF
--- a/src/Stripe.net/Services/_base/Service.cs
+++ b/src/Stripe.net/Services/_base/Service.cs
@@ -10,34 +10,44 @@ namespace Stripe
     using System.Threading.Tasks;
     using Stripe.Infrastructure;
 
+    /// <summary>Abstract base class for all services.</summary>
+    /// <typeparam name="EntityReturned">
+    /// The type of <see cref="IStripeEntity"/> that this service returns.
+    /// </typeparam>
     public abstract class Service<EntityReturned>
         where EntityReturned : IStripeEntity
     {
         private IStripeClient client;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Service{EntityReturned}"/> class.
+        /// </summary>
         protected Service()
         {
         }
 
-        protected Service(string apiKey)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Service{EntityReturned}"/> class with a
+        /// custom <see cref="IStripeClient"/>.
+        /// </summary>
+        /// <param name="client">The client used by the service to send requests.</param>
+        protected Service(IStripeClient client)
         {
-            this.ApiKey = apiKey;
+            this.client = client;
         }
-
-        public string ApiKey { get; set; }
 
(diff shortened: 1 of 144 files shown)


src/Stripe.net/Services/_base/Service.cs
```diff
@@ -10,34 +10,44 @@ namespace Stripe
...
         {
         }
 
-        protected Service(string apiKey)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Service{EntityReturned}"/> class with a
+        /// custom <see cref="IStripeClient"/>.
+        /// </summary>
+        /// <param name="client">The client used by the service to send requests.</param>
+        protected Service(IStripeClient client)
```
remi-stripe: why do we have 2 constructors instead of just one with a default value of null? (mostly for my own education)

src/Stripe.net/Services/_base/Service.cs
```diff
@@ -290,11 +300,6 @@ protected RequestOptions SetupRequestOptions(RequestOptions requestOptions)
                 requestOptions = new RequestOptions();
             }
 
-            if (string.IsNullOrEmpty(requestOptions.ApiKey) && !string.IsNullOrEmpty(this.ApiKey))
```
remi-stripe: Can you clarify why this disappeared? Is it because the StripeClient itself will do the validation?

ob-stripe: So previously, there were 4 (!) different ways to set up an API key. In order of decreasing priority:

1. `RequestOptions.ApiKey`
2. `Service.ApiKey`
3. `StripeClient.ApiKey`
4. `StripeConfiguration.ApiKey`

The removed lines implemented #2. Basically "if `RequestOptions.ApiKey` is not already set (i.e. the user did not set an API key via #1) and `Service.ApiKey` is set, then set `RequestOptions.ApiKey` to `Service.ApiKey` (so that down the road the client will use this as the API key for the request).

After this PR, #2 is gone.

brandur-stripe:
Good question from Remi above, but looks good to me!

src/Stripe.net/Services/_base/Service.cs
ob-stripe: Good question! I think this dates back to Jayme's original design... It's possible that optional arguments weren't even a thing in C# when he started the library.

I think we could plausibly merge both constructors into one, though the current implementation works and I'm unsure if there are tricky differences between method overloading and optional arguments, so I'd rather leave it as is for now.

ob-stripe:
Thanks both!

PR #1644, opened 2019-06-06
Add options classes for Get/GetAsync methods

### LINKED ISSUE #1630: PaymentIntent is not expandable when creating a subscription
Hi,
I'm currently following the billing tutorial [here](https://stripe.com/docs/billing/subscriptions/payment) and when creating a subscription using the CreateAsync method : `Subscription oSubscription = await _oSubscriptionService.CreateAsync(oOptions);` the subscription does not contain a `PaymentIntent ` in `oSubscription.LatestInvoice.PaymentIntent`. I have set the property `_oSubscriptionService.ExpandLatestInvoice = true;` but it stops there : I get the latest invoice without the PaymentIntent.  Currently the only solution I found was retrieving the PaymentIntent via the PaymentIntentService with the PaymentIntentId provided in the LatestInvoice.

Is there a way to get it directly as shown in the guide ? Am I missing something ?

Thanks,

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add options classes for all `Get` and `GetAsync` methods.

Most of these simply derive from `BaseOptions` and don't have any properties of their own, but this allows users to expand nested fields in retrieve requests.

Fixes #1630.

### DIFF
--- a/src/Stripe.net/Services/Account/AccountGetOptions.cs
+++ b/src/Stripe.net/Services/Account/AccountGetOptions.cs
@@ -1,0 +1,6 @@
+namespace Stripe
+{
+    public class AccountGetOptions : BaseOptions
+    {
+    }
+}
--- a/src/Stripe.net/Services/Account/AccountService.cs
+++ b/src/Stripe.net/Services/Account/AccountService.cs
@@ -9,7 +9,7 @@ namespace Stripe
         ICreatable<Account, AccountCreateOptions>,
         IDeletable<Account>,
         IListable<Account, AccountListOptions>,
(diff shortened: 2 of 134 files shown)


ob-stripe:
> Are you sure that you can still call `Get(id, request_options);`? If not, isn't it worth adding an overload just for this?

No, this will fail at compile time. Users will need to update their code to either pass `null` for the options: `Get(id, null, requestOptions)` or specify the argument explicitly: `Get(id, requestOptions: requestOptions);`.

I don't think it's worth adding an overload for this.

PR #1655, opened 2019-06-10
Add setters for expandable field accessors

### LINKED ISSUE #1651: Recommended strategy for stubbing entities in unit tests in version 27.0.0
OS version: Windows 10
.NET Core version: 2.2
Stripe.net version: 27.0.0

Before version 27.0.0 it was trivial to create unit tests with mocking/stubbing behavior, e.g.

```
[TestMethod]
public async Task Example()
{
    Moq<ISomeServiceThatMakesStripeApiRequest>().Setup(m => m.GetCustomer("cus_x")).ReturnsAsync(new Customer{DefaultSourceId = "src_x"});
    var result = await _target.SomeWrapperMethodAsync("cus_x");
    ...
}
```

In version 27.0.0 the method above no longer compiles because many of the core resources (e.g. Customer) now contain properties that are readonly.

What's the recommended approach for creating stubbed resources (e.g. a `Customer` with `DefaultSourceId` set to a test value) now? Of course we could create our own wrapper objects for all your entities, but this would be a maintenance pain.

We're concerned that you guys have neglected unit-test considerations with version 27.0.0.

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

In #1481, I changed the way expandable fields are handled and removed setters for the ID and expanded properties matching these fields.

This turned out to be inconvenient for users using these setters in unit tests, so this PR adds setters back.

Fixes #1651.

### DIFF
--- a/src/Stripe.net/Entities/Accounts/AccountSettingsBranding.cs
+++ b/src/Stripe.net/Entities/Accounts/AccountSettingsBranding.cs
@@ -13,13 +13,21 @@ namespace Stripe
         /// <para>Expandable.</para>
         /// </summary>
         [JsonIgnore]
-        public string IconId => this.InternalIcon.Id;
+        public string IconId
+        {
+            get => this.InternalIcon.Id;
+            set => this.InternalIcon.Id = value;
+        }
 
         /// <summary>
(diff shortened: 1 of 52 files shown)


ob-stripe:
Released as 27.1.0.

PR #1656, opened 2019-06-10
Fix json property name for id_number_provided

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Released as 27.1.1.

PR #1658, opened 2019-06-10
Add `ConfigureAwait(false)` to all async calls

### LINKED ISSUE #1653: List PaymentMethod is in dead lock and always timeout in version 27.0.0
var service = new PaymentMethodService();
var options = new PaymentMethodListOptions
{
CustomerId = customerId,
Limit = 1,
Type = "card"
};
var paymentMethods = service.List(options);

This function is in dead lock and always timeout in version 27.0.0.
Changed back to 26.1.0 then works fine.

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Some async calls were missing a `ConfigureAwait(false)`, which can cause a deadlock in applications where the main thread is busy with something else (e.g. serving HTTP requests).

I've verified that all the async calls in the library now have `ConfigureAwait(false)`, and verified that the problem described in #1653 no longer occurs with this patch.

I'll try thinking of a way to write a test for this to avoid regressions in the future, but I think it might be difficult :/

Fixes #1653.

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/StripeClient.cs
+++ b/src/Stripe.net/Infrastructure/Public/StripeClient.cs
@@ -116,7 +116,8 @@ namespace Stripe
         {
             var request = new StripeRequest(this, method, path, options, requestOptions);
 
-            var response = await this.HttpClient.MakeRequestAsync(request, cancellationToken);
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released as 27.1.2.

PR #1660, opened 2019-06-11
Fix possible NRE in expandable field accessors

### LINKED ISSUE #1659: 27.1 setters are back, but throw NullReferenceException
With 27.1.0 setters are back, but trying to set a value results in NullReferenceException being throws. As an example trying to set both Customer and CustomerId of Charge throws.

Rather bizarre for payment related project to push untested changes in this manner. They are obviously untested as literally any test will catch it immediately.

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

A few changes to avoid possible `NullReferenceException`s when using accessors for expandable fields:
- add safe navigation operator to all getters
- use new helper methods for all setters:
    - when setting the ID, `SetExpandableFieldId` will create the `ExpandableField<T>` instance if it doesn't exist already, and resets the expanded object if the IDs don't match
    - when setting the expanded object, `SetExpandableFieldObject` will create the `ExpandableField<T>` instance if it doesn't exist already.

I've written some generic tests for `StripeEntity` to test the helper methods, and also a few tests for `Customer.DefaultSource` so that we test at least one "real" expandable field. It doesn't seem realistic to write tests for every single expandable field accessor, so I guess we'll just have to be careful with copy/paste when adding new expandable fields.

(I have a vague idea for simplifying the expandable field accessor mess, but I'll keep that for a future major version.)

(Also, I think the entity classes should not have public setters at all and should be implemented as immutable [value objects](https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/microservice-ddd-cqrs-patterns/implement-value-objects), but that would require writing very long constructor methods, so not going to do that until we can codegen .NET.)

Fixes #1659.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,6 +3,7 @@
   "files.insertFinalNewline": true,
   "files.trimFinalNewlines": true,
   "files.trimTrailingWhitespace": true,
+  "window.title": "${rootName}${separator}${activeEditorMedium}",
   "[csharp]": {
     "editor.detectIndentation": false,
     "editor.insertSpaces": true,
--- a/src/Stripe.net/Entities/Accounts/AccountSettingsBranding.cs
+++ b/src/Stripe.net/Entities/Accounts/AccountSettingsBranding.cs
@@ -15,8 +15,8 @@ namespace Stripe
         [JsonIgnore]
         public string IconId
         {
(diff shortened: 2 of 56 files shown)


ob-stripe:
Released as 27.1.3.

PR #1661, opened 2019-06-12
Fix handling of null values with AnyOf

ob-stripe:
r? @remi-stripe @fred-stripe 
cc @stripe/api-libraries 

Fix handling of `null` values with `AnyOf` in the encoder. `null` values are now ignored, like they would be for a non-`AnyOf` property.

### DIFF
--- a/src/Stripe.net/Infrastructure/FormEncoding/FormEncoder.cs
+++ b/src/Stripe.net/Infrastructure/FormEncoding/FormEncoder.cs
@@ -101,7 +101,7 @@ namespace Stripe.Infrastructure.FormEncoding
                     break;
 
                 case IAnyOf anyOf:
-                    flatParams = FlattenParamsValue(anyOf.Value, keyPrefix);
+                    flatParams = FlattenParamsAnyOf(anyOf, keyPrefix);
                     break;
 
                 case INestedOptions options:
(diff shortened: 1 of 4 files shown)


ob-stripe:
> Is it worth adding a test for Source which is critical in this library so that we don't break it again? the created one is a bit of an edge-case and something we wouldn't care to remove the test for.

Yep, good idea. I added a new `ChargeCreateOptionsTest` test class to test the various cases for `ChargeCreateOptions.Source`.

> Also, is there a world in which you wanted to unset something and now it's broken (like how AdditionalOwners could be unset before)?

I think we're good:
- list fields can be unset by setting the value to an empty list
- metadata keys can be unset by setting the key to either `null` or an empty string (in this case, we do encode `null` as an empty string since we know for sure that the `null` value was explicitly inserted in the dictionary by the user)

Both cases are already covered in `FormEncoderTest`.

Can you think of any other cases for unsetting values?

ob-stripe:
ptal @remi-stripe

remi-stripe:
> Can you think of any other cases for unsetting values?

There's this whole "to unset a hash, send an empty string" like to empty `Charge.shipping` you do `shipping=""`. Does that still work? Or would it be introduced as a separate concept like we did the `Empty` in stripe-java?

ob-stripe:
> There's this whole "to unset a hash, send an empty string" like to empty Charge.shipping you do shipping="". Does that still work? Or would it be introduced as a separate concept like we did the Empty in stripe-java?

AFAICT that was never supported in stripe-dotnet. It would be easy to implement something similar to stripe-java's `EmptyParam.EMPTY` with `AnyOf` though, but let's do that separately since it would be a new feature.

ob-stripe:
Released as 27.1.3.

PR #1664, opened 2019-06-14
Add support for `merchant_amount` `merchant_currency`  to Issuing Transaction

cjavilla-stripe:
?r @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 [Oo]bj/
 *.nupkg
 
-# user 
+# user
 *.user
 *.userprefs
 .vs
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ go get -u github.com/stripe/stripe-mock
 stripe-mock
 ```
 
-Run all tests:
+Run all tests from the `src/StripeTests` directory:
 
 ```sh
 dotnet test
--- a/src/Stripe.net/Entities/Issuing/Transactions/Transaction.cs
+++ b/src/Stripe.net/Entities/Issuing/Transactions/Transaction.cs
@@ -126,6 +126,12 @@ namespace Stripe.Issuing
         [JsonProperty("livemode")]
         public bool Livemode { get; set; }
 
+        [JsonProperty("merchant_amount")]
+        public long MerchantAmount { get; set; }
+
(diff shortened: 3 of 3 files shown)


remi-stripe:
LGTM but assigning to @ob-stripe since you changed other files

.gitignore
```diff
@@ -12,3 +12,4 @@ project.lock.json
 
 .DS_Store
 .idea/
+.vscode/*
```
ob-stripe: The VSCode configuration is deliberately committed to the repo, so let's not do this.

README.md
```diff
@@ -115,7 +115,7 @@ stripe-mock
 Run all tests:
 
 ```sh
-dotnet test
+dotnet test ./src/StripeTests/StripeTests.csproj
```
ob-stripe: Hm, I guess that's better since it's fair to assume the command would be run from the repo's root, but we should change the other two commands below then.

Also, there's no need to specify the `.csproj` file, just the directory is enough:
```
dotnet test ./src/StripeTests
```

I'm not sure forward slashes would work on Windows though 🤔 Maybe let's keep the commands as is and just add a note saying they should be run from the `src/StripeTests` directory?

.gitignore
cjavilla-stripe: Ah. I missed that there was already a .vscode dir because when I opened vs code, it created two more files and I thought these were the only newly generated:

```
	.vscode/launch.json
	.vscode/tasks.json
```

I can just manually ignore those.

README.md
cjavilla-stripe: Sounds good :)

README.md
```diff
@@ -112,7 +112,7 @@ go get -u github.com/stripe/stripe-mock
 stripe-mock
 ```
 
-Run all tests:
+Run all tests from the src/StripeTests directory:
```
ob-stripe: Nit:
```suggestion
Run all tests from the `src/StripeTests` directory:
```

cjavilla-stripe: :) I almost did this the first time.

ob-stripe:
LGTM, thanks @cjavilla-stripe! :)

PR #1665, opened 2019-06-14
Add support for `submit_type` to Checkout/SessionCreate

cjavilla-stripe:
?r @remi-stripe 
cc @stripe/api-libraries


remi-stripe:
re-assigning to you as we're adding the property on the resource too right now so this PR will need to be updated.

Issue #1670, opened 2019-06-20
Custom Newtonsoft JSON Options cause header error from telemetry


richardlawley:
NET Framework 4.7.1, Stripe.Net 27.3.1, JSON.NET 12.0.2

If you make multiple requests using the same StripeClient, the second request adds a telemetry header containing the last request's latency.  The value of this header is a JSON value, serialised using `JsonConvert.SerializeValue`.

If the calling application has specified custom default options for JsonConvert (specifically referring to the Formatting Option), this value can contain newlines and causes every request after the initial one to fail:

> System.FormatException: 'New-line characters in header values must be followed by a white-space character.'

This was thrown from System.Net.Http.Headers.HttpHeaders.ParseAndAddValue with name = "X-Stripe-Client-Telemetry" and value as below (newlines left encoded to make it obvious):

> "{\r\n  \"last_request_metrics\": {\r\n    \"request_id\": \"req_iXuBVBUsCQUDdF\",\r\n    \"request_duration_ms\": 485\r\n  }\r\n}"

To resolve this, you should create your own JsonConverter instance with your own fixed options, rather than relying on the defaults which a calling application may have changed.

ob-stripe:
Thanks a lot for the detailed report and the fix @richardlawley! Much appreciated.

PR #1671, opened 2019-06-20
Remove formatting from Telemetry

richardlawley:
Remove formatting from Telemetry as this causes an error on second and subsequent requests when custom JSON.NET options are in use which specify Formatting.Indented.

Fixes: #1670


brandur-stripe:
Thanks Richard!

It looks like the build is failing due to a Coveralls key not being available on your branch (it's a secure key so probably not inserted for third party contributors). I'll try to get that fixed separately.

brandur-stripe:
Released as 27.3.2.

PR #1672, opened 2019-06-20
Only run Coveralls if $COVERALLS_REPO_TOKEN is available

brandur:
I noticed that builds from third party contributors apparently fail
because no Coveralls key is available (`$COVERALLS_REPO_TOKEN` is a
secure token and AppVeyor won't put it in all build environments). Here
we still run coverage, but only invoke the Coveralls step if a key is
available for it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -57,7 +57,12 @@ test_script:
 after_test:
   - ps: Write-Host $("`n               RUNNING COVERAGE               `n") -BackgroundColor DarkCyan
   - dotnet test -c Debug -f netcoreapp2.1 src/StripeTests/StripeTests.csproj --no-build /p:CollectCoverage=true /p:CoverletOutputFormat=opencover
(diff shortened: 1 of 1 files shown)


ob-stripe:
🤔 Yeah, it's weird that the command doesn't show up at all in the build log. What you have is probably working, but maybe we should change it to use PowerShell instead? It's what we use in the rest of the file so it would be a little more consistent as well. Something like this should work:

```ps
if (Test-Path env:COVERALLS_REPO_TOKEN) {
    .\tools\csmacnz.Coveralls --opencover -i src/StripeTests/coverage.opencover.xml --useRelativePaths
}
```

brandur-stripe:
Haha, thanks OB!

And yeah, +1 PowerShell. I flipped flopped between Bash and PowerShell originally based on "this is SO simple that maybe I should just do it in Bash" versus "well, everything else is PowerShell already ...".

Issue #1674, opened 2019-06-21
Plan Tiers won't allow null for UpTo


jrstuart42:
when creating a plan with tiers i get an error "Stripe.StripeException : Missing required param: tiers[3][up_to]." i filled in the other necessary information for plan and the tiers look like the code below. UpTo is a nullable long. it should be able to take null as a value considering stripe allows null for it too.

```c#
Plan plan = new Plan()
{
    //other necessary info omitted
    Tiers = new List<PlanTier>()
    {
        new PlanTier()
        {
            UnitAmount = 7000,
            UpTo = 500
        },
        new PlanTier()
        {
            UnitAmount = 6000,
            UpTo = 1000
        },
        new PlanTier()
        {
            UnitAmount = 5000,
            UpTo = null
        }
    }
};
List<PlanTierOptions> planTierOptions = plan.Tiers.Select(pt => new PlanTierOptions()
{
    FlatAmount = pt.FlatAmount,
    UnitAmount = pt.UnitAmount,
    UpTo = pt.UpTo
}).ToList();
PlanCreateOptions createOptions = new PlanCreateOptions()
{
    Active = true,
    Amount = plan.Amount,
    Currency = plan.Currency,
    Id = plan.PlanID,
    Interval = plan.Interval,
    IntervalCount = plan.IntervalCount,
    Nickname = plan.Nickname,
    Product = plan.ProductID,
    Tiers = planTierOptions,
    TiersMode = plan.TiersMode,
    TrialPeriodDays = plan.TrialPeriodDays,
    UsageType = plan.UsageType
};
```

ob-stripe:
@jrstuart42 `up_to` cannot be `null`, it must be either an integer or the `inf` special value: https://stripe.com/docs/billing/subscriptions/tiers.

You'd need to change your code like this:
```c#
Plan plan = new Plan
{
    Tiers = new List<PlanTier>
    {
        new PlanTier
        {
            UnitAmount = 7000,
            UpTo = 500,
        },
        new PlanTier
        {
            UnitAmount = 6000,
            UpTo = 1000,
        },
        new PlanTier
        {
            UnitAmount = 5000,
            UpTo = PlanTierUpTo.Inf,
        }
    }
};
```

Closing as I don't think this is a bug with the library, but feel free to reply.

jrstuart42:
that didn't work. it gave the error "StripeException: Invalid positive integer"

ob-stripe:
[...]

PR #1675, opened 2019-06-21
Better string enums

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Right now we're using regular `enum`s for a few special values, like `PlanTierUpTo.Inf`.

The issue with this approach is that `enum` are really integers, but in our case it's not desirable to have a conversion between `enum` and `int`. This is especially true for `PlanTierUpTo`, because `PlanTierOptions.UpTo` is defined as `AnyOf<long?, PlanTierUpTo?>`. If for some reason the `PlanTierUpTo` is implicitly or explicitly converted to `long`, then the value will be accepted and sent as an int instead of a string in the request, leading to unexpected results (cf. #1674).

This PR shows a possible solution: stop using `enum`s and instead use regular classes with a private constructor and the possible values as static properties. This is technically a breaking change because of the type change, but the impact on users should be very minimal because the syntax doesn't change: you can still use `PlanTierUpTo.Inf` just like before.

### DIFF
--- a/src/Stripe.net/Services/_base/StringEnum.cs
+++ b/src/Stripe.net/Services/_base/StringEnum.cs
@@ -1,0 +1,46 @@
+namespace Stripe
+{
+    /// <summary>
+    /// Abstract base class for string enum parameters.
+    /// </summary>
+    /// <remarks>
+    /// This class is used to define request parameters that can only take certain string values.
+    /// We use it instead of defining a regular <c>enum</c> with
+    /// <see cref="System.Runtime.Serialization.EnumMemberAttribute"/> annotations for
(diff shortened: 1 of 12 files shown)


src/Stripe.net/Services/_base/StringEnum.cs
```diff
@@ -0,0 +1,17 @@
+namespace Stripe
+{
+    public abstract class StringEnum
```
brandur-stripe: You may not have done it because this is still a prospective change, but before merging could you add some documentation on this class that talks a little about why it exists?

ob-stripe:
Added comments, ptal @brandur-stripe

PR #1679, opened 2019-06-24
Add CollectionMethod for Invoices and Subscriptions

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/InvoiceItems/InvoiceItem.cs
+++ b/src/Stripe.net/Entities/InvoiceItems/InvoiceItem.cs
@@ -122,6 +122,13 @@ namespace Stripe
         [JsonProperty("tax_rates")]
         public List<TaxRate> TaxRates { get; set; }
 
+        /// <summary>
+        /// Set to <c>true</c> if we grouped proration items into one,
+        /// <c>false</c> if not.
+        /// </summary>
+        [JsonProperty("unified_proration")]
+        public bool UnifiedProration { get; set; }
+
         [JsonProperty("unit_amount")]
         public long? UnitAmount { get; set; }
     }
--- a/src/Stripe.net/Entities/Invoices/Invoice.cs
+++ b/src/Stripe.net/Entities/Invoices/Invoice.cs
@@ -52,11 +52,10 @@ namespace Stripe
         public bool AutoAdvance { get; set; }
 
         /// <summary>
-        /// One of <see cref="Billing" />. When charging automatically, Stripe will attempt to pay
-        /// this subscription at the end of the cycle using the default source attached to the
-        /// customer. When sending an invoice, Stripe will email your customer an invoice with
-        /// payment instructions.
+        /// This field has been renamed to <c>CollectionMethod</c> and will be removed
+        /// in a future API version.
         /// </summary>
+        [Obsolete("Use CollectionMethod")]
         [JsonProperty("billing")]
         public Billing? Billing { get; set; }
 
@@ -82,6 +81,16 @@ namespace Stripe
         [JsonConverter(typeof(ExpandableFieldConverter<Charge>))]
         internal ExpandableField<Charge> InternalCharge { get; set; }
         #endregion
+
+        /// <summary>
+        /// Either <c>charge_automatically</c>, or <c>send_invoice</c>. When
+        /// charging automatically, Stripe will attempt to pay this invoice
+        /// using the default source attached to the customer. When sending an
+        /// invoice, Stripe will email this invoice to the customer with payment
+        /// instructions. Defaults to <c>charge_automatically</c>.
+        /// </summary>
+        [JsonProperty("collection_method")]
+        public string CollectionMethod { get; set; }
 
         [JsonProperty("created")]
         [JsonConverter(typeof(DateTimeConverter))]
--- a/src/Stripe.net/Entities/Invoices/InvoiceLineItem.cs
+++ b/src/Stripe.net/Entities/Invoices/InvoiceLineItem.cs
@@ -70,5 +70,12 @@ namespace Stripe
 
         [JsonProperty("type")]
         public string Type { get; set; }
+
+        /// <summary>
+        /// Set to <c>true</c> if we grouped proration items into one,
+        /// <c>false</c> if not.
+        /// </summary>
+        [JsonProperty("unified_proration")]
+        public bool UnifiedProration { get; set; }
     }
 }
--- a/src/Stripe.net/Entities/Subscriptions/Subscription.cs
+++ b/src/Stripe.net/Entities/Subscriptions/Subscription.cs
@@ -17,11 +17,10 @@ namespace Stripe
         public decimal? ApplicationFeePercent { get; set; }
 
         /// <summary>
-        /// One of <see cref="Billing" />. When charging automatically, Stripe will attempt to pay
-        /// this subscription at the end of the cycle using the default source attached to the
-        /// customer. When sending an invoice, Stripe will email your customer an invoice with
-        /// payment instructions.
+        /// This field has been renamed to <c>CollectionMethod</c> and will be removed
+        /// in a future API version.
         /// </summary>
+        [Obsolete("Use CollectionMethod")]
         [JsonProperty("billing")]
         public Billing? Billing { get; set; }
 
@@ -49,6 +48,16 @@ namespace Stripe
         [JsonProperty("canceled_at")]
         [JsonConverter(typeof(DateTimeConverter))]
         public DateTime? CanceledAt { get; set; }
+
+        /// <summary>
+        /// Either <c>charge_automatically</c>, or <c>send_invoice</c>. When
+        /// charging automatically, Stripe will attempt to pay this subscription
+        /// at the end of the cycle using the default source attached to the
+        /// customer.  When sending an invoice, Stripe will email your customer
+        /// an invoice with payment instructions.
+        /// </summary>
+        [JsonProperty("collection_method")]
+        public string CollectionMethod { get; set; }
 
         [JsonProperty("created")]
         [JsonConverter(typeof(DateTimeConverter))]
--- a/src/Stripe.net/Enums/Billing.cs
+++ b/src/Stripe.net/Enums/Billing.cs
@@ -1,9 +1,15 @@
 namespace Stripe
 {
+    using System;
     using System.Runtime.Serialization;
     using Newtonsoft.Json;
     using Newtonsoft.Json.Converters;
 
+    /// <summary>
+    /// The field this enum was used for has been renamed to
+    /// <c>CollectionMethod</c> and now accepts string.
+    /// </summary>
+    [Obsolete("Use CollectionMethod instead")]
     [JsonConverter(typeof(StringEnumConverter))]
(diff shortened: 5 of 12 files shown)


src/Stripe.net/Entities/Invoices/Invoice.cs
```diff
@@ -57,6 +57,7 @@ public class Invoice : StripeEntity<Invoice>, IHasId, IHasMetadata, IHasObject
         /// customer. When sending an invoice, Stripe will email your customer an invoice with
         /// payment instructions.
         /// </summary>
+        [Obsolete("Use CollectionMethod")]
```
remi-stripe: I would recommend changing the documentation above too.

src/Stripe.net/Entities/Invoices/InvoiceLineItem.cs
```diff
@@ -70,5 +70,8 @@ public class InvoiceLineItem : StripeEntity<InvoiceLineItem>, IHasId, IHasMetada
 
         [JsonProperty("type")]
         public string Type { get; set; }
+
+        [JsonProperty("unified_proration")]
+        public bool UnifiedProration { get; set; }
```
remi-stripe: Can you add the documentation for that field? We're trying to at least be complete on new fields. While not perfect it helps a bit over time!

src/Stripe.net/Enums/Billing.cs
```diff
@@ -4,6 +4,7 @@ namespace Stripe
     using Newtonsoft.Json;
     using Newtonsoft.Json.Converters;
 
+    // Deprecated. Please use CollectionMethod
```
remi-stripe: Ha, so obviously I didn't remember we had an enum for this field. We likely should do the same enum for CollectionMethod after all.

@ob-stripe Do you think we should have an enum? If we do, do you think we need one per resource, or `InvoiceCollectionMethod` should be used for all 3?
I really don't like `Billing` as being the unique name of that enum so I think the new one should at least be prefixed.

ob-stripe: I'd leave it as a string for now, it doesn't really make sense to use enums for just a couple of properties.

src/Stripe.net/Entities/Invoices/InvoiceLineItem.cs
cjavilla-stripe: added

src/Stripe.net/Entities/Invoices/Invoice.cs
cjavilla-stripe: Added.

src/Stripe.net/Entities/Invoices/Invoice.cs
```diff
@@ -52,11 +52,13 @@ public class Invoice : StripeEntity<Invoice>, IHasId, IHasMetadata, IHasObject
         public bool AutoAdvance { get; set; }
 
         /// <summary>
+        /// Deprecated - use CollectionMethod
```
remi-stripe: Sorry I think I was unclear. https://stripe.com/docs/api/invoices/object#invoice_object-billing the docs explicitly say this now:
> This field has been renamed to collection_method and will be removed in a future API version.

So I think the docs in stripe-dotnet should say the exact same thing, and removing the old docs for the field entirely

In other places in the library we don't because the doc string hasn't been updated, but since we did here, it's just easier.

cjavilla-stripe: got it.

src/Stripe.net/Entities/Invoices/Invoice.cs
```diff
@@ -52,11 +52,10 @@ public class Invoice : StripeEntity<Invoice>, IHasId, IHasMetadata, IHasObject
         public bool AutoAdvance { get; set; }
 
         /// <summary>
-        /// One of <see cref="Billing" />. When charging automatically, Stripe will attempt to pay
-        /// this subscription at the end of the cycle using the default source attached to the
-        /// customer. When sending an invoice, Stripe will email your customer an invoice with
-        /// payment instructions.
+        /// This field has been renamed to collection_method and will be removed
```
remi-stripe: So it should say `CollectionMethod` since you're in .NET and it should be in <c></c>

cjavilla-stripe: :)

src/Stripe.net/Entities/InvoiceItems/InvoiceItem.cs
```diff
@@ -122,6 +122,13 @@ public Subscription Subscription
         [JsonProperty("tax_rates")]
         public List<TaxRate> TaxRates { get; set; }
 
+        /// <summary>
+        /// Set to <c>true</c> if we grouped proration items into one,
+        /// <c>false</c> if not.
+        /// </summary>
+        [JsonProperty("unified_proration")]
+        public bool UnifiedProration { get; set; }
```
remi-stripe: @ob-stripe should this be `bool?`? I think it's still partly gated and might not be returned at all.

src/Stripe.net/Entities/Invoices/InvoiceLineItem.cs
```diff
@@ -70,5 +70,12 @@ public class InvoiceLineItem : StripeEntity<InvoiceLineItem>, IHasId, IHasMetada
 
         [JsonProperty("type")]
         public string Type { get; set; }
+
+        /// <summary>
+        /// Set to <c>true</c> if we grouped proration items into one,
+        /// <c>false</c> if not.
+        /// </summary>
+        [JsonProperty("unified_proration")]
+        public bool UnifiedProration { get; set; }
```
remi-stripe: same as above @ob-stripe

remi-stripe:
Changes lgtm. Assigning to @ob-stripe for approval.

src/Stripe.net/Entities/InvoiceItems/InvoiceItem.cs
cjavilla-stripe: yeah it didn't return for me when I was testing, should probably be `bool?` will wait for ob

cjavilla-stripe: Making it nullable based on the docs. Happy to switch back if @ob-stripe thinks it should be `bool` instead.

ob-stripe: According to the spec, it's not nullable. It's fine to declare a field as non-nullable as long as the API does not return a `null` value for it (i.e. either the API returns a `true` or `false` value for the field, or the field is absent from the JSON).

cjavilla-stripe: ok fixing. thanks!

cjavilla-stripe:
@ob-stripe added `CollectionMethod` to update invoice params.

cjavilla-stripe:
r? @ob-stripe 

Rebased and tested on latest stripe-mock

src/Stripe.net/Enums/Billing.cs
ob-stripe: Regular comments are not visible to users (lest they actually read the library's source code). You should change this to use an [XML comment](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/) and/or the `Obsolete` attribute (probably both).

cjavilla-stripe: Thanks! Updated.

cjavilla-stripe:
Squashed. I was told never to merge these, so I'm reassigning to @remi-stripe  :)

src/Stripe.net/Entities/Invoices/Invoice.cs
```diff
@@ -78,6 +77,16 @@ public Charge Charge
...
 
+        /// <summary>
+        /// Either <c>charge_automatically</c>, or <c>send_invoice</c>. When
+        /// charging automatically, Stripe will attempt to pay this invoice
+        /// using the default source attached to the customer. When sending an
+        /// invoice, Stripe will email this invoice to the customer with payment
+        /// instructions. Defaults to <c>charge_automatically</c>.
+        /// </summary>
+        [JsonProperty("collection_method")]
+        public string CollectionMethod { get; set; }
```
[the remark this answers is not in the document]
cjavilla-stripe: Fixed

src/Stripe.net/Entities/Subscriptions/Subscription.cs
```diff
@@ -54,6 +53,16 @@ public class Subscription : StripeEntity<Subscription>, IHasId, IHasMetadata, IH
...
 
+        /// <summary>
+        /// Either <c>charge_automatically</c>, or <c>send_invoice</c>. When
+        /// charging automatically, Stripe will attempt to pay this subscription
+        /// at the end of the cycle using the default source attached to the
+        /// customer.  When sending an invoice, Stripe will email your customer
+        /// an invoice with payment instructions.
+        /// </summary>
+        [JsonProperty("collection_method")]
+        public string CollectionMethod { get; set; }
```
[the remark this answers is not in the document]
cjavilla-stripe: Done

src/Stripe.net/Services/Invoices/InvoiceCreateOptions.cs
```diff
@@ -40,6 +39,16 @@ public class InvoiceCreateOptions : BaseOptions
...
+        /// <summary>
+        /// Either <c>charge_automatically</c>, or <c>send_invoice</c>. When
+        /// charging automatically, Stripe will attempt to pay this invoice
+        /// using the default source attached to the customer. When sending an
+        /// invoice, Stripe will email this invoice to the customer with payment
+        /// instructions. Defaults to <c>charge_automatically</c>.
+        /// </summary>
+        [JsonProperty("collection_method")]
+        public string CollectionMethod { get; set; }
+
```
remi-stripe: Sorry I didn't flag them all originally and wanted you to recheck each one. That one should be before `Customer` and `CustomFields`

src/Stripe.net/Services/SubscriptionSchedules/SubscriptionScheduleSharedOptions.cs
```diff
@@ -35,6 +34,17 @@ public abstract class SubscriptionScheduleSharedOptions : BaseOptions
...
+        /// Either <c>charge_automatically</c>, or <c>send_invoice</c>. When
+        /// charging automatically, Stripe will attempt to pay the underlying
+        /// subscription at the end of each billing cycle using the default
+        /// source attached to the customer. When sending an invoice, Stripe
+        /// will email your customer an invoice with payment instructions.
+        /// Defaults to <c>charge_automatically</c> on creation.
+        /// </summary>
+        [JsonProperty("collection_method")]
+        public string CollectionMethod { get; set; }
+
```
remi-stripe: This should be before `DefaultSourceId` and `DefaultPaymentMethodId`

cjavilla-stripe:
@remi-stripe sorry, I thought since there were two comments those were the only two. I rechecked them all, I'm not sure if that one should go after the collapsable charge block or not, not sure how those regions work.

PR #1680, opened 2019-06-26
Allow null API key in StripeClient

### LINKED ISSUE #1650: API Key error on stripe terminal
Windows 10
.Net 4.7.2
Stripe 27.0.0

Upgrading stripe terminal out of beta to v1 and thought i'd upgrade stripe.net and api as well.

Calling the following was fine before the upgrade:

```vb
Return New Stripe.Terminal.ConnectionTokenService().Create(Nothing, New 
     Stripe.RequestOptions With {
    .ApiKey = GetKey(XXXXXX)
})
```

Now, returns the following error:
```
No API key provided. Set your API key using `StripeConfiguration.ApiKey = "<API-KEY>"`. You can generate API keys from the Stripe Dashboard. See https://stripe.com/docs/api/authentication for details or contact support at https://support.stripe.com/email if you have any questions.
```

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries-releasers 

Don't enforce an API key when creating `StripeClient`, so that users can create clients without API keys and pass the key only via `RequestOptions` on a per-request basis.

We still check for empty strings or whitespace, and we do check that an API key was provided (either by `RequestOptions` or by the client) when creating an actual API request.

Fixes #1650.

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/StripeClient.cs
+++ b/src/Stripe.net/Infrastructure/Public/StripeClient.cs
@@ -34,24 +34,19 @@ namespace Stripe
         /// if <c>apiKey</c> is empty or contains whitespace
         /// </exception>
         public StripeClient(
-            string apiKey,
+            string apiKey = null,
             string clientId = null,
             IHttpClient httpClient = null,
             string apiBase = null,
             string connectBase = null,
             string filesBase = null)
         {
-            if (apiKey == null)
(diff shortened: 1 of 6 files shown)


ob-stripe:
Thanks Brandur!

ob-stripe:
Released as 27.4.0.

PR #1682, opened 2019-06-26
Add support for the `SetupIntent` resource and APIs

remi-stripe:
This PR does not add the docs because the strings are not finalized. I will commit to fixing those (and the PaymentIntent ones) after the API is public. We would like to merge ahead of that though.

r? @ob-stripe 
cc @stripe/api-libraries @danwang-stripe

### DIFF
--- a/src/Stripe.net/Entities/SetupIntents/SetupIntent.cs
+++ b/src/Stripe.net/Entities/SetupIntents/SetupIntent.cs
@@ -5,22 +5,13 @@ namespace Stripe
     using Newtonsoft.Json;
     using Stripe.Infrastructure;
 
-    public class PaymentIntent : StripeEntity<PaymentIntent>, IHasId, IHasMetadata, IHasObject
+    public class SetupIntent : StripeEntity<SetupIntent>, IHasId, IHasMetadata, IHasObject
     {
         [JsonProperty("id")]
         public string Id { get; set; }
 
         [JsonProperty("object")]
         public string Object { get; set; }
-
-        [JsonProperty("amount")]
-        public long? Amount { get; set; }
-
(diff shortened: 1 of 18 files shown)


src/Stripe.net/Entities/SetupIntents/SetupIntent.cs
```diff
@@ -0,0 +1,130 @@
...
+            set => this.InternalApplication = SetExpandableFieldObject(value, this.InternalApplication);
+        }
+
+        [JsonProperty("application")]
+        [JsonConverter(typeof(ExpandableFieldConverter<Application>))]
+        internal ExpandableField<Application> InternalApplication { get; set; }
+        #endregion
+
+        [JsonProperty("cancellation_reason")]
+        public string CancellationReason { get; set; }
```
remi-stripe: Not in the spec yet but the PR is being merged right now.

src/Stripe.net/Services/SetupIntents/SetupIntentGetOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+    using Newtonsoft.Json;
+
+    public class SetupIntentGetOptions : BaseOptions
+    {
+        /// <summary>
+        /// The client secret of the SetupIntent. Required if a publishable key is used to
+        /// retrieve the setup intent.
+        /// </summary>
+        [JsonProperty("client_secret")]
+        public string ClientSecret { get; set; }
```
remi-stripe: I kept those since some developers use our library in mobile apps

remi-stripe:
re-assigning to myself as there's a breaking change being merged in.

PR #1686, opened 2019-07-01
Properly support `OffSession` to be a bool or an enum value

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries


remi-stripe:
Yeah, I'm so glad @ob-stripe built this recently!

PR #1689, opened 2019-07-02
Add documentation for the `PaymentIntent` resource and parameters

remi-stripe:
This adds documentation for all properties and parameters. It also removes the Shared options for Create and Update as this tends to lead to bugs. We also mark `transfer_data[amount]` as not deprecated anymore as it will be ungated soon.

r? @ob-stripe 
cc @stripe/api-libraries


src/Stripe.net/Services/PaymentIntents/PaymentIntentUpdateOptions.cs
```diff
@@ -1,9 +1,125 @@
...
+        /// use case for connected accounts</a> for details.
+        /// </summary>
+        [JsonProperty("transfer_group")]
+        public string TransferGroup { get; set; }
+
+        [Obsolete("Use PaymentMethodTypes")]
+        [JsonProperty("allowed_source_types")]
+        public List<string> AllowedSourceTypes { get; set; }
+
+        [Obsolete("This field was never supported in the API so marking as deprecated/")]
```
ob-stripe: If this was indeed never supported by the API, I think we can simply remove the field without a major version.

remi-stripe: Yeah but it will likely be supported in a few weeks so I wanted to avoid the churn :p

ob-stripe: We'd still have to remove the `Obsolete` attribute, but fair enough :)

Can you change the comment to just "This field was never supported in the API" (or maybe "This field is not yet supported in the API")?

ob-stripe: Haha, too slow I guess. Not a big deal.

PR #1699, opened 2019-07-09
Init options with forced parameter values

### LINKED ISSUE #1676: Cards are not being deserialized properly
Using: Stripe.net 26.0.0 on .NET Core 3 preview 5

I know this is a preview build of .NET Core, but since every other piece of the library works fine, I am not sure if the issue is related to Stripe or .NET Core 3.

When using the CardService class to retrieve cards from the API, they are not being deserialized correctly. Every value on the Card ends up being null. This happens both when using the List methods and when using the Get methods. The StripeResponse.ObjectJson property does have all the values that should be there, but the object itself does not. Current workaround is to just deserialize it myself using the ObjectJson property.

ob-stripe:

Some options classes have forced parameter values, e.g. `CardListOptions` always includes `object=card`. However if the user passes `null` then the parameter is not sent.

This is kind of an ugly ad-hoc fix. Ideally we would always create an instance of the appropriate options class, but doing this would require us to make all the helper methods in `Service` generic to accept the options class as a type param, and I'd rather not do that just yet.

Fixes #1676.

### DIFF
--- a/src/Stripe.net/Services/BankAccounts/BankAccountService.cs
+++ b/src/Stripe.net/Services/BankAccounts/BankAccountService.cs
@@ -60,17 +60,17 @@ namespace Stripe
 
         public virtual StripeList<BankAccount> List(string customerId, BankAccountListOptions options = null, RequestOptions requestOptions = null)
         {
-            return this.ListNestedEntities(customerId, options, requestOptions);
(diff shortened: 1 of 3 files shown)


brandur-stripe:
+1 Remi's comments, but LGTM otherwise. (Thanks for the quick fix!)

PR #1700, opened 2019-07-09
Add PaymentSourceService

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add `PaymentSourceService`. Like `ExternalAccountService` but for customer payment sources.

### DIFF
--- a/src/Stripe.net/Services/PaymentSourceService/PaymentSourceCreateOptions.cs
+++ b/src/Stripe.net/Services/PaymentSourceService/PaymentSourceCreateOptions.cs
@@ -1,0 +1,39 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class PaymentSourceCreateOptions : BaseOptions
+    {
+        /// <summary>
+        /// Either a token, like the ones returned by
+        /// <a href="https://stripe.com/docs/stripe.js">Stripe.js</a>, or a
+        /// <see cref="SourceBankAccount"/> instance containing a user's bank account details.
+        /// </summary>
+        [JsonProperty("bank_account")]
+        [JsonConverter(typeof(AnyOfConverter))]
+        public AnyOf<string, SourceBankAccount> BankAccount { get; set; }
+
+        /// <summary>
+        /// A token, like the ones returned by
+        /// <a href="https://stripe.com/docs/stripe.js">Stripe.js</a>.
+        /// </summary>
+        [JsonProperty("card")]
+        [JsonConverter(typeof(AnyOfConverter))]
+        public AnyOf<string, CardCreateNestedOptions> Card { get; set; }
+
+        /// <summary>
+        /// A set of key-value pairs that you can attach to a card object. It can be useful for
+        /// storing additional information about the card in a structured format.
+        /// </summary>
+        [JsonProperty("metadata")]
+        public Dictionary<string, string> Metadata { get; set; }
+
+        /// <summary>
+        /// Please refer to full <a href="https://stripe.com/docs/api">documentation</a> instead.
+        /// </summary>
+        [JsonProperty("source")]
+        public string Source { get; set; }
+   }
+}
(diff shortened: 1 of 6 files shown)


remi-stripe:
@ob-stripe Thanks for making this happen quickly! I should have been a bit more explicit in my other review. I think we need one small method to do this like on `CustomerService.ListPaymentSources` or something like this. I'm a little bit worried adding a brand new service like this might be confusing as IPaymentSource is technically deprecated so adding a brand new logic like that might be confusing.

I'm not opposed, I've wanted this for a while, I just wonder if we'll end up regretting it in a few months once PaymentMethods support most source types.

src/Stripe.net/Services/PaymentSourceService/PaymentSourceCreateOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class PaymentSourceCreateOptions : BaseOptions
+    {
+        /// <summary>
+        /// Either a token, like the ones returned by
+        /// <a href="https://stripe.com/docs/stripe.js">Stripe.js</a>, or a
+        /// <see cref="SourceBankAccount"/> instance containing a user's bank account details.
+        /// </summary>
+        [JsonProperty("bank_account")]
+        [JsonConverter(typeof(AnyOfConverter))]
+        public AnyOf<string, SourceBankAccount> BankAccount { get; set; }
```
remi-stripe: I don't think we should support `bank_account` and `card`. Those are legacy parameters that were deprecated in 2015-02-18. The details should be passed in `source` instead: https://stripe.com/docs/api/cards/create#create_card-source

src/Stripe.net/Services/PaymentSourceService/PaymentSourceListOptions.cs
```diff
@@ -0,0 +1,11 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+
+    public class PaymentSourceListOptions : ListOptions
+    {
+        /// <summary>Filter sources according to a particular object type.</summary>
+        [JsonProperty("object")]
+        public string Object { get; set; }
```
remi-stripe: We likely would have to show a list of enums since it's not documented anywhere right?

src/Stripe.net/Services/_base/Service.cs
```diff
@@ -258,7 +258,7 @@ protected EntityReturned CreateEntity(BaseOptions options, RequestOptions reques
                 string itemId = null;
                 foreach (var item in page)
                 {
-                    itemId = ((IHasId)item).Id;
+                    itemId = (item as IHasId)?.Id;
```
remi-stripe: What is this change doing?

src/Stripe.net/Services/PaymentSourceService/PaymentSourceCreateOptions.cs
ob-stripe: Fine by me, but both params are still in the spec right now.

src/Stripe.net/Services/PaymentSourceService/PaymentSourceListOptions.cs
ob-stripe: We probably should, but the spec doesn't include a list of valid values (neither in the description string or by defining `object` as an enum).

src/Stripe.net/Services/_base/Service.cs
ob-stripe: Makes the code a little bit safer. If for some reason `item` is not castable to `IHasId`, then `itemId` will be `null` and the iterator will stop as if it had reached the end of the list. The current version crashes if `item` cannot be cast to `IHasId`. (In practice, this should never happen: all listable resources have IDs. I think stripe-mock was returning invalid data in one of the tests but I didn't dig too deep.)

remi-stripe: Ack. Can we make this its own PR? sorry to be annoying but this is the kind of change that will cause us trouble because we suddenly start returning more results/things.

ob-stripe:
This is pretty stale, so I'm closing it. We can always open another PR if needed.

PR #1701, opened 2019-07-09
Make autopagination a bit safer

ob-stripe:
cc @stripe/api-libraries 

Make autopagination a bit safer by using the `as` keyword to cast page items to `IHasId`, and using the safe navigation operator to get the ID. If for some reason the cast fails, `itemId` will be `null` and the iterator will stop per the condition on line 265.

In practice, this should never happen as listable resources should always have IDs, but I ran into this in a test, probably because of a weird stripe-mock fixture.

### DIFF
--- a/src/Stripe.net/Services/_base/Service.cs
+++ b/src/Stripe.net/Services/_base/Service.cs
@@ -258,7 +258,7 @@ namespace Stripe
                 string itemId = null;
                 foreach (var item in page)
                 {
-                    itemId = ((IHasId)item).Id;
+                    itemId = (item as IHasId)?.Id;
                     yield return item;
                 }
 


remi-stripe:
Deferring to brandur on that one as I can't seem to understand what could cause this. We should also make this part of a major version to be ultra safe.

brandur-stripe:
I'd never noticed before how well the `as` and `?` operators work together — nice!

This seems like a good improvement, but just want to point out that there's usually a side effect in this kind of change in that it can have the effect of making problems less visible — e.g. where before that bad fixture would throw an error and allow you to easily notice it, now the library might just stop paginating or whatever, which could be confusing, and might lead to making the actual problem of the bad fixture harder to find as you have to trace your way back to it. I can go either way though, so up to you!

LGTM.

ob-stripe:
@brandur-stripe Very good point. On second thought this "fix" does sweep potential problems under the rug, so I'm not going to merge it.

PR #1702, opened 2019-07-09
Don't create AnyOf instances with null values

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Modifies the implicit operators used to create `AnyOf<>` instances from value types to return `null` instead of creating a new `AnyOf` instance that contains `null` when the value is itself `null`,

Previous behavior:
```c#
AnyOf<int, string> anyOf = null;
anyOf = (string)null;
Assert.NotNull(anyOf);
// anyOf is now a (non-null) AnyOf<int, string> instance with `Value` set to `null`
```

New behavior:
```c#
AnyOf<int, string> anyOf = null;
anyOf = (string)null;
Assert.Null(anyOf);
// anyOf is still null
```

I've also changed `ToString()`, as previously it would crash if `Value` was null.

### DIFF
--- a/src/Stripe.net/Services/_base/AnyOf.cs
+++ b/src/Stripe.net/Services/_base/AnyOf.cs
@@ -18,7 +18,7 @@ namespace Stripe
 
         /// <summary>Returns a string that represents the current object.</summary>
         /// <returns>A string that represents the current object.</returns>
-        public override string ToString() => this.Value.ToString();
+        public override string ToString() => this.Value == null ? "AnyOf(null)" : this.Value.ToString();
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Stripe.net/Services/_base/AnyOf.cs
```diff
@@ -103,14 +103,14 @@ public override Type Type
         /// </summary>
         /// <param name="value">The value to convert.</param>
         /// <returns>An <see cref="AnyOf{T1, T2}"/> object that holds the value.</returns>
-        public static implicit operator AnyOf<T1, T2>(T1 value) => new AnyOf<T1, T2>(value);
+        public static implicit operator AnyOf<T1, T2>(T1 value) => value == null ? null : new AnyOf<T1, T2>(value);
```
ob-stripe: Unfortunately, because this is a generic class, I don't think it's possible to write a helper method to do this in a single place.

ob-stripe:
> OOC, was there some particular impetus for this, or did you just notice recently and decide to try to make it more correct?

A user ran into this when writing tests for their Stripe integration.

The previous behavior was not strictly a bug since the library would omit the parameter in the request in both cases (whether the `AnyOf` property is `null` or is non-null but contains a `null` value), but they were surprised that they were getting back a non-null value after explicitly assigning a null value.

ob-stripe:
Released as 27.9.1.

PR #1707, opened 2019-07-12
Add support for undeclared fields in StripeEntity

ob-stripe:
cc @stripe/api-libraries 

Similar to https://github.com/stripe/stripe-java/pull/799, this lets users access fields that don't have a matching property in the Stripe.net model class.

This relies on Newtonsoft.Json's [`JsonExtensionData`](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_JsonExtensionDataAttribute.htm) attribute. Like the Java PR, this feature leaks the underlying JSON library to users. However, unlike the Java PR, I don't think we need an extended warning here:
1. Unlike Java, `ExtraFields` will contain _only_ fields that don't have a matching property, so it cannot be use to access "normal" declared fields.
2. We're unlikely to change the JSON library used by Stripe.net

Happy to add a comment if you think it's preferable though!

### DIFF
--- a/src/Stripe.net/Entities/_base/StripeEntity.cs
+++ b/src/Stripe.net/Entities/_base/StripeEntity.cs
@@ -1,13 +1,19 @@
 namespace Stripe
 {
+    using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using System.Reflection;
     using System.Runtime.CompilerServices;
     using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
 
     [JsonObject(MemberSerialization.OptIn)]
     public abstract class StripeEntity : IStripeEntity
     {
+        /// <summary>Dictionary containing extra resource fields.</summary>
+        [JsonExtensionData]
+        public IDictionary<string, JToken> ExtraFields { get; set; }
+
         [JsonIgnore]
         public StripeResponse StripeResponse { get; set; }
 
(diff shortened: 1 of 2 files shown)


brandur-stripe:
> Happy to add a comment if you think it's preferable though!

Nah, I think you're right. This seems fine.

One possible thing to flag while I'm thinking about it: there is a potential failure mode whereby a user might jump onto an "extra property" for use while a feature is in prerelease, then later we promote that property to a _real_ property which causes it to stop appearing in the extra properties, which then causes a bug in the user's integration now that it's missing. Because they were only accessing an untyped map, the compiler doesn't save them with an error or anything.

Not sure what the best thing to do about that would be though — we might just have to flag to those users to be very careful when using these properties and performing upgrades.

LGTM.

remi-stripe:
> Not sure what the best thing to do about that would be though — we might just have to flag to those users to be very careful when using these properties and performing upgrades.

Honestly, this is almost a blocker for me. This will happen consistently as we add gated features and then release them as minor versions. I think it makes this too dangerous.

ob-stripe:
Yeah, good point. Let's scratch this and wait until someone actually needs this? It would be fairly trivial to reimplement the stripe-java approach (i.e. deserialize the JSON payload a second time, but in an untyped `Dictionary<string, JToken>` map instead of a model class).

PR #1714, opened 2019-07-19
Remove deprecated ThreeDSecure object and APIs

remi-stripe:
This endpoint has been deprecated for 3 years. It also has not been used with dotnet in a month.

I would be fine with closing as won't do just because the endpoint is still considered supported.

r? @ob-stripe 
cc @stripe/api-libraries


remi-stripe:
Discussed internally, we won't do for now as it's still part of the official openapi spec.

PR #1718, opened 2019-07-23
Fix NullReferenceException when auto-paginating without params

### LINKED ISSUE #1717: BalanceTransactionService.ListAutoPaging Null Reference Exception
When using the `ListAutoPaging` feature for `BalanceTransactions`, the service throws a null reference exception as it tries to go past the number of records in the result set. 

i.e. the default option is to limit the returned result set to 10 records. 
if I run the following, 10 records will be added to my list and the loop will throw the exception when trying to access an 11th record. 

```csharp
foreach (var balanceTransaction in _balanceTransactionService.ListAutoPaging())
{
      balanceTransactions.Add(balanceTransaction);
}
```

OSVersion: Windows 10 : Build 17134
.NETVersion : dotnet core 2.1
StripeVersion: 27.16.0

ob-stripe:
cc @stripe/api-libraries 

Fixes an issue where the auto-pagination method would crash with a `NullReferenceException` when trying to fetch the second page if no parameters were provided.

Fixes #1717.

### DIFF
--- a/src/Stripe.net/Services/_base/Service.cs
+++ b/src/Stripe.net/Services/_base/Service.cs
@@ -276,7 +276,9 @@ namespace Stripe
                     break;
                 }
 
+                options = options ?? new ListOptions();
                 options.StartingAfter = itemId;
+
                 page = this.Request<StripeList<T>>(
                     HttpMethod.Get,
                     url,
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released as 27.16.1.

PR #1723, opened 2019-07-29
Add Klarna Source

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

I want to submit the Klarna and SourceOrder stuff separately here. Did I get these references right?

### DIFF
--- a/src/Stripe.net/Entities/Sources/Source.cs
+++ b/src/Stripe.net/Entities/Sources/Source.cs
@@ -66,6 +66,14 @@ namespace Stripe
         public Dictionary<string, string> Metadata { get; set; }
 
         /// <summary>
+        /// Information about the items and shipping associated with the source.
+        /// Required for transactional credit (for example Klarna) sources
+        /// before you can charge it.
+        /// </summary>
+        [JsonProperty("source_order")]
+        public SourceSourceOrder SourceOrder { get; set; }
+
+        /// <summary>
         /// Information about the owner of the payment instrument that may be used or required by particular source types.
         /// </summary>
         [JsonProperty("owner")]
@@ -96,7 +104,7 @@ namespace Stripe
         public string Status { get; set; }
 
         /// <summary>
-        /// The type of the source. The type is a payment method, one of card, three_d_secure, giropay, sepa_debit, ideal, sofort, or bancontact.
+        /// The type of the source. The type is a payment method, one of card, three_d_secure, giropay, sepa_debit, ideal, klarna, sofort, or bancontact.
         /// </summary>
         [JsonProperty("type")]
         public string Type { get; set; }
@@ -146,6 +154,9 @@ namespace Stripe
 
         [JsonProperty("ideal")]
         public SourceIdeal Ideal { get; set; }
+
+        [JsonProperty("klarna")]
+        public SourceKlarna Klarna { get; set; }
 
         [JsonProperty("multibanco")]
         public SourceMultibanco Multibanco { get; set; }
--- a/src/Stripe.net/Entities/Sources/SourceKlarna.cs
+++ b/src/Stripe.net/Entities/Sources/SourceKlarna.cs
@@ -1,0 +1,82 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+
+    public class SourceKlarna : StripeEntity<SourceIdeal>
+    {
+        [JsonProperty("background_image_url")]
+        public string BackgroundImageUrl { get; set; }
+
+        [JsonProperty("client_token")]
+        public string ClientToken { get; set; }
+
+        [JsonProperty("first_name")]
+        public string FirstName { get; set; }
+
+        [JsonProperty("last_name")]
+        public string LastName { get; set; }
+
+        [JsonProperty("locale")]
+        public string Locale { get; set; }
+
+        [JsonProperty("logo_url")]
+        public string LogoUrl { get; set; }
+
+        [JsonProperty("page_title")]
+        public string PageTitle { get; set; }
+
+        [JsonProperty("pay_later_asset_urls_descriptive")]
+        public string PayLaterAssetUrlsDescriptive { get; set; }
+
+        [JsonProperty("pay_later_asset_urls_standard")]
+        public string PayLaterAssetUrlsStandard { get; set; }
+
+        [JsonProperty("pay_later_name")]
+        public string PayLaterName { get; set; }
+
+        [JsonProperty("pay_later_redirect_url")]
+        public string PayLaterRedirectUrl { get; set; }
+
+        [JsonProperty("pay_now_asset_urls_descriptive")]
+        public string PayNowAssetUrlsDescriptive { get; set; }
+
+        [JsonProperty("pay_now_asset_urls_standard")]
+        public string PayNowAssetUrlsStandard { get; set; }
+
+        [JsonProperty("pay_now_name")]
+        public string PayNowName { get; set; }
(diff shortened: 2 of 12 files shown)


src/Stripe.net/Entities/Sources/SourceKlarna.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// The last name of the customer.
+        /// </summary>
+        [JsonProperty("last_name")]
+        public string LastName { get; set; }
+
+        /// <summary>
+        /// Format locales as a language tag consisting of a two-letter language
+        /// code combined with a two-letter country code, according to RFC 1766.
+        /// Examples are en-us for U.S. English, en-gb for British English, and
+        /// sv-se for Swedish (in Sweden). (RFC1766).
```
remi-stripe: Hmmm that one is extremely specific and would never make it to our public docs (or really should not). I'd recommend simplifying the wording here

src/Stripe.net/Entities/Sources/SourceKlarna.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        [JsonProperty("pay_over_time_redirect_url")]
+        public string PayOverTimeRedirectUrl { get; set; }
+
+        [JsonProperty("payment_method_categories")]
+        public string PaymentMethodCategories { get; set; }
+
+        /// <summary>
+        /// The ISO-3166 2-letter country code of the customer's location.
+        /// </summary>
```
remi-stripe: same, we don't really mention the ISO thing. We should, for sure, but we should avoid those really specific wordings in the library

cjavilla-stripe:
r? @remi-stripe 

Thanks for taking a look...


Couple of questions:

1) should Order be SourceSourceOrder or just Order like in Java https://github.com/stripe/stripe-java/pull/800/files ? (following from that OrderShippingOptions, OrderItem etc.) 
2) should Order Item Type be an enum? I can't remember if we use enums in this language.
3) I removed some of the docs for that Klarna type, but expected we would want the docs on these Source Order types. Should we keep those doc strings?

remi-stripe:
> should Order be SourceSourceOrder or just Order like in Java https://github.com/stripe/stripe-java/pull/800/files ? (following from that OrderShippingOptions, OrderItem etc.)

We don't nest classes so it can't be `Order` otherwise it would conflict as they are in the same namespace. It has to be `SourceSourceOrder` like in go

> should Order Item Type be an enum? I can't remember if we use enums in this language.

No enums for now, not until we auto-generate the library.

> I removed some of the docs for that Klarna type, but expected we would want the docs on these Source Order types. Should we keep those doc strings?

I usually put the docs if they are in the API reference, otherwise I don't

src/Stripe.net/Entities/Sources/Source.cs
```diff
@@ -65,6 +65,12 @@ public class Source : StripeEntity<Source>, IHasId, IHasMetadata, IHasObject, IP
         [JsonProperty("metadata")]
         public Dictionary<string, string> Metadata { get; set; }
 
+        /// <summary>
+        /// Information about the items and shipping associated with the source.
+        /// </summary>
```
remi-stripe: I would keep the klarna comment from the API ref here

src/Stripe.net/Services/Sources/SourceSourceOrderItemOptions.cs
```diff
@@ -0,0 +1,23 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class SourceSourceOrderItemOptions : INestedOptions
+    {
+        [JsonProperty("amount")]
```
remi-stripe: Those have official docs in the API reference. Could you add those?

src/Stripe.net/Services/Sources/SourceSourceOrderShippingOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+
+    public class SourceSourceOrderShippingOptions : ShippingOptions
+    {
+        [JsonProperty("carrier")]
+        public string Carrier { get; set; }
+
+        [JsonProperty("tracking_number")]
+        public string TrackingNumber { get; set; }
+    }
+}
```
remi-stripe: There's `ChargeShippingOptions` for this (the name is definitely not great but that's what we use in other places). Not sure what the best approach is here :(
Having a custom options one when the resource one is generic is weird. But maybe a problem from another time. Will defer to @ob-stripe on that one.

src/Stripe.net/Services/Sources/SourceSourceOrderItemOptions.cs
cjavilla-stripe: Added doc strings from the API reference.

cjavilla-stripe:
Updated.

r? @remi-stripe 

ptal

remi-stripe:
LGTM, assigning to @ob-stripe to be safe

src/Stripe.net/Entities/Sources/SourceSourceOrderItem.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    using Newtonsoft.Json;
+
+    public class SourceSourceOrderItem : StripeEntity<SourceSourceOrderItem>
+    {
+        /// <summary>
+        /// The amount (price) for this order item.
+        /// </summary>
+        [JsonProperty("amount")]
+        public long Amount { get; set; }
```
ob-stripe: This attribute is flagged as nullable in the spec, so should be `long?`.

src/Stripe.net/Entities/Sources/SourceSourceOrderItem.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// </summary>
+        [JsonProperty("description")]
+        public string Description { get; set; }
+
+        /// <summary>
+        /// The quantity of this order item. When type is <c>sku</c>, this is
+        /// the number of instances of the SKU to be ordered.
+        /// </summary>
+        [JsonProperty("quantity")]
+        public long? Quantity { get; set; }
```
ob-stripe: This attribute is not flagged as nullable in the spec, so should be `long` (though maybe we can keep it nullable anyway to avoid a future breaking change... @remi-stripe wdyt?).

remi-stripe: I'd be fine with not nullable since `Amount` is not nullable

remi-stripe: Ha `Amount` *is* nullable, so we should make Quantity nullable
and also ask why one can be and not the other

src/Stripe.net/Entities/Sources/SourceSourceOrderItem.cs
cjavilla-stripe: Updated to nullable.

src/Stripe.net/Entities/Sources/SourceSourceOrderItem.cs
cjavilla-stripe: How would I figure out who to ask why quantity is not nullable in the spec?

cjavilla-stripe:
r? @ob-stripe 

I think this one is ready for another look. Squashed, so the comments may have lost context, but I think we're good with one action item to ask why source order item quantity is not nullable in the spec.

PR #1733, opened 2019-08-08
Add missing IHas* interfaces everywhere

ob-stripe:
cc @stripe/api-libraries 

This PR:
- adds a wholesome test to check that Stripe entities and options classes that can implement a specific `IHas*` interface do inherit the interface
- adds the interfaces to all options classes (previously they were only applied to entities classes)


ob-stripe:
Thanks for the quick review!

ob-stripe:
Released as 27.22.0.

PR #1745, opened 2019-08-14
Fix duplicate expand values issue

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

When using the `ExpandX` properties on a service class, expand values in the actual HTTP request would be duplicated if the same options class instance is reused for multiple requests. This is most notable the case with auto-pagination.

This fixes the issue by only adding expand values if they're not already present in the `Expand` list of the options class instance.


ob-stripe:
Released as 27.25.1.

PR #1752, opened 2019-08-19
Bump Stylecop.Analyzers to latest version

ob-stripe:
cc @stripe/api-libraries 

Bump Stylecop.Analyzers (linter) to latest version, and fix all new warnings.

Nothing major, mostly tweaking and adding documentation strings, and renaming type parameters in generic classes/interfaces so they start with `T`.

### DIFF
--- a/src/StripeTests/StripeTests.csproj
+++ b/src/StripeTests/StripeTests.csproj
@@ -6,6 +6,7 @@
     <AssemblyName>StripeTests</AssemblyName>
     <PackageId>StripeTests</PackageId>
     <RuntimeIdentifiers>win10-x64</RuntimeIdentifiers>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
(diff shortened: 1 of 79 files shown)


src/StripeTests/StripeTests.csproj
```diff
@@ -6,6 +6,7 @@
     <AssemblyName>StripeTests</AssemblyName>
     <PackageId>StripeTests</PackageId>
     <RuntimeIdentifiers>win10-x64</RuntimeIdentifiers>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
ob-stripe: It's surprising to enable documentation output in a test project, but apparently disabling this can cause strange issues with the compiler, so it's recommended to enable it even if not needed: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2192#issuecomment-241396037

PR #1771, opened 2019-08-29
Add EnableTelemetry flag

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds an `EnableTelemetry` flag to let users disable request latency telemetry if they wish.


ob-stripe:
Released as 28.7.0.

Issue #1782, opened 2019-09-11
List credit notes error


realnickivanov:
OS: Windows Server 2012 R2 Standard 64bit
.Net: 4.6.1
Stripe.Net: 27.24.0.0

Running the following code
```
var iteration = await _stripeCreditNoteService.ListAsync(new CreditNoteListOptions
{
    CustomerId = stripeCustomerId,
    Limit = 100,
    StartingAfter = startingAfter
});
```
I get the following exeption
```
Stripe.StripeException: Invalid response object from API: "{
  "object": "list",
  "data": [
    {
      "id": "cn_1F45mVH8CGtPy8T0TXecJF4H",
      "object": "credit_note",
      "amount": 100,
      "created": 1565009959,
      "currency": "eur",
      "customer": "cus_FZE647sLbykyuz",
      "customer_balance_transaction": null,
      "invoice": "in_1F45dzH8CGtPy8T0sNFR5YGW",
      "livemode": false,
      "memo": "Test",
      "metadata": {
      },
      "number": "U6FITOAI-0001-CN-01",
      "pdf": "https://pay.stripe.com/credit_notes/*************/***********/pdf",
      "reason": "duplicate",
      "refund": "re_1F45mVH8CGtPy8T0ZSO8OQBt",
      "status": "issued",
      "type": "post_payment",
      "voided_at": null
    }
  ],
  "has_more": false,
  "url": "/v1/credit_notes"
}
"
   at Stripe.StripeClient.ProcessResponse[T](StripeResponse response)
   at Stripe.StripeClient.<RequestAsync>d__25`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Stripe.Service`1.<RequestAsync>d__23`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
```

For example see request req_7dEJlJX9faTa9F 

I also tried Stripe.Net 27.25.1. Same error.

remi-stripe:
[...]

PR #1783, opened 2019-09-11
Allow VoidedAt on CreditNote to be nullable

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/1782

r? @ob-stripe 
cc @stripe/api-libraries 

Is this a breaking change?

### DIFF
--- a/src/Stripe.net/Entities/CreditNote.cs
+++ b/src/Stripe.net/Entities/CreditNote.cs
@@ -190,6 +190,6 @@ namespace Stripe
         /// </summary>
         [JsonProperty("voided_at")]
         [JsonConverter(typeof(DateTimeConverter))]
-        public DateTime VoidedAt { get; set; }
+        public DateTime? VoidedAt { get; set; }
     }
 }


ob-stripe:
> Is this a breaking change?

Technically, yeah, it is. `?` is just syntactic sugar, this PR changes the type of `VoidedAt` from `DateTime` to `System.Nullable<DateTime>`. Users would have to call `VoidedAt.HasValue` to check if the value is null or not, and `VoidedAt.Value` to get the actual `DateTime` struct if it's not null.

remi-stripe:
@ob-stripe  Yeah I know it technically is, but is it in this case? I'm assuming almost no one could deserialize CreditNote until that fix

ob-stripe:
Yeah, we can probably release this as a patch or minor since it's really a bug fix, with a note that it's technically breaking.

remi-stripe:
I'm happy to wait for the next major otherwise if it's coming soon. Can I leave to you otherwise?

PR #1789, opened 2019-09-20
Add Last4 in AU BECS Source and Deprecating Last3

shengwei-stripe:
@remi-stripe @ob-stripe 
This is to update AUDD `SourceAuBecsDebit` Parameter. Deprecating `Last3` and Add `Last4` 
Related Pull request (https://github.com/stripe/stripe-go/pull/952)

cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/Sources/SourceAuBecsDebit.cs
+++ b/src/Stripe.net/Entities/Sources/SourceAuBecsDebit.cs
@@ -1,5 +1,6 @@
 namespace Stripe
 {
+    using System;
     using Newtonsoft.Json;
 
     public class SourceAuBecsDebit : StripeEntity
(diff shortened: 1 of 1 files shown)


src/Stripe.net/Entities/Sources/SourceAuBecsDebit.cs
```diff
@@ -13,7 +14,11 @@ public class SourceAuBecsDebit : StripeEntity
         [JsonProperty("fingerprint")]
         public string Fingerprint { get; set; }
 
+        [Obsolete("This property is deprecated, please use Last4 going forward.")]
```
remi-stripe: I realized that `AccountNumber` is in the resource and shouldn't be right? It's not returned on the resource. We likely should also mark it as obsolete.

shengwei-stripe: @remi-stripe  you are right. AccountNumber is not returned any more. Make it as obsolete too.

shengwei-stripe:
> Also you're still not adding `SourceMandate` at all as the resource, only params. Is that expected? I also don't see it in java, but it's partially in go.

Yes @remi-stripe , I did not see `SourceMandate` use anywhere as API resource. The `SourceMandate` was there in Golang before. It was tested for a very old API I think

PR #1804, opened 2019-10-07
Enable test for JSON<->property name consistency

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Enables the JSON<->property name consistency test.

Adds a `AllowNameMismatch` attribute to flag allowed exceptions.

### DIFF
--- a/src/Stripe.net/Services/Persons/PersonSharedOptions.cs
+++ b/src/Stripe.net/Services/Persons/PersonSharedOptions.cs
@@ -1,6 +1,5 @@
 namespace Stripe
 {
-    using System;
     using System.Collections.Generic;
     using Newtonsoft.Json;
     using Stripe.Infrastructure;
@@ -65,6 +64,7 @@ namespace Stripe
         public PersonRelationshipOptions Relationship { get; set; }
 
         [JsonProperty("ssn_last_4")]
+        [AllowNameMismatch]
         public string SSNLast4 { get; set; }
 
         [JsonProperty("verification")]
--- a/src/Stripe.net/Infrastructure/AllowNameMismatch.cs
+++ b/src/Stripe.net/Infrastructure/AllowNameMismatch.cs
@@ -1,0 +1,17 @@
+namespace Stripe.Infrastructure
+{
+    using Newtonsoft.Json;
+
+    /// <summary>
+    /// Used to indicate that a particular property is allowed to have a name that isn't a strict
+    /// <c>snake_case</c> -> <c>CamelCase</c> conversion from the JSON name defined its
+    /// <see cref="JsonPropertyAttribute" /> property.
+    /// </summary>
+    /// <remarks>
+    /// This is only used for an internal wholesome test.
+    /// </remarks>
+    [System.AttributeUsage(System.AttributeTargets.Property)]
+    internal class AllowNameMismatch : System.Attribute
+    {
+    }
+}
--- a/src/Stripe.net/Services/Invoices/InvoiceUpcomingInvoiceItemOption.cs
+++ b/src/Stripe.net/Services/Invoices/InvoiceUpcomingInvoiceItemOption.cs
@@ -2,6 +2,7 @@ namespace Stripe
 {
     using System.Collections.Generic;
     using Newtonsoft.Json;
+    using Stripe.Infrastructure;
 
     public class InvoiceUpcomingInvoiceItemOption : INestedOptions, IHasMetadata
     {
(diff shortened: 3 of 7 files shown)


src/Stripe.net/Services/ExternalAccounts/ExternalAccountUpdateOptions.cs
```diff
@@ -12,6 +13,7 @@ public class ExternalAccountUpdateOptions : BaseOptions, IHasMetadata
         public bool? DefaultForCurrency { get; set; }
 
         [JsonProperty("external_account")]
+        [AllowNameMismatch]
         public ExternalAccountCardUpdateOptions ExternalAccountCard { get; set; }
```
remi-stripe: ooc couldn't you use your new magic thing that takes N parameters for the same property name instead?

src/Stripe.net/Services/Persons/PersonSharedOptions.cs
```diff
@@ -65,6 +64,7 @@ public abstract class PersonSharedOptions : BaseOptions, IHasMetadata
         public PersonRelationshipOptions Relationship { get; set; }
 
         [JsonProperty("ssn_last_4")]
+        [AllowNameMismatch]
         public string SSNLast4 { get; set; }
```
remi-stripe: Should this have been `SsnLast4` instead?

src/Stripe.net/Services/Sources/SourceThreeDSecureCreateOptions.cs
```diff
@@ -1,13 +1,15 @@
...
+    using Stripe.Infrastructure;
 
     public class SourceThreeDSecureCreateOptions : INestedOptions
     {
         [JsonProperty("customer")]
         public string Customer { get; set; }
 
         [JsonProperty("card")]
+        [AllowNameMismatch]
         public string CardId { get; set; }
```
remi-stripe: Why not rename to `Card` here?

src/Stripe.net/Services/Persons/PersonSharedOptions.cs
ob-stripe: Probably, though that's not the reason for the mismatch: both `SsnLast4` and `SSNLast4` become `ssn_last4` instead of `ssn_last_4` when converted to snake_case.

remi-stripe: ah yeah makes sense, let's leave it in that case

src/Stripe.net/Services/Sources/SourceThreeDSecureCreateOptions.cs
ob-stripe: 🤷‍♂  It's 3DS and you didn't change it in #1738 so I assumed you just didn't want to change the API. But I'm happy to change it if you think it's preferable.

src/Stripe.net/Services/ExternalAccounts/ExternalAccountUpdateOptions.cs
ob-stripe: Nah, it doesn't work like that. `AnyOf<>` lets you wrap multiple types, not multiple names.

src/Stripe.net/Services/Sources/SourceThreeDSecureCreateOptions.cs
remi-stripe: Yeah this one is worth changing, I must have missed it. It's on Sources which is still used so if you are willing to take that's likely better

src/Stripe.net/Services/ExternalAccounts/ExternalAccountUpdateOptions.cs
remi-stripe: Sure I know. I'm saying we should remove `ExternalAccountCard` and `ExternalAccountBankAccount` and use `AnyOf` + `ExternalAccount` no?

src/Stripe.net/Services/Sources/SourceThreeDSecureCreateOptions.cs
ob-stripe: 👍  Fixed in #1805.

src/Stripe.net/Services/ExternalAccounts/ExternalAccountUpdateOptions.cs
ob-stripe: Looks like `ExternalAccountCard` never worked anyway. Fixed in #1806.

ob-stripe:
ptal @remi-stripe

PR #1812, opened 2019-10-10
Remove the ThreeDSecure property on Card which is not in the spec

remi-stripe:
This property can sometimes be a hash depending on the account's settings causing deserialization to sometimes fail. Since it's not in the spec anymore and has been removed from stripe-java let's remove it here too to be consistent.

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Released as 31.0.0.

PR #1825, opened 2019-10-28
Do not use DefaultSettings when serializing or deserializing JSON

ob-stripe:
cc @stripe/api-libraries 

Newtonsoft.Json lets user define global default settings for serializing and deserializing JSON via the [`JsonConvert.DefaultSettings`](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonConvert_DefaultSettings.htm) property.

Those settings are used when serializing/deserializing using the static methods in `JsonConvert`. Even if you provide a `JsonSerializerSettings` instance, those settings will be merged with the default settings.

Because we were using `JsonConvert.SerializeObject` / `JsonConvert.DeserializeObject` in a number of places in the library, this means that users could affect the serialization/deserialization by setting `JsonConvert.DefaultSettings`. I don't think that's desirable: the library should use its own settings, independent of the global default settings.

This PR defines new utility methods `JsonUtils.SerializeObject` / `JsonUtils.DeserializeObject` which are exactly the same as the ones provided by Newtonsoft.Json's `JsonConvert`, except that the global default settings are not used (practically speaking, the difference is that the `JsonSerializer` class is instantiated with `JsonSerializer.Create(settings)` instead of `JsonSerializer.CreateDefault(settings)`.

Here's the list of places where we were using Newtonsoft.Json's `JsonConvert` methods and are now using our own `JsonUtils` methods:
- deserialization:
    - `StripeEntity.FromJson` (both generic and non-generic variants)
        - we use our own `StripeConfiguration.SerializerSettings` instance here, so users are still able to customize the deserialization by modifying this property if they need to
    - `EventUtility.ParseEvent`
        - same as above
- serialization:
    - `StripeEntity.ToJson`
        - same as above, we use `StripeConfiguration.SerializerSettings` so users are still able to customize the serialization
    - `FormEncoder.FlattenParamsValue`:
        - this is kind of a hack to encode enum values to their string representation. There's really no reason to modify the behavior here, so we're not providing any settings instance
    - `RequestTelemetry.MaybeAddTelemetryHeader` and `SystemNetHttpClient.BuildStripeClientUserAgentString`:
        - we want the JSON strings to be as "standard" as possible here so not passing any settings instance either

I'm tagging this as `breaking-api-change` because it may affect users who are modifying `JsonConvert.DefaultSettings` and who may know need to modify `StripeConfiguation.SerializerSettings` instead, but the vast majority of users should not be affected.

(For context, what motivated this PR is that I noticed some malformed telemetry headers on the backend:
```json
{"$id":"1","last_request_metrics":{"$id":"2","request_id":"req_123","request_duration_ms":234}}
```

The `$id` keys are inserted because the user set `PreserveReferencesHandling` to `PreserveReferencesHandling.All` in `JsonConvert.DefaultSettings`.)

### DIFF
--- a/src/Stripe.net/Entities/_base/StripeEntity.cs
+++ b/src/Stripe.net/Entities/_base/StripeEntity.cs
@@ -4,6 +4,7 @@ namespace Stripe
     using System.Reflection;
     using System.Runtime.CompilerServices;
     using Newtonsoft.Json;
+    using Stripe.Infrastructure;
 
     [JsonObject(MemberSerialization.OptIn)]
     public abstract class StripeEntity : IStripeEntity
@@ -19,7 +20,7 @@ namespace Stripe
         /// <returns>The deserialized Stripe object from the JSON string.</returns>
         public static IHasObject FromJson(string value)
         {
-            return JsonConvert.DeserializeObject<IHasObject>(value, StripeConfiguration.SerializerSettings);
+            return JsonUtils.DeserializeObject<IHasObject>(value, StripeConfiguration.SerializerSettings);
         }
 
         /// <summary>Deserializes the JSON to the specified Stripe object type.</summary>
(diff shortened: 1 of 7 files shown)


remi-stripe:
approving but flagging your tests are failing!

brandur-stripe:
Late to the party, but for the record: huge +1 on this. Eliminates an entire class of annoying-to-everyone, difficult-to-debug bug.

Issue #1848, opened 2019-11-25
ListAutoPaging doesn't work with EndingBefore if there is more than one page


sandfield-andy:
I was trying to use the auto pagination functionality to pull all balance transactions after a specific transaction by making use of the EndingBefore property in the list options.

Retrieving the first page works fine but if there is more than one page it seems ListAutoPaging sets the StartingAfter property in the options object I've passed through which then throws an exception stating the following: 

> Received both starting_after and ending_before parameters. Please pass in only one.

I would either expect it to prevent you from passsing ending_before to begin with or handle it.

Here's an example of the code:

```
BalanceTransactionService service = new BalanceTransactionService();
RequestOptions requestOptions = new RequestOptions
{
	StripeAccount = _stripeAccountID
};
BalanceTransactionListOptions balanceOptions = new BalanceTransactionListOptions
{
	EndingBefore = _lastStripeTransactionID,
	Limit = 100 // This is the maximum allowed.
};

// service.ListAutoPaging results in >100 transactions
foreach (BalanceTransaction transaction in service.ListAutoPaging(balanceOptions, requestOptions))
{
	// Process transaction
}
```

Let me know if you need any further information.

remi-stripe:
@sandfield-andy At the moment we don't support auto-pagination with `ending_before` which is why you're hitting this error. For now, the easiest is to paginate as usual (from the most recent objects first) and stop once you see the id you were looking for.

I'm going to tag as future since this is something we want to add support for in the future.

remi-stripe:
@sandfield-andy We just release a fix in 33.8.0 that supports this!

sandfield-andy:
Awesome, thanks @remi-stripe!

PR #1854, opened 2019-11-26
Add support for auto paging with ending_before

cjavilla-stripe:
Looking for feedback on this solution re #1848 

## Reviewers
r? @ob-stripe 
cc @stripe/api-libraries @remi-stripe 

## Summary

Adds support for auto paging with `ending_before` by iterating backwards through pages. Currently, we only support iterating forwards and are forcing the `starting_after` option to be used. 

`ending_before` and `starting_after` are mutually exclusive, so when attempting to pass `EndingBefore` with auto paginate, it fails because we default to setting `StartingAfter` to retrieve the "next page."

We don't have a great "get previous page" or "get next page" construct in dotnet, yet. 

I followed this for reference: 

https://github.com/stripe/stripe-ruby/pull/865/files

### DIFF
--- a/src/Stripe.net/Services/_base/Service.cs
+++ b/src/Stripe.net/Services/_base/Service.cs
@@ -252,8 +252,23 @@ namespace Stripe
                 options,
                 requestOptions);
 
+            options = options ?? new ListOptions();
+            bool iterateBackward = false;
+
+            // Backward iterating activates if we have an `EndingBefore`
+            // constraint and not a `StartingAfter` constraint
+            if (!string.IsNullOrEmpty(options.EndingBefore) && string.IsNullOrEmpty(options.StartingAfter))
+            {
+                iterateBackward = true;
+            }
+
             while (true)
             {
+                if (iterateBackward)
+                {
+                    page.Reverse();
+                }
+
                 string itemId = null;
                 foreach (var item in page)
                 {
@@ -275,8 +290,14 @@ namespace Stripe
                     break;
                 }
 
-                options = options ?? new ListOptions();
-                options.StartingAfter = itemId;
+                if (iterateBackward)
+                {
+                    options.EndingBefore = itemId;
+                }
+                else
+                {
+                    options.StartingAfter = itemId;
+                }
 
                 page = this.Request<StripeList<T>>(
                     HttpMethod.Get,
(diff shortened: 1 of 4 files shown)


cjavilla-stripe:
Tagging @remi for a first pass :)

src/Stripe.net/Services/_base/Service.cs
```diff
@@ -252,8 +252,23 @@ protected TEntityReturned CreateEntity(BaseOptions options, RequestOptions reque
...
+            if (!string.IsNullOrEmpty(options.EndingBefore) && string.IsNullOrEmpty(options.StartingAfter))
+            {
+                iterateBackward = true;
+            }
+
             while (true)
             {
+                if (iterateBackward)
+                {
+                    page.Reverse();
```
remi-stripe: I'm not sure why you have to reverse the whole page instead of plucking from the first element in the case of `EndingBefore`?

src/Stripe.net/Services/_base/Service.cs
```diff
@@ -252,8 +252,23 @@ protected TEntityReturned CreateEntity(BaseOptions options, RequestOptions reque
                 options,
                 requestOptions);
 
+            options = options ?? new ListOptions();
+            bool iterateBackward = false;
+
+            // Backward iterating activates if we have an `EndingBefore`
+            // constraint and not a `StartingAfter` constraint
+            if (!string.IsNullOrEmpty(options.EndingBefore) && string.IsNullOrEmpty(options.StartingAfter))
```
remi-stripe: What if you want to paginate between 2 ids? I never fully understand how they relate to each other.

cjavilla-stripe: Our API raises if you pass both. 🤷‍♂

src/Stripe.net/Services/_base/Service.cs
cjavilla-stripe: I reversed because this would make the items in order, and it's also what we're doing in Ruby. 

Because we're iterating backwards, the actual page comes in in order.

Without reversing the page, assuming ID's are incrementing. Paging with the Starting After (with pages of size 3) would look like:

10
9
8
7
6
5
4
3
2
1

And paging with EndingBefore:

3
2
1
4
5
6
7
8
9
10

By reversing the underlying list, when you use Ending before, you get
1
2
3
4
5
6
7
8
9
10

PR #1869, opened 2019-12-05
New public method EventUtility.ValidateSignature

clement911:
New public method `EventUtility.ValidateSignature` to validate a webhook payload without parsing the event.

Rationale:
When we receive a webhook, we want to do the following steps:
-Verify the signature
-Save the payload in our database
-Return HTTP 200

This is a best practice to return a response to Stripe immediately and save the webhook for later processing.

We **don't want to parse the event** when receiving the webhook.
-It's an added perf cost that is not necessary
-In addition, if we have changed our account API version, our webhook may not be able to parse the events, until we update it with the matching version of Stripe.net 

I hope it all makes sense?

### DIFF
--- a/src/Stripe.net/Services/Events/EventUtility.cs
+++ b/src/Stripe.net/Services/Events/EventUtility.cs
@@ -15,6 +15,8 @@ namespace Stripe
         internal static readonly UTF8Encoding SafeUTF8
             = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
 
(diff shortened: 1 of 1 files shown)


clement911:
This is related to https://github.com/stripe/stripe-dotnet/issues/1874
Specifically whether the parsing of the event can fail if there is a version mismatch.

ob-stripe:
Released as 34.2.0.

Thanks again! :)

clement911:
Thank you for merging this!

PR #1877, opened 2019-12-12
Add support for `AuthorizationControls` on Issuing `Cardholder`

remi-stripe:
New version of https://github.com/stripe/stripe-dotnet/pull/1576

r? @ob-stripe 
cc @stripe/api-libraries


the800kid:
@remi-stripe,
I have noticed that the authorization spending limit option was only added to the Cardholder.  Is there a reason why the spending limit was not added to the Card level too?

PR #1881, opened 2019-12-13
Exposing the StripeTypeRegistry.ObjectsToTypes publicly

clement911:
We are building some highly generic stripe components internally that will need a mapping from object to type.
Rather than maintaining our own, we'd prefer to leverage the one already existing in the library.
I wrapped the dictionary in a read only dictionary so that it can't be modified by consumers.


clement911:
Sure it makes sense.
All done.

ob-stripe:
Thanks! We'll merge and release this soon.

ob-stripe:
Released as 34.6.0.

clement911:
Thank you this is very useful!

PR #1885, opened 2019-12-18
Public Service.Client setter

clement911:
We would love this change to build more generic code to `new` up services and set the `Client`, without knowing the concrete type of the service.

This might also be useful if services are dependency injected but may need to use different `IStripeClient` depending on context.

### DIFF
--- a/src/Stripe.net/Services/_base/Service.cs
+++ b/src/Stripe.net/Services/_base/Service.cs
@@ -41,10 +41,14 @@ namespace Stripe
         public virtual string BaseUrl => this.Client.ApiBase;
 
         /// <summary>
-        /// Gets the client used by this service to send requests. If no client was set when the
+        /// Gets or sets the client used by this service to send requests. If no client was set when the
         /// service instance was created, then the default client in
         /// <see cref="StripeConfiguration.StripeClient"/> is used instead.
         /// </summary>
+        /// <remarks>
+        /// Setting the client at runtime may not be thread-safe.
+        /// If you wish to use a custom client, it is recommended that you pass it to the service's constructor and not change it during the service's lifetime.
+        /// </remarks>
         public IStripeClient Client
         {
             get => this.client ?? StripeConfiguration.StripeClient;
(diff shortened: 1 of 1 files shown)


ob-stripe:
The client can already be injected in the [constructor](https://github.com/stripe/stripe-dotnet/blob/844b632303638e4efa3f40a336e71e1d16f5dc3c/src/Stripe.net/Services/_base/Service.cs#L34), and I'm a bit worried that setting the client post-construction might not be thread safe. 

Could you provide more information about how you'd use this setter and why injecting the client at construction is not an option?

clement911:
Hi @ob-stripe,

I did mention that it could be useful for dependency injection, but the main use case is actually building some generic code.

For example, imagine a class similar to this:

```
public class MyGenericClass<TStripeService, TListOptions, TStripeEntity>
        where TStripeService : IListable<TStripeEntity, TListOptions>, new()
        where TListOptions : ListOptions, new()
        where TStripeEntity : IStripeEntity, IHasId
    {
        public void ProcessEntities(IStripeClient client)
        {
            var svc = new TStripeService();
            svc.Client = client;

            var svcOptions = new TListOptions();

            foreach (var entity in svc.ListAutoPaging(svcOptions))
            {
                //do something
            }
        }
    }
```

This is why we could benefit from a `Client` setter.

Note that we need to use different API keys so we have different `StripeClient`s and can't retrieve it from `StripeConfiguration` as it assumes a single API key.

Having different API keys is also problematic for dependency injection.
Which of the `StripeClient` should be injected in the constructor? It depends. Having a setter allows us to set the correct `StripeClient` instance at runtime.

I hope this makes sense?

ob-stripe:
Hm, ok. I'm still not sure you couldn't achieve this just by creating a new service and injecting the client via the service's constructor, but I'm not opposed to this change provided we give appropriate warnings.

Could you make the following doc changes?

1. Update the method's `<summary>` block to say "Gets or sets" instead of just "Gets"
2. Add a `<remarks>` block with something to that effect:
```
Setting the client at runtime may not be thread-safe. If you wish to use a custom client, it is recommended that you pass it to the service's constructor and not change it during the service's lifetime.
```

clement911:
Sure, it's all done.

ob-stripe:
Thanks!

Released as 34.10.0.

PR #1894, opened 2020-01-10
Fix sepa debit deserialization on MandatePaymentMethodDetails

matthewling-stripe:
While testing the MandateService, I noticed that I could not retrieve SEPA debit information. On further inspection, it seems that SEPA debit data was not being deserialized properly. This change should allow the `reference` and `url` attributes to be surfaced via the MandateService and Mandate Entity.


remi-stripe:
cc @stripe/api-libraries for awareness

PR #1896, opened 2020-01-12
Add support for `CreditNoteLineItem`

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Services/CreditNotes/CreditNoteLineOptions.cs
+++ b/src/Stripe.net/Services/CreditNotes/CreditNoteLineOptions.cs
@@ -1,0 +1,73 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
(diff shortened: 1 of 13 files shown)


src/Stripe.net/Services/CreditNotes/CreditNoteLineOptions.cs
```diff
@@ -0,0 +1,65 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class CreditNoteLineOptions : INestedOptions
```
cjavilla-stripe: Should this have metadata?

remi-stripe: it should, fixed!

PR #1902, opened 2020-01-22
Make CreditNote OutOfBandAmount nullable fixes #1901

### LINKED ISSUE #1901: CreditNote out_of_band_amount should be nullable
Trying to retrieve a list of credit notes through the CreditNoteService.ListAsync method I am receiving an error. The initial exception only says there was an 'Invalid response object from API'
 but upon further investigation I can see that I am trying to retrieve a credit note with a null out_of_band_amount. `"out_of_band_amount": null`

The OutOfBandAmount is set as `long` on the [entity,](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Entities/CreditNotes/CreditNote.cs#L155) so the Json deserialisation is failing.

Could you please allow a null value on the OutOfBandAmount so that this list of Credit Notes can be retrieved? Or is this an error on the API side and that null should actually be a zero?

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

fixes #1901

### DIFF
--- a/src/Stripe.net/Entities/CreditNotes/CreditNote.cs
+++ b/src/Stripe.net/Entities/CreditNotes/CreditNote.cs
@@ -152,7 +152,7 @@ namespace Stripe
         /// Amount that was credited outside of Stripe.
         /// </summary>
         [JsonProperty("out_of_band_amount")]
-        public long OutOfBandAmount { get; set; }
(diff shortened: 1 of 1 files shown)


ob-stripe:
Released as 34.16.1.

PR #1908, opened 2020-02-04
Add support for returning orders

cjavilla-stripe:
While deprecated, this is currently available in the API so we should support it.

This is my first time adding a custom method to a service like this where the return type is different than the underlying entity. I.e. returning an order, returns an OrderReturn object, I think I did this right, but could use another set of eyes. Another note, I created OrderReturnItemOptions as a new type because OrderItemOptions has the additional `currency` field. LMK if that was the wrong move.

r? @remi-stripe 
cc @stripe/api-libraries @ob-stripe

### DIFF
--- a/src/Stripe.net/Services/Orders/OrderReturnOptions.cs
+++ b/src/Stripe.net/Services/Orders/OrderReturnOptions.cs
@@ -1,0 +1,14 @@
+namespace Stripe
+{
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class OrderReturnOptions : BaseOptions
+    {
+        /// <summary>
+        /// List of items to return.
+        /// </summary>
+        [JsonProperty("items")]
+        public List<OrderReturnItemOptions> Items { get; set; }
+    }
+}
--- a/src/Stripe.net/Services/Orders/OrderService.cs
+++ b/src/Stripe.net/Services/Orders/OrderService.cs
@@ -69,6 +69,16 @@ namespace Stripe
             return this.RequestAsync(HttpMethod.Post, $"{this.InstanceUrl(orderId)}/pay", options, requestOptions, cancellationToken);
         }
 
+        public virtual OrderReturn Return(string orderId, OrderReturnOptions options = null, RequestOptions requestOptions = null)
+        {
+            return this.Request<OrderReturn>(HttpMethod.Post, $"{this.InstanceUrl(orderId)}/returns", options, requestOptions);
+        }
+
+        public virtual Task<OrderReturn> ReturnAsync(string orderId, OrderReturnOptions options = null, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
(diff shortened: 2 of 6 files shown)


src/StripeTests/Services/Orders/OrderServiceTest.cs
```diff
@@ -38,22 +39,34 @@ public class OrderServiceTest : BaseStripeTest
                 },
             };
 
-            this.updateOptions = new OrderUpdateOptions
+            this.listOptions = new OrderListOptions
```
cjavilla-stripe: I alphabetized the initialization here to help with autogen and added the new returnOptions

src/Stripe.net/Services/Orders/OrderReturnOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+
+    public class OrderReturnOptions : BaseOptions
+    {
+        /// <summary>
+        /// List of items to return.
+        /// </summary>
+        [JsonProperty("items")]
+        public List<OrderReturnItemOptions> Items { get; set; }
```
remi-stripe: This naming feels like a new pattern. Usually we don't put the action/word name in the options class. But if we don't here, it conflicts with the one for Create which is the same hash but wit `currency` on top of it.

@ob-stripe Can you check and confirm that's the right naming scheme?

src/Stripe.net/Services/Orders/OrderService.cs
```diff
@@ -69,6 +69,16 @@ public virtual Task<Order> PayAsync(string orderId, OrderPayOptions options = nu
             return this.RequestAsync(HttpMethod.Post, $"{this.InstanceUrl(orderId)}/pay", options, requestOptions, cancellationToken);
         }
 
+        public virtual OrderReturn Return(string orderId, OrderReturnOptions options = null, RequestOptions requestOptions = null)
```
remi-stripe: In other languages like PHP the method is named `returnOrder` instead. I think the short name is better if supported in dotnet but also flagging to @ob-stripe

cjavilla-stripe: Good call out. I followed the other custom method in this file called `Pay` but happy to name it whatever.

cjavilla-stripe:
Thanks @remi-stripe, I reviewed Order and OrderReturn against the API ref and they have all the fields listed (added some of the doc comments).

r? @ob-stripe Can you take a pass at this one?

src/Stripe.net/Services/Orders/OrderReturnOptions.cs
ob-stripe: Yeah, I think so. We're just "lucky" that in most cases the shape is shared between e.g. create and update methods, but more generally we probably shouldn't be sharing parameters, at least until the OpenAPI spec actually gives us that information.

src/Stripe.net/Services/Orders/OrderService.cs
ob-stripe: 👍  for `Return`

PR #1911, opened 2020-02-07
Renaming a couple directories and files to follow convention

cjavilla-stripe:
Renaming a couple directories to follow a convention that we'll use for autogen this has no functional impact on the library, just moving things around so it's more standardized.

r? @remi-stripe 
cc @stripe/api-libraries


ob-stripe:
Nit, but I think we should special case `Balance` and `ThreeDSecure` to not be pluralized.

cjavilla-stripe:
Okay will do

cjavilla-stripe:
Updated Balance and ThreeDSecure to be singular.

PR #1912, opened 2020-02-08
Rename all parameters in services' methods to be consistent

remi-stripe:
This is an important step to make codegen easier to review. The idea is that the resource id will be called `id` and the parent's id `parentId`. That way codegen doesn't have to try and be smart about each parameter name and when we look at the diff with the real services files it should add less confusion.

This is technically a breaking change because in C# we have named arguments and someone could technically be passing arguments that way so it will go in the integration branch

r? @ob-stripe 
cc @stripe/api-libraries @cjavilla-stripe

### DIFF
--- a/src/Stripe.net/Services/Accounts/AccountService.cs
+++ b/src/Stripe.net/Services/Accounts/AccountService.cs
@@ -35,24 +35,24 @@ namespace Stripe
             return this.CreateEntityAsync(options, requestOptions, cancellationToken);
         }
 
-        public virtual Account Delete(string accountId, RequestOptions requestOptions = null)
+        public virtual Account Delete(string id, RequestOptions requestOptions = null)
         {
-            return this.DeleteEntity(accountId, null, requestOptions);
+            return this.DeleteEntity(id, null, requestOptions);
         }
 
-        public virtual Task<Account> DeleteAsync(string accountId, RequestOptions requestOptions = null, CancellationToken cancellationToken = default(CancellationToken))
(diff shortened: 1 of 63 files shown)


remi-stripe:
For completeness, most of it was done by a few regexp and then some manual fixes:

First replace all top-level CRUD methods

Input:
```
.*(Get|Delete|Update)(Async)*)\(string [a-zA-Z]+Id,( [^s].*)
(.*{.*)
(.*(Get|Delete|Update)Entity(Async)*)\([a-zA-Z]+Id,
```
Replaced by:
```
$1(string id,$4
$5
$6(id,
```

Then replace nested RUD methods
```
(.*(Get|Delete|Update)(Async)*)\(string [a-zA-Z]+Id, string [a-zA-Z]+Id,( .*)
(.*{.*)
(.*(Get|Delete|Update)NestedEntity(Async)*)\([a-zA-Z]+Id, [a-zA-Z]*Id,
```
Replaced by
```
$1(string parentId, string id,$4
$5
$6(parentId, id,
```

Then replace nested Create
```
(.*(Create)(Async)*)\(string [a-zA-Z]+Id,( [^s].*)
(.*{.*)
(.*(Create)NestedEntity(Async)*)\([a-zA-Z]+Id,
```
Replaced by
```
$1(string parentId,$4
$5
$6(parentId,
```

Then replace nested List (async and autopaging)
```
(.*(List)(Async|Autopaging)*)\(string [a-zA-Z]+Id,( [^s].*)
(.*{.*)
(.*(List)NestedEntities(Async|Autopaging)*)\([a-zA-Z]+Id,
```
Replaced by
```
$1(string parentId,$4
$5
$6(parentId,
```

PR #1919, opened 2020-02-12
Use a static instance for default `HttpClient`

### LINKED ISSUE #1918: Following best-practices for HttpClient
In the latest version of this library, creating a `StripeClient` with the minimal constructor creates a new `SystemNetHttpClient`, which in turn creates a new `HttpClient`.

In some cases, multiple `StripeClient`s need to be created in an application (e.g. when dealing with multiple Stripe accounts, each with a different API key).

The "best practices" for HttpClient instances are [mentioned by Microsoft](https://docs.microsoft.com/en-us/aspnet/web-api/overview/advanced/calling-a-web-api-from-a-net-client#create-and-initialize-httpclient) as follows: 

> HttpClient is intended to be instantiated once and reused throughout the life of an application.

My questions:
* Why isn't `SystemNetHttpClient` using a static, application-wide `HttpClient` instance?
* When passing a custom `SystemNetHttpClient` instance to `StripeClient`, is there any danger in using a static, application-wide `HttpClient` instance?

ob-stripe:
r? @brandur-stripe @richardm-stripe 
cc @stripe/api-libraries 

Update `Stripe.SystemNetHttpClient` to use a static instance of `System.Net.HttpClient` when no HTTP client is passed to the constructor. This means that multiple clients will share the same instance of `System.Net.HttpClient`.

This is what Microsoft advises in https://docs.microsoft.com/en-us/aspnet/web-api/overview/advanced/calling-a-web-api-from-a-net-client#create-and-initialize-httpclient:
> **HttpClient** is intended to be instantiated once and reused throughout the life of an application.

This is thread-safe. The only method we're using is `SendAsync`, which is explicitly listed as thread-safe in the documentation: https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?redirectedfrom=MSDN&view=netframework-4.8#remarks.

The default instance is wrapped in a [`Lazy<>`](https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1) initializer, so it's only initialized the first time it's needed (to avoid initializing it if it's never used, e.g. because the user is always supplying their own instance).

It's difficult to write a test for this because `httpClient` is private, but I temporarily bumped its visibility and verified that this test passed:
```c#
var client1 = new SystemNetHttpClient();
var client2 = new SystemNetHttpClient();

Assert.NotSame(client1, client2);
Assert.Same(client1.HttpClient, client2.HttpClient);
```

Fixes #1918.

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
+++ b/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
@@ -21,6 +21,9 @@ namespace Stripe
     /// </summary>
     public class SystemNetHttpClient : IHttpClient
     {
+        private static readonly Lazy<System.Net.Http.HttpClient> LazyDefaultHttpClient
+            = new Lazy<System.Net.Http.HttpClient>(BuildDefaultSystemNetHttpClient);
+
         private readonly System.Net.Http.HttpClient httpClient;
 
         private readonly int maxNetworkRetries;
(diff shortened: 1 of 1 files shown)


ob-stripe:
Thanks Brandur!

PR #1923, opened 2020-02-17
Add support for the new file purpose additional verification

martineo14:
After adapting my project to the new version of stripe connect we found that the new file purpuse `additional_verification` is not supported in the class FilePurpose part of the Constants folder.

#### Our actual implementation with version 34.16.0

image

#### We would like for the new version 😉

image

Thanks.


remi-stripe:
Just merged and released as 34.23.0

PR #1930, opened 2020-02-25
Removing deprecated field.

clement911:
Is this a deprecated field? It doesn't appear to be documented.


remi-stripe:
@clement911 This is not a deprecated field and is supported in the API today. It's just gated for specific users for now and will possibly be public in the future!
I'm going to close the PR as we wouldn't merge it, I hope it's okay!

clement911:
Got it thanks.

Issue #1931, opened 2020-02-27
[Feature Request] Delete SubscriptionItem should take options to set `prorate = false`


JessePerry:
This is a feature request or a missing feature.
Create and Update SubscriptionItem can be called with options like this:
```
await subscriptionItemService.UpdateAsync(subItem.Id, new SubscriptionItemUpdateOptions { Quantity = requestedQuantity, Prorate = _isProrate });
```
But Delete has no such options object or method. So if you delete a sub item with this api client, it will always prorate=true.

In the api docs, it's possible to give options to [delete ](https://stripe.com/docs/api/subscription_items/delete) 

Delete and DeleteAsync should take a parameter SubscriptionItemDeleteOptions to mimic the ability of Create/update and allow deletion with prorata = false.

remi-stripe:
@JessePerry Thanks a lot for raising this! Today, the majority of the Delete operations don't support passing any parameters and this endpoint is more an exception.

It seems we never fixed the library to properly reflect that. I'm going to work on a fix which will be straightforward but it will be a breaking change so it will be released in the next major version that is coming out next week.

remi-stripe:
Okay PR in progress here https://github.com/stripe/stripe-dotnet/pull/1932 if you want to follow along!

ob-stripe:
This was fixed in v35.0.0.

PR #1932, opened 2020-02-27
Add support for passing parameters when deleting a Subscription Item

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/1931

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Services/SubscriptionItems/SubscriptionItemDeleteOptions.cs
+++ b/src/Stripe.net/Services/SubscriptionItems/SubscriptionItemDeleteOptions.cs
@@ -1,0 +1,36 @@
+namespace Stripe
+{
+    using System;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class SubscriptionItemDeleteOptions : BaseOptions
+    {
+        /// <summary>
+        /// Delete all usage for the given subscription item. Only allowed when the current plan’s
+        /// <c>usage_type</c> is metered.
+        /// </summary>
+        [JsonProperty("clear_usage")]
+        public bool? ClearUsage { get; set; }
+
+        /// <summary>
+        /// Determines how to handle
+        /// <a href="https://stripe.com/docs/billing/subscriptions/billing-cycle#prorations">prorations</a>
+        /// when the billing cycle changes. The value defaults to
+        /// <c>create_prorations</c>, indicating that proration invoice items should
+        /// be created. Prorations can be disabled by setting the value to
+        /// <c>none</c>.
+        /// </summary>
+        [JsonProperty("proration_behavior")]
+        public string ProrationBehavior { get; set; }
+
+        /// <summary>
+        /// If set, the proration will be calculated as though the subscription was updated at the
+        /// given time. This can be used to apply the same proration that was previewed with the
+        /// upcoming invoice endpoint.
+        /// </summary>
+        [JsonProperty("proration_date")]
+        [JsonConverter(typeof(DateTimeConverter))]
+        public DateTime? ProrationDate { get; set; }
+    }
+}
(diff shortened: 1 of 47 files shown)


src/Stripe.net/Services/SubscriptionItems/SubscriptionItemDeleteOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <summary>
+        /// Determines how to handle
+        /// <a href="https://stripe.com/docs/billing/subscriptions/billing-cycle#prorations">prorations</a>
+        /// when the billing cycle changes. The value defaults to
+        /// <c>create_prorations</c>, indicating that proration invoice items should
+        /// be created. Prorations can be disabled by setting the value to
+        /// <c>none</c>.
+        /// </summary>
+        [JsonProperty("proration_behavior")]
+        public string ProrationBehavior { get; set; }
```
remi-stripe: Not yet supported but will be by the time we merge, I don't want to add `prorate` since it will be deprecated at the same time.

src/Stripe.net/Services/SubscriptionItems/SubscriptionItemService.cs
```diff
@@ -6,7 +6,7 @@ namespace Stripe
 
     public class SubscriptionItemService : Service<SubscriptionItem>,
         ICreatable<SubscriptionItem, SubscriptionItemCreateOptions>,
-        IDeletable<SubscriptionItem>,
+        IDeletableWithParams<SubscriptionItem, SubscriptionItemDeleteOptions>,
```
remi-stripe: This seemed the least intrusive. The alternative is to change all `Delete` to take params by default even though this is the only method today that supports this (and Terminal objects for legacy reasons that is unsupported anyway).

What do you think is best here?

ob-stripe: I'm not sure there's a lot of value in adding a different interface for just this one service. I think we should just remove the `IDeletable` interface from the service, and in a future major version update the interface to accept params (which means creating parameter classes for all delete methods), like we did for the `IRetrievable` interface a while back.

remi-stripe: Fair. Let's ignore that part and answer the real question of: would we move all delete methods to take an empty options hash for that one exception?

ob-stripe: I'm personally in favor. The user impact is limited (just need to update callsites to `Delete(id, requestOptions)` to be either `Delete(id, null, requestOptions)` or `Delete(id, requestOptions: requestOptions)`) and it makes the API more consistent and future proof.

remi-stripe:
@ob-stripe I added parameters for all Delete methods. That PR also adds the `SubscriptionItemDeleteOptions` parameters. The API doesn't yet support `proration_behavior` but it's being built right now

src/Stripe.net/Services/_interfaces/IDeletable.cs
```diff
@@ -3,11 +3,12 @@ namespace Stripe
     using System.Threading;
     using System.Threading.Tasks;
 
-    public interface IDeletable<TEntity>
+    public interface IDeletable<TEntity, TOptions>
         where TEntity : IStripeEntity, IHasId
+        where TOptions : BaseOptions, new()
     {
-        TEntity Delete(string id, RequestOptions requestOptions = null);
+        TEntity Delete(string id, TOptions options, RequestOptions requestOptions = null);
```
ob-stripe: Mind making the argument optional with `TOptions options = null`, and doing the same thing for the async method? And update `IRetrievable`/`INestedRetrievable` to do the same thing while you're at it?

This would allow the following invocations to work:

```c#
SubscriptionItemService s1 = new SubscriptionItemService();
s1.Delete(id); // works already, because `options` is optional in `SubscriptionItemService.Delete()`

IDeletable s2 = (IDeletable) s1;
s2.Delete(id); // doesn't work currently, because `options` is not optional in `IDeletable.Delete()`
```

remi-stripe:
Fixed, PTAL @ob-stripe

PR #1933, opened 2020-02-28
Drop support for .NET Standard 1.2

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Drop support for .NET Standard 1.2.

### DIFF
--- a/src/StripeTests/StripeTests.csproj
+++ b/src/StripeTests/StripeTests.csproj
@@ -2,7 +2,7 @@
   <Import Project="..\..\netfx.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.1;netcoreapp2.1;netcoreapp3.0;net452</TargetFrameworks>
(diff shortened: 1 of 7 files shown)


src/StripeTests/StripeTests.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\netfx.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.1;netcoreapp2.1;netcoreapp3.0;net452</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0;netcoreapp3.1;net452</TargetFrameworks>
```
remi-stripe: why `netcoreapp2.0` suddenly when we had 2.1 before?

ob-stripe: I just want to make sure we're testing with the minimum version we're supporting. It shouldn't make any difference since .NET Core Runtime roughly adheres to semver, but better safe than sorry.

PR #1935, opened 2020-03-02
Remove conditional compilation macros in tests

ob-stripe:
cc @stripe/api-libraries 

- Remove all conditional compilation preprocessor macros in tests
- Change `net452` target framework in tests to `net45` (I can't remember why we were using `net452`, but tests pass with `net45` now so we might as well test with the minimum version we support)
- Bump `Moq` test dependency to latest version


ob-stripe:
Thanks CJ!

PR #1925, opened 2020-02-21
Add `stripe_net_target_framework` key to extended user agent

ob-stripe:
r? @remi-stripe 

Add `stripe_net_target_framework` key to `X-Stripe-Client-User-Agent` header.

PR #1936, opened 2020-03-03
Get accurate runtime version

ob-stripe:
r? @brandur-stripe @remi-stripe @richardm-stripe 
cc @stripe/api-libraries 

This is an attempt to finally get accurate runtime versions in the extended user agent.

Some context: getting the runtime version with .NET in a consistent manner that works across all different runtimes (Framework/Core/Mono...) and versions is ridiculously hard. It has become a lot easier in recent times (as of .NET Core 3.0, [`RuntimeInformation.FrameworkDescription`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.frameworkdescription?view=netframework-4.8) _finally_ returns a sane value), but unfortunately we need to support older versions. The whole point of having accurate runtime versions is to let us know when we can safely drop support for older versions.

Thankfully, the [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet) project has largely solved the problem by throwing a lot of code at it. I've been meaning to use their solution for a while now, and even [asked them](https://github.com/dotnet/BenchmarkDotNet/issues/1078) to consider releasing this specific part as a separate library (they said no).

Until recently we were unable to use their code because we were still supporting .NET Standard 1.2. We've dropped support for that version in this integration branch (cf. #1933) based on limited telemetry that I introduced in #1925, so we can now copy BenchmarkDotNet's solution.

The code is fairly complex because of the number of cases to handle, and is not especially pretty, but it works. I haven't written any tests for it because its very nature makes it hard to test, but I've manually verified that it reports sane values with .NET Core 2.0, .NET Core 3.1, Mono (on macOS) and .NET Framework 4.8 (on Windows, using Appveyor).

The code is largely copied from these files:
- https://github.com/dotnet/BenchmarkDotNet/blob/4740478f46f3c98d97d4a5b470260ebd7efa571b/src/BenchmarkDotNet/Portability/RuntimeInformation.cs
- https://github.com/dotnet/BenchmarkDotNet/blob/4740478f46f3c98d97d4a5b470260ebd7efa571b/src/BenchmarkDotNet/Helpers/FrameworkVersionHelper.cs
- https://github.com/dotnet/BenchmarkDotNet/blob/4740478f46f3c98d97d4a5b470260ebd7efa571b/src/BenchmarkDotNet/Environments/Runtimes/CoreRuntime.cs

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
+++ b/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
@@ -21,6 +21,16 @@ namespace Stripe
     /// </summary>
     public class SystemNetHttpClient : IHttpClient
     {
+        private const string StripeNetTargetFramework =
+#if NETSTANDARD2_0
+            "netstandard2.0"
+#elif NET45
+            "net45"
+#else
+            "unknown"
+#endif
+        ;
+
         private static readonly Lazy<System.Net.Http.HttpClient> LazyDefaultHttpClient
             = new Lazy<System.Net.Http.HttpClient>(BuildDefaultSystemNetHttpClient);
 
@@ -204,28 +214,10 @@ namespace Stripe
                 { "bindings_version", StripeConfiguration.StripeNetVersion },
                 { "lang", ".net" },
                 { "publisher", "stripe" },
-                { "lang_version", RuntimeInformation.GetLanguageVersion() },
+                { "lang_version", RuntimeInformation.GetRuntimeVersion() },
                 { "os_version", RuntimeInformation.GetOSVersion() },
+                { "stripe_net_target_framework", StripeNetTargetFramework },
             };
-
-#if NET45
-            string monoVersion = RuntimeInformation.GetMonoVersion();
-            if (!string.IsNullOrEmpty(monoVersion))
-            {
-                values.Add("mono_version", monoVersion);
-            }
-#endif
-
-            var stripeNetTargetFramework =
-#if NET45
-                "net45"
-#elif NETSTANDARD2_0
-                "netstandard2.0"
-#else
-                "unknown"
-#endif
-            ;
-            values.Add("stripe_net_target_framework", stripeNetTargetFramework);
 
             if (this.appInfo != null)
             {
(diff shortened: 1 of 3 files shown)


src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
```diff
@@ -204,29 +214,11 @@ private string BuildStripeClientUserAgentString()
                 { "bindings_version", StripeConfiguration.StripeNetVersion },
                 { "lang", ".net" },
                 { "publisher", "stripe" },
-                { "lang_version", RuntimeInformation.GetLanguageVersion() },
+                { "lang_version", RuntimeInformation.GetRuntimeVersion() },
```
remi-stripe: Is using the same property the right decision here? Won't it skew our results and make it harder to inspect?

src/Stripe.net/Infrastructure/RuntimeInformation.cs
```diff
@@ -1,53 +1,264 @@
...
+
+            return $".NET Framework {releaseVersion}";
         }
 
-#if NET45
-        public static string GetMonoVersion()
+        internal static string MapToReleaseVersion(string servicingVersion)
         {
-            Type monoRuntimeType = typeof(object).Assembly.GetType("Mono.Runtime");
+            // the following code assumes that .NET 4.5 is the oldest supported version
```
remi-stripe: I don't fully grasp why we do this instead of sending `servicingVersion` or sanitizing it ourselves?

src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
ob-stripe: The current value is more or less useless (it tells us the proportion of Framework vs. Core users, but no useful version numbers). I think we can start sending meaningful values and update our internal metrics to only look at the `lang_version` field when the bindings version is >= 35.0.

src/Stripe.net/Infrastructure/RuntimeInformation.cs
ob-stripe: The "servicing version" includes patch/build numbers that are not particularly relevant, so this method (adapted from [here](https://github.com/dotnet/BenchmarkDotNet/blob/4740478f46f3c98d97d4a5b470260ebd7efa571b/src/BenchmarkDotNet/Helpers/FrameworkVersionHelper.cs#L38-L53)) sanitizes the servicing version to an actual release version (https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies#version-information).

PR #1939, opened 2020-03-03
Add ip address property

mh884:
I'm trying to solve the issue which the Radar is keep blocking the Client IP address which the Stripe API is not sending from the Back-end i tried to add IP and the user_agent in the Call inside paymentMethods to get the Client IP address and the user Agent to be sent to the Stripe to be used in the Radar

I create this PR to get more clarify if this will do the job or not 

Request POST body
{
  "card": {
    "cvc": "***",
    "exp_month": "4",
    "exp_year": "2023",
    "number": "************5556"
  },
  "type": "card",
  "ip": "12.345.678.9",
  "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.86 Safari/537.36"
}

### DIFF
--- a/src/Stripe.net/Entities/PaymentMethods/PaymentMethod.cs
+++ b/src/Stripe.net/Entities/PaymentMethods/PaymentMethod.cs
@@ -70,5 +70,11 @@ namespace Stripe
 
         [JsonProperty("type")]
         public string Type { get; set; }
+
+        [JsonProperty("ip")]
+        public string Ip { get; set; }
+
+        [JsonProperty("user_agent")]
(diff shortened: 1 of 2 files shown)


remi-stripe:
@mh884 You shouldn't have to send an IP address and this is not a valid parameter in the API today so it wouldn't be added to the library. I'd recommend reaching out to our support team instead and they can help you diagnose what needs to happen: https://support.stripe.com/contact

PR #1947, opened 2020-03-06
Support for asynchronous autopagination

### LINKED ISSUE #1946: ListAutoPaging async version
I don't see why not?

What do you think?

ob-stripe:
r? @brandur-stripe @cjavilla-stripe @remi-stripe 
cc @stripe/api-libraries 

This PR adds support for asynchronous autopagination, using the new [`IAsyncEnumerable<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.iasyncenumerable-1?view=netcore-3.1) interface introduced in [C# 8](https://docs.microsoft.com/en-us/archive/msdn-magazine/2019/november/csharp-iterating-with-async-enumerables-in-csharp-8).

Normally, this would require .NET Standard 2.1, but we can use it without changing our target of .NET Standard 2.0 thanks to the [Microsoft.Bcl.AsyncInterfaces](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/) package. Unfortunately there is (AFAIK) no way to support this with .NET Framework 4.5, so I had to add a gazillion `#if !NET45` preprocessor macros.

This PR is very large and I should have done a better job of splitting it into smaller commits, but it shouldn't be too hard to review:
- the core of the implementation is in `Service.cs`, in the new `ListRequestAutoPagingAsync<T>` method: https://github.com/stripe/stripe-dotnet/pull/1947/files#diff-0de0406b0f5569892ed33a084f9759ceR275. The implementation is very similar to the old synchronous `ListRequestAutoPaging<T>` method. The main differences are:
    - it uses `RequestAsync` instead of `Request` to fetch the pages
    - it has an additional `CancellationToken` parameter, and there are a couple of `cancellationToken.ThrowIfCancellationRequested();` in the loops to check if we need to stop iterating
- `Service.cs` has a new `ListEntitiesAutoPagingAsync` helper method: https://github.com/stripe/stripe-dotnet/pull/1947/files#diff-0de0406b0f5569892ed33a084f9759ceR169, and similarly `ServiceNested.cs` has a new `ListNestedEntitiesAutoPagingAsync` helper method: https://github.com/stripe/stripe-dotnet/pull/1947/files#diff-efcfb49f6a90c39650225af44e8ad09fR143
- the `IListable` and `INestedListable` interfaces were updated with a new `ListAutoPagingAsync` signature
- every service was updated to add the new `ListAutoPagingAsync` methods

For tests:
- `AutoPagingTest.cs` was updated to add tests for the new async methods. `ListAutoPagingAsync`, ListAutoPagingAsync_NoParams`, `ListAutoPagingAsync_StartingAfter` and `ListAutoPagingAsync_EndingBefore` are copies of the tests for the synchronous methods. `ListAutoPagingAsync_WithCancellation` is a new test unique to the async version to test the cancellation behavior.
- Every `ListAutoPaging` test in service test classes was updated to use `First()` instead of `ToList()`. [`ToList()`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.tolist?view=netframework-4.8) forces the conversion of the `IEnumerable` into a `List`, which could cause many requests if there were many pages to fetch. This is not actually an issue when using stripe-mock, but better safe than sorry
- New `ListAutoPagingAsync` tests were added to all service test classes.

Fixes #1946.

### DIFF
--- a/src/Stripe.net/Services/TaxIds/TaxIdService.cs
+++ b/src/Stripe.net/Services/TaxIds/TaxIdService.cs
@@ -67,5 +67,12 @@ namespace Stripe
         {
             return this.ListNestedEntitiesAutoPaging(customerId, options, requestOptions);
         }
+
+#if !NET45
+        public virtual IAsyncEnumerable<TaxId> ListAutoPagingAsync(string customerId, TaxIdListOptions options = null, RequestOptions requestOptions = null, CancellationToken cancellationToken = default)
+        {
+            return this.ListNestedEntitiesAutoPagingAsync(customerId, options, requestOptions, cancellationToken);
+        }
+#endif
     }
 }
--- a/src/Stripe.net/Services/_base/Service.cs
+++ b/src/Stripe.net/Services/_base/Service.cs
@@ -6,9 +6,11 @@ namespace Stripe
     using System.Net;
     using System.Net.Http;
     using System.Reflection;
+#if !NET45
+    using System.Runtime.CompilerServices;
+#endif
     using System.Threading;
     using System.Threading.Tasks;
-    using Stripe.Infrastructure;
 
     /// <summary>Abstract base class for all services.</summary>
     /// <typeparam name="TEntityReturned">
@@ -163,6 +165,20 @@ namespace Stripe
                 requestOptions);
         }
 
+#if !NET45
+        protected IAsyncEnumerable<TEntityReturned> ListEntitiesAutoPagingAsync(
+            ListOptions options,
+            RequestOptions requestOptions,
+            CancellationToken cancellationToken)
+        {
+            return this.ListRequestAutoPagingAsync<TEntityReturned>(
+                this.ClassUrl(),
+                options,
+                requestOptions,
+                cancellationToken);
+        }
+#endif
+
         protected TEntityReturned UpdateEntity(
             string id,
             BaseOptions options,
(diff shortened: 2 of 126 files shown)


clement911:
❤️

src/Stripe.net/Services/_base/Service.cs
```diff
@@ -245,6 +261,91 @@ protected TEntityReturned CreateEntity(BaseOptions options, RequestOptions reque
...
+        {
+            return this.ListRequestAutoPagingAsync<T>(url, options, requestOptions)
+                .ToEnumerable();
+        }
+
+        protected async IAsyncEnumerable<T> ListRequestAutoPagingAsync<T>(
+            string url,
+            ListOptions options,
+            RequestOptions requestOptions,
+            [EnumeratorCancellation] CancellationToken cancellationToken = default)
```
ob-stripe: This is not very well documented, but what the [`EnumeratorCancellation`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.enumeratorcancellationattribute?view=netcore-3.1) attribute does is add support for two different ways of passing the cancellation token:
- directly as a normal method parameter
- using the [`.WithCancellation()`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.threading.threadingtools.withcancellation?view=visualstudiosdk-2019) extension method

src/Stripe.net/Stripe.net.csproj
```diff
@@ -38,6 +39,12 @@
     <CodeAnalysisRuleSet>..\_stylecop\StyleCopRules.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="1.1.0" />
```
ob-stripe: This is the backport package that adds support for `IAsyncEnumerable` in .NET Standard 2.0.

src/Stripe.net/Stripe.net.csproj
```diff
@@ -38,6 +39,12 @@
...
+    <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.5.0" />
+    <PackageReference Include="System.Linq.Async" Version="4.0.0" />
```
ob-stripe: This adds support for async enumerators to LINQ methods. We use this to convert the `IAsyncEnumerable` to a regular `IEnumerable` (so that the synchronous version `ListRequestAutoPaging` is just a thin wrapper over the async `ListRequestAutoPagingAsync`).

src/StripeTests/Services/AutoPagingTest.cs
```diff
@@ -287,6 +287,356 @@ public void ListAutoPaging_EndingBefore()
...
+            {
+                Limit = 2,
+            };
+            var models = new List<PageableModel>();
+
+            var source = new CancellationTokenSource();
+
+            try
+            {
+                await foreach (var model in service.ListAutoPagingAsync(options).WithCancellation(source.Token))
```
ob-stripe: This illustrates the new `await foreach` syntax added in C# 8, as well as support for `.WithCancellation()` to pass cancellation tokens.

src/Stripe.net/Services/TaxIds/TaxIdService.cs
```diff
@@ -67,5 +67,12 @@ public virtual IEnumerable<TaxId> ListAutoPaging(string customerId, TaxIdListOpt
         {
             return this.ListNestedEntitiesAutoPaging(customerId, options, requestOptions);
         }
+
+#if !NET45
+        public virtual IAsyncEnumerable<TaxId> ListAutoPagingAsync(string customerId, TaxIdListOptions options = null, RequestOptions requestOptions = null, CancellationToken cancellationToken = default)
```
cjavilla-stripe: Wondering if these `customerId` parameter names will be easy to derive from the spec, or if it would be easier to use `id` or `parentId` in the nested cases...

remi-stripe: They shouldn't need to. In another PR for your codegen branch we renamed all of those to `parentId` and `id` instead. It's likely just that ob used a codemod or a regex to generate those

ob-stripe: Yep, this PR is like 90% codemod. Sorry for the additional toil CJ :/

ob-stripe:
Thanks both!

clement911:
Thank you!

PR #1948, opened 2020-03-06
Use default literal

ob-stripe:
cc @stripe/api-libraries 

Use the [default literal](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/default#default-literal) when the compiler can infer the type.


ob-stripe:
Thanks CJ!

PR #1949, opened 2020-03-11
Style changes to match codegen

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

If we don't like this style, we can change easily with autogen later, but this will help with diffs. Should be able to cherry pick this into master, I think.


cjavilla-stripe:
cool. that's a 5 char change in codegen and there were like 1/3 this way and 2/3 the other way

PR #1950, opened 2020-03-11
Alphabetize properties

cjavilla-stripe:
r? @remi-stripe
cc @stripe/api-libraries 

Just alphabetizing properties.

### DIFF
--- a/src/StripeTests/Services/Coupons/CouponCreateOptionsTest.cs
+++ b/src/StripeTests/Services/Coupons/CouponCreateOptionsTest.cs
@@ -15,7 +15,7 @@ namespace StripeTests
                 Duration = "forever",
             };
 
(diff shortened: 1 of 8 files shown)


cjavilla-stripe:
broke, then fixed a test. ptal @remi-stripe

src/StripeTests/Services/Coupons/CouponCreateOptionsTest.cs
```diff
@@ -15,7 +15,7 @@ public void Serialize()
                 Duration = "forever",
             };
 
-            Assert.Equal("percent_off=25&duration=forever", FormEncoder.CreateQueryString(options));
+            Assert.Equal("duration=forever&percent_off=25", FormEncoder.CreateQueryString(options));
```
remi-stripe: ah you those can be sneaky!

PR #1951, opened 2020-03-11
Add a handful of missing features

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Note: I didn't add support for listing topups by amount, because we don't have a LongRangeOptions yet. Other than that, these are all (I think!) of the fields that were discovered for options classes when running autogen.

### DIFF
--- a/src/Stripe.net/Services/ApplePayDomains/ApplePayDomainListOptions.cs
+++ b/src/Stripe.net/Services/ApplePayDomains/ApplePayDomainListOptions.cs
@@ -4,5 +4,7 @@ namespace Stripe
 
     public class ApplePayDomainListOptions : ListOptions
     {
+        [JsonProperty("domain_name")]
+        public string DomainName { get; set; }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Stripe.net/Services/Issuing/Cardholders/CardholderListOptions.cs
```diff
@@ -7,6 +7,9 @@ public class CardholderListOptions : ListOptionsWithCreated
         [JsonProperty("email")]
         public string Email { get; set; }
 
+        [JsonProperty("is_default")]
+        public bool? IsDefault { get; set; }
+
```
remi-stripe: Remove that one, it's going away (shouldn't be in the spec, will be removed soon) so let's not add this

cjavilla-stripe: Will do. Thanks for knowing the future.

Issue #1954, opened 2020-03-13
Setting a field to null?


clement911:
If I understand correctly, setting an option's property to `null` will cause the field to not be sent over the wire.

If that is the case, how I can actually set a field to a null value (i.e clearing the value).

For example, if I want to clear the customer's phone number.

I can set `CustomerUpdateOptions.Phone` to `null`. But it won't be serialized at all right?

Unless I'm missing something obvious here...

remi-stripe:
@clement911 In the API, if you pass an empty string it will unset the string field.

```
var options2 = new CustomerUpdateOptions {
  Email = "",
  Phone = "",
};
Customer customer2 = service.Update(customer.Id, options2);
```

This would return `email: null` and `phone: null` in the API.

clement911:
What if I want to set an empty string?
What about other types such as numerics and dates and bools?

remi-stripe:
You can't set an empty string in that case, it will just set the field to null already. This is how the API itself works, so it's not really specific to stripe-dotnet.

You can not unset number of booleans with stripe-dotnet though our API rarely supports this in the first place.

clement911:
hmm ok.
Seems like quite a limitation...

clement911:
Although I'm not sure if it's a big deal in practice...

remi-stripe:
In my experience, it's not. Most booleans are set to true or false, we rarely have a null "state". We're also moving more to enums instead to represent the "true/false/unset" state where it matters.

clement911:
Ok, thanks for the quick response.

PR #1956, opened 2020-03-18
[wip] Add support for unsettable fields

ob-stripe:


src/Stripe.net/Services/Customers/CustomerUpdateOptions.cs
```diff
@@ -24,16 +24,16 @@ public class CustomerUpdateOptions : BaseOptions, IHasMetadata
         public long? Balance { get; set; }
 
         [JsonProperty("coupon")]
-        public string Coupon { get; set; }
+        public AnyOf<string, EmptyParam> Coupon { get; set; }
 
         [JsonProperty("default_source")]
-        public string DefaultSource { get; set; }
+        public AnyOf<string, EmptyParam> DefaultSource { get; set; }
```
remi-stripe: I am fairly sure you can never unset a default source. If the openapi spec says otherwise, the spec is wrong.

src/Stripe.net/Services/Customers/CustomerUpdateOptions.cs
```diff
@@ -24,16 +24,16 @@ public class CustomerUpdateOptions : BaseOptions, IHasMetadata
...
 
         [JsonProperty("description")]
-        public string Description { get; set; }
+        public AnyOf<string, EmptyParam> Description { get; set; }
 
         [JsonProperty("email")]
-        public string Email { get; set; }
+        public AnyOf<string, EmptyParam> Email { get; set; }
```
remi-stripe: Does it make sense to support this on string since you can simply pass an empty string today and it just works?

src/Stripe.net/Services/Customers/CustomerUpdateOptions.cs
ob-stripe: I copied this from stripe-java, so the spec is likely wrong.

src/Stripe.net/Services/Customers/CustomerUpdateOptions.cs
ob-stripe: That's certainly debatable. Is this:

```c#
var options = new CustomerUpdateOptions { Email = EmptyParam.Empty };
```

better/clearer than this:

```c#
var options = new CustomerUpdateOptions { Email = "" };
// or even
var options = new CustomerUpdateOptions { Email = string.Empty };
```
?

In stripe-java we do define an overload with `EmptyParam` even for strings, but we disallow empty strings, which we don't do in stripe-dotnet.

Let's discuss this with @cjavilla-stripe and @richardm-stripe next week.

clement911:
Will this fix https://github.com/stripe/stripe-dotnet/issues/1954 and allow setting fields to null?

Wouldn't `EmptyParam.Empty` be better named `NullParam.Null` since it's setting to null and not empty string?
This will work with other types such as `long`, `DateTime`, `bool` too right? If so, the rename would even make more sense because in .net we talk about a nullable long, not an empty long (for example).

What do you think?

ob-stripe:
> Will this fix #1954 and allow setting fields to null?

As long as the field can be unset on the API's side, yes. That's not the case for all fields.

> Wouldn't EmptyParam.Empty be better named NullParam.Null since it's setting to null and not empty string?

Technically there isn't really a difference. Because Stripe's API uses form-encoding for requests, there isn't an explicit notion of "null" -- if you unset a field named `foo`, the encoded request body will look like `foo=`.

That said that's more of an implementation detail and I agree that `EmptyParam.Empty` might not be the best name. We'll take this into consideration. Thanks for the feedback!

clement911:
> Technically there isn't really a difference. Because Stripe's API uses form-encoding for requests, there isn't an explicit notion of "null" -- if you unset a field named foo, the encoded request body will look like foo=.
> 
> That said that's more of an implementation detail and I agree that EmptyParam.Empty might not be the best name.

Indeed that sounds like an implementation detail.

Consider the fact that if you set foo= and load foo back from the API, the API would return null and not "". If I understand correctly...

ob-stripe:
Closing due to staleness.

PR #1957, opened 2020-03-19
Add net461 as a target framework

ob-stripe:
cc @stripe/api-libraries 

Add `net461` (i.e. .NET Framework 4.6.1) as a target framework.

.NET Framework 4.6.1 is compatible with .NET Standard 2.0 (see https://docs.microsoft.com/en-us/dotnet/standard/net-standard), so in theory this shouldn't be needed, but there are a couple of reasons for this PR:
- currently, users running on .NET Framework 4.6.1 or more recent are using the `net45` assembly rather than the `netstandard2.0` one. This is not an issue at the moment, but it will be once #1947 is merged as those users would not be able to benefit from async autopagination despite the fact that their runtime can support it
- help us better measure the proportion of users still running on .NET Framework >= 4.5 but < 4.6.1. I suspect there are few such users and we'd eventually want to drop support for those versions

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
+++ b/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
@@ -24,6 +24,8 @@ namespace Stripe
         private const string StripeNetTargetFramework =
 #if NETSTANDARD2_0
             "netstandard2.0"
+#elif NET461
+            "net461"
 #elif NET45
             "net45"
 #else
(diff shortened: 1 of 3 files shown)


ob-stripe:
> Seems peaceful, but I'm not 100% sure I understand why we can remove the References to System.* are those all implied with the target frameworks?

Sorry, should have clarified that bit -- those references were unneeded, we simply don't use them in the library. The only ones we need are:
- `System.Net.Http`: obviously, used to send HTTP requests to Stripe's API
- `Sytem.Configuration`: used as an alternative way of setting the global API key: https://github.com/stripe/stripe-dotnet/blob/0d7fc51063c11d7d2e9fdd2d8c76231c6bc5bf15/src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs#L44-L48

ob-stripe:
Thanks!

PR #1958, opened 2020-03-19
[wip] stripe-dotnet v36.0

### LINKED ISSUE #1946: ListAutoPaging async version

ob-stripe:
- [x] Add net461 as a target framework (#1957)
- [x] Support for asynchronous autopagination (#1947)

Fixes #1946 (via #1947).


ob-stripe:
Closing this, the next major version is delayed for the time being. Will reopen individual PRs for the two non-breaking features in this branch.

PR #1959, opened 2020-03-19
Add new fields to Issuing `Card` and `Authorization`

remi-stripe:
Multiple API changes for Issuing `Card` and `Authorization`
* Add `Amount` and `AmountCurrency` to `Authorization`
* Add `MerchantAmount` and `MerchantCurrency` to `Authorization`
* Add `PendingRequest` to `Authorization`
* Add `ReplacedBy` to `Card`
* Add `Amount` and `Metadata` to `AuthorizationApproveOptions`
* Add `Metadata` to `AuthorizationDeclineOptions`

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/Issuing/Authorizations/Authorization.cs
+++ b/src/Stripe.net/Entities/Issuing/Authorizations/Authorization.cs
@@ -20,6 +20,12 @@ namespace Stripe.Issuing
         public string Object { get; set; }
 
         /// <summary>
+        /// The total amount in the card's currency that was authorized or rejected.
+        /// </summary>
+        [JsonProperty("amount")]
+        public long Amount { get; set; }
+
+        /// <summary>
         /// Whether the authorization has been approved.
         /// </summary>
         [JsonProperty("approved")]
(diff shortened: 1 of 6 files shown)


src/Stripe.net/Entities/Issuing/Authorizations/Authorization.cs
```diff
@@ -32,16 +38,11 @@ public class Authorization : StripeEntity<Authorization>, IHasId, IHasMetadata,
         [JsonProperty("authorization_method")]
         public string AuthorizationMethod { get; set; }
 
-        /// <summary>
-        /// The amount that has been authorized. This will be 0 when the object is created, and
-        /// increase after it has been approved.
-        /// </summary>
+        [Obsolete("This field is considered deprecated.")]
```
ob-stripe: If we're not pointing to a replacement field/method, I think we can simply tag the field as `[Obsolete]` without a message. The current message does not provide any useful information.

remi-stripe: Ack fair. It's going away tomorrow so I'll keep as is but will do that next time

remi-stripe:
@ob-stripe re-assigning to you as I had to modify `RequestHistory` too

PR #1961, opened 2020-03-21
Rename a lot of core classes to match the correct naming scheme of the library

remi-stripe:
Took the opportunity to rename a lot of older classes that were never prefixed, paving the way for the release of the codegen-ed library in the next few months.

List of changes:
  * Renamed  the class `Fee` to `BalanceTransactionFee` on `BalanceTransaction`
  * Renamed the class `Outcome` to `ChargeOutcome` on `Charge`
  * Renamed the classes `Evidence` to `DisputeEvidence` and `EvidenceDetails` to `DisputeEvidenceDetails` on `Dispute`
  * Renamed the class `OutcomeRule` to `Rule` and moved it to the `Radar` namespace
  * Renamed the classes `ShippingMethod` to `OrderShippingMethod`, `StatusTransitions` to `OrderStatusTransitions` and `DeliveryEstimate` to `OrderShippingMethodDeliveryEstimate` on `Order`
  * Renamed the class `Inventory` to `SkuInventory` on `Sku`
  * Fixed the type of `NextPendingInvoiceItemInvoice` on `Subscription` to be a `DateTime?`
  * `FraudDetails` on `Charge` is now a class `ChargeFraudDetails` instead of a dictionary
  * Removed `Connect` on the `WebhookEndpoint` resource as it is not supported

cc @stripe/api-libraries


src/Stripe.net/Entities/Charges/ChargeOutcomeRule.cs
```diff
@@ -2,7 +2,7 @@ namespace Stripe
 {
     using Newtonsoft.Json;
 
-    public class OutcomeRule : StripeEntity<OutcomeRule>, IHasId
+    public class ChargeOutcomeRule : StripeEntity<ChargeOutcomeRule>, IHasId
```
remi-stripe: So. I wanted to flag that one explicitly @ob-stripe.
It's quite an old property but, if I am not mistake, this should be a real API resource with an `object` field since this is expandable on `Charge.outcome.rule`. Really curious what your thoughts are here?

ob-stripe: Strictly speaking we don't _need_ the `object` field if the expandable field is not polymorphic. But AFAIK this is the only case of an expandable field where the expanded object is not an API resource with an `object` field.

I don't think it's worth fixing, but we should discuss this internally to decide if this is something we'd want to allow on other API resources in the future.

remi-stripe: The reason I'm asking is that, if we added `object` later the class would have the object value as a name. When right now it's kinda `ChargeOutcomeRule` and in a way it's annoying

ob-stripe:
@remi-stripe Should we close this for now?

ob-stripe:
I've reset the `integration-v36` branch, so you'll need to rebase to remove the extra commits.

src/Stripe.net/Entities/Charges/ChargeOutcomeRule.cs
remi-stripe: Do you think we should call this one `Rule` in the Radar namespace in that case so that in the future, if we add `object` it will likely be `object: "radar.rule"`?

remi-stripe:
@ob-stripe Re-assigning to you for review again. Flagging I fixed another issue raised by @clement911 in https://github.com/stripe/stripe-dotnet/issues/1969#issuecomment-606604073

src/Stripe.net/Entities/Charges/ChargeOutcomeRule.cs
ob-stripe: We should do the same thing as stripe-java to facilitate future codegen, so yes, `ChargeOutcomeRule` should be renamed to `Rule` in the Radar namespace: https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/radar/Rule.java. (Thanks for checking!)

remi-stripe: Fixed, PTAL

PR #1970, opened 2020-03-28
Add `RawJObject` getter to `StripeEntity`

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add `RawJObject` getter to `StripeEntity`. This method returns a `JObject` that can be used to access any properties contained in the JSON response, even if the library does not have a proper accessor for it.

This is similar to the escape hatch we have in stripe-java: https://github.com/stripe/stripe-java/pull/799

### DIFF
--- a/src/Stripe.net/Entities/_base/StripeEntity.cs
+++ b/src/Stripe.net/Entities/_base/StripeEntity.cs
@@ -4,11 +4,51 @@ namespace Stripe
     using System.Reflection;
     using System.Runtime.CompilerServices;
     using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
     using Stripe.Infrastructure;
 
     [JsonObject(MemberSerialization.OptIn)]
     public abstract class StripeEntity : IStripeEntity
     {
+        [JsonIgnore]
+        private JObject rawJObject;
+
+        /// <summary>
+        /// Gets the raw <see cref="JObject">JObject</see> exposed by the Newtonsoft.Json library.
+        /// This can be used to access properties that are not directly exposed by Stripe's .NET
+        /// library.
+        /// </summary>
+        /// <remarks>
+        /// You should always prefer using the standard property accessors whenever possible. This
+        /// accessor is not considered fully stable and might change or be removed in future
+        /// versions.
+        /// </remarks>
+        /// <returns>The raw <see cref="JObject">JObject</see>.</returns>
+        [JsonIgnore]
+        public JObject RawJObject
+        {
+            get
+            {
+                // Lazily initialize the object the first time the getter is called.
+                if (this.rawJObject == null)
+                {
+                    if (this.StripeResponse == null)
+                    {
+                        return null;
+                    }
+
+                    this.rawJObject = JObject.Parse(this.StripeResponse.Content);
+                }
+
+                return this.rawJObject;
+            }
+
+            protected set
+            {
+                this.rawJObject = value;
+            }
+        }
+
         [JsonIgnore]
         public StripeResponse StripeResponse { get; set; }
 
--- a/src/StripeTests/Entities/_base/StripeEntityTest.cs
+++ b/src/StripeTests/Entities/_base/StripeEntityTest.cs
@@ -5,8 +5,14 @@ namespace StripeTests
     using Stripe.Infrastructure;
     using Xunit;
 
-    public class StripeEntityTest
+    public class StripeEntityTest : BaseStripeTest
     {
+        public StripeEntityTest(
+            StripeMockFixture stripeMockFixture)
+            : base(stripeMockFixture)
+        {
+        }
+
         [Fact]
         public void FromJsonAuto()
         {
@@ -139,6 +145,29 @@ namespace StripeTests
             Assert.NotNull(o.Nested);
             Assert.Equal("id_expanded", o.Nested.Id);
             Assert.Equal(42, o.Nested.Bar);
+        }
+
+        [Fact]
+        public void RawJObject()
+        {
+            var service = new SubscriptionService(this.StripeClient);
+            var subscription = service.Get("sub_123");
+
+            Assert.NotNull(subscription);
+
+            // Access `id`, a string element
+            Assert.Equal(subscription.Id, subscription.RawJObject["id"]);
+
+            // Access `created`, a number element
+            Assert.Equal(subscription.Created, subscription.RawJObject["created"]);
+
+            // Access `plan[id]`, a nested string element
+            Assert.Equal(subscription.Plan.Id, subscription.RawJObject["plan"]["id"]);
+
+            // Access `items[data][0][id]`, a deeply nested string element
+            Assert.Equal(
+                subscription.Items.Data[0].Id,
+                subscription.RawJObject["items"]["data"][0]["id"]);
         }
 
         private class TestEntity : StripeEntity<TestEntity>


remi-stripe:
@ob-stripe What's the difference with the approach we tried last year in https://github.com/stripe/stripe-dotnet/pull/1707

src/Stripe.net/Entities/_base/StripeEntity.cs
```diff
@@ -4,11 +4,46 @@ namespace Stripe
...
+        /// <returns>The raw <see cref="JObject">JObject</see>.</returns>
+        [JsonIgnore]
+        public JObject RawJObject
+        {
+            get
+            {
+                // Lazily initialize the object the first time the getter is called.
+                if (this.rawJObject == null)
+                {
+                    this.rawJObject = JObject.Parse(this.StripeResponse.Content);
```
remi-stripe: Now that I think about it, why don't we just tell users to do that themselves?

src/Stripe.net/Entities/_base/StripeEntity.cs
```diff
@@ -4,11 +4,46 @@ namespace Stripe
...
+        /// accessor is not considered fully stable and might change or be removed in future
+        /// versions.
+        /// </remarks>
+        /// <returns>The raw <see cref="JObject">JObject</see>.</returns>
+        [JsonIgnore]
+        public JObject RawJObject
+        {
+            get
+            {
+                // Lazily initialize the object the first time the getter is called.
```
remi-stripe: Can't we just do that at the time we parse the response to avoid having to parse twice?

ob-stripe:
> @ob-stripe What's the difference with the approach we tried last year in #1707

All fields are accessible, not just those that don't have normal accessors. So if someone uses `RawJObject` to access a field, and we later add a normal accessor for that field, their integration won't suddenly stop working.

src/Stripe.net/Entities/_base/StripeEntity.cs
```diff
@@ -4,11 +4,46 @@ namespace Stripe
...
+
+        /// <summary>
+        /// Gets the raw <see cref="JObject">JObject</see> exposed by the Newtonsoft.Json library.
+        /// This can be used to access properties that are not directly exposed by Stripe's .NET
+        /// library.
+        /// </summary>
+        /// <remarks>
+        /// You should always prefer using the standard property accessors whenever possible. This
+        /// accessor is not considered fully stable and might change or be removed in future
+        /// versions.
```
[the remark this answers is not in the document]
ob-stripe: In practice that will very likely be the case, but this exposes a `JObject` which is a Newtonsoft.Json class, so we don't have full control over it.

src/Stripe.net/Entities/_base/StripeEntity.cs
ob-stripe: We could. It's literally one line of code, and `JObject` is very easy to use. If you want to close this PR and just advise users to use that line of code instead, that's fine by me!

src/Stripe.net/Entities/_base/StripeEntity.cs
ob-stripe: No, there is no way to parse the JSON once but deserialize it into two different objects (the proper `StripeEntity` subclass, and the raw `JObject`).

src/Stripe.net/Entities/_base/StripeEntity.cs
remi-stripe: Hmmm I guess us deserializing means we keep using the correct settings internally. Let's ship this since it's easier to discover I think.

clement911: shouldn't you check that the `StripeResponse` is not null?

clement911:
It's not new but one thing that's a bit strange is that `IStripeEntity.StripeResponse` is only populated on the top level entity returned by the API.
Same thing for this new `RawJObject` I guess.

I was thinking we could move them to a new `IStripeRootEntity` interface, only implemented by root objects.
However, for listable entities, they would be the root entity if `Get` is used but not if `List` is used (in that case the `StripeList<T>` would be the root entity). So it's probably not the best design.

Example showing the current issue:

`new CustomerService().Get(id).RawJObject["SecretField"]` => OK

`new CustomerService().List().First().RawJObject["SecretField"]` => NullReference exception

Ideally, there would be a way to populate the `RawObject` for ALL entities in the returned object graph. I wonder if the `Newtonsoft` can do that...

clement911:
Also not sure if you looked into System.Text.Json?

clement911:
I'm pretty sure a custom json deserializer would be able to set a JObject property on all entities.

ob-stripe:
Hey @clement911, thanks for all the feedback as usual :)

I think having the `StripeResponse` be available only on the top-level object returned by the API makes sense and is appropriate for all reasonable use cases I can think of, but I'm happy to hear about counter examples.

For `RawJObject` specifically, I do agree that it would be very convenient to have it available at every level (so you can use the "real" accessors up and keep type safety up until the point you can't), but I don't think we can accomplish that without a new custom deserializer as you mentioned, which I'd prefer to avoid. We have quite a few custom deserializers already and I'd rather not add any new ones if we can avoid it. Using `RawJObject` should be a "last resort" so hopefully this doesn't come up too often. That said, we can ship this as is and add support for `RawJObject` at every level as a future improvement.

System.Text.Json looks interesting, and I would definitely favor a standard library solution over an external dependency like Newtonsoft.Json, but it looks like it's only available as of .NET Core 3.0. Unless it's backported to .NET Standard 2.0, we will be unable to use it :(

Good point about the NRE, will add a check.

ob-stripe:
Released as 35.11.0.

sstalder:
This breaks my dotnet core 3 project not using Newtonsoft.Json as their serializer, and I would assume anyone using System.Text.Json.

This is due to the JsonIgnore attribute from Newtonsoft not being used by System.Text.Json as it has it's own attribute.

There is some more information under this section on the migration guide: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#conditionally-ignore-a-property

> An unhandled exception has occurred while executing the request.
System.NotSupportedException: The collection type 'Newtonsoft.Json.Linq.JObject' on 'Stripe.Checkout.Session.RawJObject' is not supported.
   at System.Text.Json.JsonPropertyInfoNotNullable`4.GetDictionaryKeyAndValueFromGenericDictionary(WriteStackFrame& writeStackFrame, String& key, Object& value)
   at System.Text.Json.JsonPropertyInfo.GetDictionaryKeyAndValue(WriteStackFrame& writeStackFrame, String& key, Object& value)
   at System.Text.Json.JsonSerializer.HandleDictionary(JsonClassInfo elementClassInfo, JsonSerializerOptions options, Utf8JsonWriter writer, WriteStack& state)
   at System.Text.Json.JsonSerializer.Write(Utf8JsonWriter writer, Int32 originalWriterDepth, Int32 flushThreshold, JsonSerializerOptions options, WriteStack& state)
   at System.Text.Json.JsonSerializer.WriteAsyncCore(Stream utf8Json, Object value, Type inputType, JsonSerializerOptions options, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.Formatters.SystemTextJsonOutputFormatter.WriteResponseBodyAsync(OutputFormatterWriteContext context, Encoding selectedEncoding)
   at Microsoft.AspNetCore.Mvc.Formatters.SystemTextJsonOutputFormatter.WriteResponseBodyAsync(OutputFormatterWriteContext context, Encoding selectedEncoding)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResultFilterAsync>g__Awaited|29_0[TFilter,TFilterAsync](ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResultExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.ResultNext[TFilter,TFilterAsync](State& next, Scope& scope, Object& state, Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeResultFilters()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResourceFilter>g__Awaited|24_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResourceExecutedContextSealed context)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeFilterPipelineAsync>g__Awaited|19_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeAsync>g__Logged|17_1(ResourceInvoker invoker)
   at Microsoft.AspNetCore.Routing.EndpointMiddleware.<Invoke>g__AwaitRequestTask|6_0(Endpoint endpoint, Task requestTask, ILogger logger)
   at Microsoft.AspNetCore.Authorization.AuthorizationMiddleware.Invoke(HttpContext context)
   at IdentityServer4.Hosting.IdentityServerMiddleware.Invoke(HttpContext context, IEndpointRouter router, IUserSession session, IEventService events)
   at IdentityServer4.Hosting.MutualTlsTokenEndpointMiddleware.Invoke(HttpContext context, IAuthenticationSchemeProvider schemes)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at IdentityServer4.Hosting.BaseUrlMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.MigrationsEndPointMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.DatabaseErrorPageMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.DatabaseErrorPageMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)

ob-stripe:
@sstalder Could you open a new issue about this?

PR #1972, opened 2020-03-31
Add net461 as a target framework

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Same as #1957, but targets `master`.


ob-stripe:
Released as 35.11.0.

PR #1976, opened 2020-03-31
Remove System.Linq.Async dependency

### LINKED ISSUE #1974: The call is ambiguous between the following methods or properties: 'System.Linq.Queryable.Where(System.Linq.IQueryable, System.Linq.Expressions.Expression<System.Func<TSource, bool>>)' and 'System.Linq.AsyncEnumerable.Where(System.Collections.Generic.IAsyncEnumerable, System.Func<TSource, bool>)'
Related to the new release of 35.11.0.

I use Stripe.Net and EntityFrameworkCore in the same project.

Now that Stripe.Net references System.Linq.Async I am getting the following errors when trying to query an EF Core DbSet (which implements IAsyncEnumerable):

The call is ambiguous between the following methods or properties: 'System.Linq.Queryable.Where(System.Linq.IQueryable, System.Linq.Expressions.Expression<System.Func<TSource, bool>>)' and 'System.Linq.AsyncEnumerable.Where(System.Collections.Generic.IAsyncEnumerable, System.Func<TSource, bool>)'

This is because AsyncEnumerable also includes a Where() function.

I have created a Gist here: https://gist.github.com/pauldbentley/7d5e60cd47509a4011d240b55d72b2e8

Is there a way I can tell the compiler not to "see" the reference to System.Linq.Async which is coming from Stripe.Net?

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Remove the dependency on System.Linq.Async, by reimplementing the `.ToEnumerable()` method directly.

Fixes #1974.

### DIFF
--- a/src/Stripe.net/Infrastructure/AsyncUtils.cs
+++ b/src/Stripe.net/Infrastructure/AsyncUtils.cs
@@ -1,0 +1,76 @@
+#if !NET45
+namespace Stripe.Infrastructure
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    internal static class AsyncUtils
+    {
+        /// <summary>
+        /// Converts an async-enumerable sequence to an enumerable sequence.
+        /// </summary>
+        /// <typeparam name="TSource">The type of the elements in the source sequence.</typeparam>
+        /// <param name="source">An async-enumerable sequence to convert to an enumerable sequence.</param>
+        /// <returns>The enumerable sequence containing the elements in the async-enumerable sequence.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="source"/> is null.</exception>
+        /// <remarks>
+        /// This code is largely borrowed from the System.Linq.Async project (see
+        /// https://github.com/dotnet/reactive/blob/7ad606b3dcd4bb2c6ae9622f8a59db7f8f52aa85/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToEnumerable.cs).
+        /// The reason we're not using System.Linq.Async directly is that it can cause issues with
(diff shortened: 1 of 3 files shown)


ob-stripe:
The source code is basically a copy/paste from System.Linq.Async's implementation: https://github.com/dotnet/reactive/blob/7ad606b3dcd4bb2c6ae9622f8a59db7f8f52aa85/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToEnumerable.cs

ob-stripe:
Sorry for force-pushing mid review :( I moved most of the code into `AsyncUtils`.

ob-stripe:
Thanks Brandur!

ob-stripe:
Released as 35.11.1.

clement911:
This code is a bit weird, they even [mentioned](https://github.com/dotnet/reactive/blob/7ad606b3dcd4bb2c6ae9622f8a59db7f8f52aa85/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToEnumerable.cs#L12) it's an anti pattern.

Async over sync is often worse than just sync so another option would be to have the !NET45 `ListRequestAutoPaging` used the same implementation as the `ListRequestAutoPaging` used fo NET45 instead of doing sync over `ListRequestAutoPagingAsync`.. Then you could remove `AsyncUtils` altogether.


Just a suggestion...

ob-stripe:
Yeah, I was a little bit on the fence about it. The autopagination implementation is complex enough that I'd rather avoid duplicating the method, so I went with this.

(Of course right now the implementation is duplicated enough because of `NET45`, but we will likely drop support for .NET Framework < 4.6.1 at some point.)

clement911:
Looks like it might cause deadlocks https://github.com/stripe/stripe-dotnet/issues/1946#issuecomment-610873922

PR #1982, opened 2020-04-08
Fix deadlock in `AsyncUtils.ToEnumerable()`

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

All async calls in libraries should use `ConfigureAwait(false)` to avoid deadlocks. This is the same issue that I fixed a while back in #1658.

This [blog post](https://medium.com/bynder-tech/c-why-you-should-use-configureawait-false-in-your-library-code-d7837dce3d7f) explains why this is necessary.

Fixes the issue reported in https://github.com/stripe/stripe-dotnet/issues/1946#issuecomment-610873922.


ob-stripe:
Released as 35.12.1.

PR #1998, opened 2020-04-14
Add support for the `BillingPortal` namespace and the `Session` API and resource

remi-stripe:
cc @stripe/api-libraries 
cc @alexander-stripe


remi-stripe:
r? @cjavilla-stripe 

This mirrors https://github.com/stripe/stripe-java/pull/1014 in stripe-java

src/Stripe.net/Infrastructure/Public/StripeTypeRegistry.cs
```diff
@@ -21,6 +21,7 @@ public static class StripeTypeRegistry
             { "balance", typeof(Balance) },
             { "balance_transaction", typeof(BalanceTransaction) },
             { "bank_account", typeof(BankAccount) },
+            { "billing_portal.session", typeof(BillingPortal.Session) },
```
cjavilla-stripe, line 24: not quite sure how this bit works, but it seems to match the issuing ones, so lgtm

PR #2000, opened 2020-04-15
Ability to inspect the configuration of SystemNetHttpClient

captin411:
These read only properties now exist on the SystemNetHttpClient
- MaxNetworkRetries
- EnableTelemetry

This will allow application developers to write tests that ensure that their configuration code has properly setup the StripeClient with the desired network retries, or telemetry settings without resorting to reflection in order to read private attributes.

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
+++ b/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
@@ -38,11 +38,7 @@ namespace Stripe
 
         private readonly System.Net.Http.HttpClient httpClient;
 
-        private readonly int maxNetworkRetries;
-
         private readonly AppInfo appInfo;
-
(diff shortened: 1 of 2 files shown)


ob-stripe:
@captin411 Thanks for the PR!

I'm fine with the proposed change, but could you replace the read-only property and the accessor with an auto-property instead?

I.e. replace:

```c#
private readonly bool enableTelemetry;

public bool EnableTelemetry { get => this.enableTelemetry; }
```

with:

```c#
public bool EnableTelemetry { get; }
```

and same thing for `MaxNetworkRetries`. It's how read-only properties are handled in the rest of the codebase.

captin411:
No problem @ob-stripe !  That does look a lot better.

ob-stripe:
Thanks @captin411!

Also, PR #2000 🥳

PR #2005, opened 2020-04-17
Swallow exceptions when gathering runtime information

### LINKED ISSUE #1986: AmbiguousMatchException from RuntimeInformation.FrameworkDescription
Since upgrading to v35, I'm getting an error from the StripeClient constructor when running on Windows 2012 R2 Server (under IIS).  This is related to the changes in v35 to enhance the user agent string being reported back to the Stripe servers.

> System.Web.HttpUnhandledException (0x80004005): Exception of type 'System.Web.HttpUnhandledException' was thrown. ---> System.Reflection.AmbiguousMatchException: Multiple custom attributes of the same type found.
>    at System.Attribute.GetCustomAttribute(Assembly element, Type attributeType, Boolean inherit)
>    at System.Runtime.InteropServices.RuntimeInformation.get_FrameworkDescription()
>    at Stripe.Infrastructure.RuntimeInformation.get_IsFullFramework()
>    at Stripe.Infrastructure.RuntimeInformation.GetRuntimeVersion()
>    at Stripe.SystemNetHttpClient.BuildStripeClientUserAgentString()
>    at Stripe.SystemNetHttpClient..ctor(HttpClient httpClient, Int32 maxNetworkRetries, AppInfo appInfo, Boolean enableTelemetry)
>    at Stripe.StripeClient..ctor(String apiKey, String clientId, IHttpClient httpClient, String apiBase, String connectBase, String filesBase)
> .... (my code)

This did not happen from my development machine (Windows 10) nor from a Console app on the same Windows Server 2012.  I eventually tracked this down to https://github.com/microsoft/ApplicationInsights-dotnet-server/issues/241 - I had Application Insights Monitor installed on the server (though was no longer using it).  Removing this prevented the problem.

I appreciate that this problem isn't technically caused by Stripe, but it might be better to trap exceptions when you're building your user agent string for cases where this bug is causing problems.  If you choose not to change anything, at least the explanation & workaround is here for other people.

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Swallow exceptions when gathering runtime information.

Fixes #1986.

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
+++ b/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
@@ -222,10 +222,23 @@ namespace Stripe
                 { "bindings_version", StripeConfiguration.StripeNetVersion },
                 { "lang", ".net" },
                 { "publisher", "stripe" },
-                { "lang_version", RuntimeInformation.GetRuntimeVersion() },
-                { "os_version", RuntimeInformation.GetOSVersion() },
                 { "stripe_net_target_framework", StripeNetTargetFramework },
             };
+
+            // The following values are in a try/catch block on the off chance that the
+            // RuntimeInformation methods fail in an unexpected way. This should ~never happen, but
+            // if it does it should not prevent users from sending requests.
+            // See https://github.com/stripe/stripe-dotnet/issues/1986 for context.
+            try
+            {
+                values.Add("lang_version", RuntimeInformation.GetRuntimeVersion());
+                values.Add("os_version", RuntimeInformation.GetOSVersion());
+            }
+            catch (Exception)
+            {
+                values.Add("lang_version", "(unknown)");
+                values.Add("os_version", "(unknown)");
+            }
 
             if (this.appInfo != null)
             {


src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
```diff
@@ -222,11 +222,23 @@ private string BuildStripeClientUserAgentString()
...
+            // if it does it should not prevent users from sending requests.
+            // See https://github.com/stripe/stripe-dotnet/issues/1986 for context.
+            try
+            {
+                values.Add("lang_version", RuntimeInformation.GetRuntimeVersion());
+                values.Add("os_version", RuntimeInformation.GetOSVersion());
+            }
+            catch (Exception)
+            {
+                // Do nothing.
```
remi-stripe, line 239: Can we add something explicit that we had an error? It would allow us to reliably know how often this happens and how skewed our stats are

ob-stripe: I think the absence of the `lang_version` and `os_version` keys works just as well as explicit magic values for that.

ob-stripe: (fixed)

PR #2010, opened 2020-04-21
Add support for the `Price` resource and APIs

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries @areagle-stripe

### DIFF
--- a/src/Stripe.net/Constants/Events.cs
+++ b/src/Stripe.net/Constants/Events.cs
@@ -569,6 +569,21 @@ namespace Stripe
         public const string PlanUpdated = "plan.updated";
 
         /// <summary>
+        /// Occurs whenever a price is created.
+        /// </summary>
+        public const string PriceCreated = "price.created";
+
+        /// <summary>
+        /// Occurs whenever a price is deleted.
+        /// </summary>
+        public const string PriceDeleted = "price.deleted";
+
+        /// <summary>
+        /// Occurs whenever a price is updated.
+        /// </summary>
+        public const string PriceUpdated = "price.updated";
+
+        /// <summary>
         /// Occurs whenever a product is created.
         /// </summary>
         public const string ProductCreated = "product.created";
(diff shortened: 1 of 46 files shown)


src/Stripe.net/Entities/SubscriptionSchedules/SubscriptionSchedulePhaseAddInvoiceItem.cs
```diff
@@ -0,0 +1,43 @@
+namespace Stripe
+{
+    using System;
+    using System.Collections.Generic;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class SubscriptionSchedulePhaseAddInvoiceItem : StripeEntity<SubscriptionSchedulePhaseAddInvoiceItem>
+    {
+        #region Expandable Price
```
remi-stripe, line 10: While undocumented, it should be there, and it's coming soon, so I added it at the same time.

src/Stripe.net/Constants/Events.cs
```diff
@@ -568,6 +568,21 @@ public static class Events
         /// </summary>
         public const string PlanUpdated = "plan.updated";
 
+        /// <summary>
```
remi-stripe, line 571: Those are not public yet but should be. I'm adding those because it avoids an extra PR once we fix the issue internally.

src/Stripe.net/Entities/SubscriptionSchedules/SubscriptionSchedulePhaseItem.cs
```diff
@@ -41,6 +41,34 @@ public Plan Plan
...
+        public Price Price
+        {
+            get => this.InternalPrice?.ExpandedObject;
+            set => this.InternalPrice = SetExpandableFieldObject(value, this.InternalPrice);
+        }
+
+        [JsonProperty("price")]
+        [JsonConverter(typeof(ExpandableFieldConverter<Price>))]
+        internal ExpandableField<Price> InternalPrice { get; set; }
+        #endregion
```
ob-stripe, line 70: This doesn't seem to be present in stripe-java.

PR #2013, opened 2020-04-22
Update Newtonsoft on .NETCOREAPP runtimes

iamcarbon:
The Newtonsoft.Json 9.0.1 dependency is missing a .NETSTANDARD2.0 target and produces a unresolvable conflict on .NETCOREAPP3.1 when targeting Linux.

This PR updates Newtonsoft.Json for .NETSTANDARD2.0 (covering the .NETCORE runtimes) to the latest version and fixes the conflicts. This also eliminates bringing in the .NETSTANDARD1.0 facades.

This PR keeps the Newtonsoft.Json library fixed at 9.0.1 for netfx consumers. 

REF: 

Screen Shot 2020-04-21 at 9 11 14 PM

### DIFF
--- a/src/Stripe.net/Stripe.net.csproj
+++ b/src/Stripe.net/Stripe.net.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\netfx.props" />
 
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


remi-stripe:
Thanks for the PR @iamcarbon! I've assigned it to @ob-stripe who can take a look!

ob-stripe:
Not opposed to this, but a couple of caveats:

- we should use the minimum version we can, to prevent potential conflicts in apps which depend on Newtonsoft.Json for something else than the Stripe.net library. [11.0.1](https://www.nuget.org/packages/Newtonsoft.Json/11.0.1) is the first version with a .NET Standard 2.0 target, so we should use that instead of 12.0.3

- this should be considered a breaking change since it has the potential to cause dependency conflicts, so we'll need to release it in the next major version

remi-stripe:
@ob-stripe Do you know what the state of this PR is? Should we be making a change now that we're planning the next major?

iamcarbon:
Bump. Yes, please consider taking this into new release. This pre netstandard dependency creates friction for anyone using the library on .NETCORE runtimes.

remi-stripe:
Closing this one in favour of https://github.com/stripe/stripe-dotnet/pull/2102 which we plan to bundle in the next major version

PR #2016, opened 2020-04-24
[wip] Support for expandable arrays of resources

ob-stripe:
It just works for the most part. Note that when accessing the array of expanded objects when the field was not expanded, it will return a non-null array containing as many `null` values as there are unexpanded IDs.


remi-stripe:
Closing in favour of https://github.com/stripe/stripe-dotnet/pull/2121 which has the feature shipping for real

Issue #2023, opened 2020-05-05
CreditNoteLineOptions Metadata


clement911:
I don't understand why `CreditNoteLineOptions.Metadata` exists given that `CreditNoteLineItem.Metadata` doesn't exist (and is not documented either).

clement911:
I just tested it out and this will actually result in a error if populated:

`Stripe.StripeException: 'Received unknown parameter: lines[0][metadata]'`

remi-stripe:
Thanks for the report. I'll flag the field as deprecated.

clement911:
Sure thing.

PR #2025, opened 2020-05-05
Mark `Metadata` on `CreditNoteLineOptions` as obsolete

remi-stripe:
Fixes https://github.com/stripe/stripe-dotnet/issues/2023
r? @ob-stripe 
cc @stripe/api-libraries


remi-stripe:
Fair enough but seems easier to mark as obsolete and just stick to our usual process. With my luck I'll do a breaking change and then it will be added again a day later :p

clement911:
Using it results in an error so it would be surprising if it was used!

PR #2039, opened 2020-05-13
stripe-dotnet v37.0

remi-stripe:
Integration branch for the next major version of stripe-dotnet

⚠️ = breaking change

- [x] ⚠️ Fix `Quantity` to be nullable on `CreditNoteLineItem`(#2036)
- [x] ⚠️ Removed deprecated parameters (#2038)
  * Removed `PrimaryColor` in `AccountBusinessProfileOptions`. Use `PrimaryColor` in `AccountSettingsBrandingOptions` instead.
  * Removed `metadata` on `CreditNoteLineOptions` as it was never supported
- [x] ⚠️ Removed deprecated `ExchangeRate` resource and APIs (#2038)
  * The `ExchangeRate` APIs have been in a small private beta for a few years and have never been used in stripe-dotnet this year. We are considering this API mostly deprecated as we're working on re-building the feature. This will require large changes to the resource and APIs once we are ready to ship it.
- [x] ⚠️ Multiple properties renamed (#2043)
  * Properties lost their `Id` suffix such as `PaymentMethodId` renamed to `PaymentMethod` on `Charge`
  * `IIN` is now `Iin` on `Card`
  * `AmountReceivd` is now `AmountReceived` on `ChargePaymentMethodDetailsBitcoin`
  * `BankCode` is now `Bank` on `ChargePaymentMethodDetailsIdeal`.
  * `CustomerPurchaseIP` is now `CustomerPurchaseIp` on `DisputeEvidence` and `DisputeEvidenceOptions`
  * `OrderItems` is now `Items` on `Order` and `OrderReturn`
  * `CardList` is now `Cards` on `Recipient`
  * `BankAddressLinePostalCode` is now `BankAddressPostalCode` on `SourceAcssDebit`
  * Properties on `StripeError` have been renamed, `ChargeId` is `Charge`, `ErrorType` is `Type` and `Parameter` is `Param`


lomithrani:
Hello, do you have any idea when the release is to be expected? Someone accidentally upgraded our api version which broke all our subscription process.

remi-stripe:
@lomithrani This should be merged on Monday but those changes are unrelated to any API version and you should be able to use 36.* without issue

lomithrani:
> @lomithrani This should be merged on Monday but those changes are unrelated to any API version and you should be able to use 36.* without issue

Webhooks for checkout session comes back with Customer Rather than customerId, I thought this was related

remi-stripe:
@lomithrani No that should be unrelated. `customer` has always been the name of the field in the API/JSON and `CustomerId` is the property that you use to access the id value in the library so this is expected.

stephen-imperfect:
Was all the renaming of properties really necessary? I value API stability.

remi-stripe:
@stephen-imperfect Thanks for reaching out about this! Originally the stripe-dotnet was entirely written by hand. Whenever a new parameter, property, API method or entire resource was designed we would add it by hand to stripe-dotnet. As we did, we sometimes were inconsistent with how we named the property or the inner class(es) unfortunately. Doing this led to code examples in our docs not matching the library, or similar properties having different names across resources and not matching the name in the API itself.

In 2020, we set to fix the library to be automatically code-generated from our official [openapi spec](https://github.com/stripe/openapi). While doing this, we designed the code to ensure classes and properties matched the exact name of the API and mirrored JSON. This led to discovering a lot of those inconsistencies around names in the library. While we could have special-cased most of the differences, it would have added tech debt to the generation and it would have led to even more inconsistencies over time. For that reason, we decided that it was best to temporarily rename a lot of those properties or classes as breaking changes so that the library was as cleaned up as possible in 37.0.0 and 39.0.0.

The library has now been automatically generated for months and is more consistent overall. It is still not perfect, as there are some subsets of changes we could not properly fix yet. For example we didn't re-design the Upcoming Invoice API's classes in big parts because we know the `LineItem` resource will be used on `Invoice` in the future to replace `InvoiceLine`.

I hope this helps!

stephen-imperfect:
@remi-stripe thanks so much for that, I truly appreciate the context!

PR #2049, opened 2020-05-19
No op changes preparing for autogen

cjavilla-stripe:
Timing out for the night, this'll put a dent in some insignificant changes, but there are likely more coming. 

r? @remi-stripe 
cc @stripe/api-libraries 

* Confirmed the removal of the using statements with Visual Studio
* Updated the expandable fields so that they are of the format, where we use the pascal case name of the field, then a newline (required to pass lint when we have a doc string), then the docstring or the JsonIgnore attribute:

```cs
#region Expandable PascalCaseNoSpacesName

/// docstring
[JsonIgnore]
```

* on save, the editor changed some of these from (no space), I'm fine going one way or another for these:
`[JsonProperty("deleted", NullValueHandling=NullValueHandling.Ignore)]` to (space around equals operator)`[JsonProperty("deleted", NullValueHandling = NullValueHandling.Ignore)]`

* fixed some alphabetization

### DIFF
--- a/src/Stripe.net/Entities/AccountLinks/AccountLink.cs
+++ b/src/Stripe.net/Entities/AccountLinks/AccountLink.cs
@@ -1,7 +1,6 @@
 namespace Stripe
 {
     using System;
-    using System.Collections.Generic;
     using Newtonsoft.Json;
     using Stripe.Infrastructure;
 
--- a/src/Stripe.net/Entities/BankAccounts/BankAccount.cs
+++ b/src/Stripe.net/Entities/BankAccounts/BankAccount.cs
@@ -13,6 +13,7 @@ namespace Stripe
         public string Object { get; set; }
 
         #region Expandable Account
+
         [JsonIgnore]
         public string AccountId
         {
@@ -48,6 +49,7 @@ namespace Stripe
         public string Currency { get; set; }
 
         #region Expandable Customer
+
         [JsonIgnore]
         public string CustomerId
         {
@@ -73,7 +75,7 @@ namespace Stripe
         /// <summary>
         /// Whether this object is deleted or not.
         /// </summary>
-        [JsonProperty("deleted", NullValueHandling=NullValueHandling.Ignore)]
+        [JsonProperty("deleted", NullValueHandling = NullValueHandling.Ignore)]
         public bool? Deleted { get; set; }
 
         [JsonProperty("fingerprint")]
--- a/src/Stripe.net/Services/Checkout/Sessions/SessionCreateOptions.cs
+++ b/src/Stripe.net/Services/Checkout/Sessions/SessionCreateOptions.cs
@@ -1,9 +1,7 @@
 namespace Stripe.Checkout
 {
-    using System;
     using System.Collections.Generic;
     using Newtonsoft.Json;
-    using Stripe.Infrastructure;
 
     public class SessionCreateOptions : BaseOptions, IHasMetadata
     {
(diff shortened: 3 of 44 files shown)


src/Stripe.net/Entities/BankAccounts/BankAccount.cs
```diff
@@ -73,7 +75,7 @@ public Customer Customer
         /// <summary>
         /// Whether this object is deleted or not.
         /// </summary>
-        [JsonProperty("deleted", NullValueHandling=NullValueHandling.Ignore)]
+        [JsonProperty("deleted", NullValueHandling = NullValueHandling.Ignore)]
```
remi-stripe, line 78: cc @ob-stripe just in case we care about this

src/Stripe.net/Services/Checkout/Sessions/SessionCreateOptions.cs
```diff
@@ -30,20 +28,20 @@ public class SessionCreateOptions : BaseOptions, IHasMetadata
...
         /// </summary>
         [JsonProperty("customer")]
         public string Customer { get; set; }
 
+        /// <summary>
+        /// The email address used to create the customer object. If you already know your
+        /// customer’s email address, use this attribute to prefill it on Checkout.
+        /// </summary>
+        [JsonProperty("customer_email")]
+        public string CustomerEmail { get; set; }
```
remi-stripe, line 43: cc @ob-stripe I never understand whether `customer_email` / `CustomerEmail` comes before/after `customer`/`Customer`
Fairly sure that in Go I chose to do the opposite.

src/Stripe.net/Services/Checkout/Sessions/SessionService.cs
```diff
@@ -35,14 +32,14 @@ public virtual Task<Session> CreateAsync(SessionCreateOptions options, RequestOp
             return this.CreateEntityAsync(options, requestOptions, cancellationToken);
         }
 
-        public virtual Session Get(string sessionId, SessionGetOptions options = null, RequestOptions requestOptions = null)
+        public virtual Session Get(string id, SessionGetOptions options = null, RequestOptions requestOptions = null)
```
remi-stripe, line 35: You likely have to undo this one change. We've discussed this in the past that it was a breaking change to do this even if we can't imagine anyone relying on the param name.

src/Stripe.net/Services/CreditNotes/CreditNoteService.cs
```diff
@@ -33,14 +33,14 @@ public virtual Task<CreditNote> CreateAsync(CreditNoteCreateOptions options, Req
             return this.CreateEntityAsync(options, requestOptions, cancellationToken);
         }
 
-        public virtual CreditNote Get(string creditNoteId, CreditNoteGetOptions options = null, RequestOptions requestOptions = null)
+        public virtual CreditNote Get(string id, CreditNoteGetOptions options = null, RequestOptions requestOptions = null)
```
remi-stripe, line 36: Same as above

src/Stripe.net/Entities/AccountLinks/AccountLink.cs
```diff
@@ -1,7 +1,6 @@
 namespace Stripe
 {
     using System;
-    using System.Collections.Generic;
```
remi-stripe, line 4: @ob-stripe Just wanted to make sure that this can't be a breaking change right?

src/Stripe.net/Services/Checkout/Sessions/SessionCreateOptions.cs
cjavilla-stripe: This is the sort order using `Array#sort` in javascript:

```js
['customer_email', 'customer', 'customer_phone'].sort((a, b) => a.localeCompare(b))
// => ["customer", "customer_email", "customer_phone"]
```

Which I think is what we want as that's what'll be used when ordering with autogen.

src/Stripe.net/Services/Checkout/Sessions/SessionService.cs
cjavilla-stripe: forgot this was potentially breaking, revered these param name changes

src/Stripe.net/Entities/BankAccounts/BankAccount.cs
ob-stripe: As long as it passes the StyleCop checks, it's fine.

src/Stripe.net/Services/Checkout/Sessions/SessionCreateOptions.cs
ob-stripe: Yeah, this is just how strings are sorted. If it helps, you can think of the null terminator byte as part of the string for sorting purposes. `customer\x00` should appear before `customer_email\x00`.

src/Stripe.net/Services/Checkout/Sessions/SessionService.cs
ob-stripe: 👍

src/Stripe.net/Entities/AccountLinks/AccountLink.cs
ob-stripe: No, if it compiles then it's fine. I wish there was a StyleCop rule to force us to remove unnecessary imports.

src/Stripe.net/Services/CreditNotes/CreditNoteService.cs
ob-stripe: 👍

PR #2058, opened 2020-05-29
Add support for BACS Debit as a `PaymentMethod`

remi-stripe:
r? @cjavilla-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails.cs
+++ b/src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails.cs
@@ -12,13 +12,16 @@ namespace Stripe
         public ChargePaymentMethodDetailsAchDebit AchDebit { get; set; }
 
         [JsonProperty("acss_debit")]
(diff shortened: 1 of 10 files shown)


src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails.cs
```diff
@@ -12,14 +12,17 @@ public class ChargePaymentMethodDetails : StripeEntity<ChargePaymentMethodDetail
         public ChargePaymentMethodDetailsAchDebit AchDebit { get; set; }
 
         [JsonProperty("acss_debit")]
-        public ChargePaymentMethodDetailsAchDebit AcssDebit { get; set; }
+        public ChargePaymentMethodDetailsAcssDebit AcssDebit { get; set; }
```
remi-stripe, line 15: Fixed this one which could break deserialization. It's technically a breaking change but it would unblock users using ACSS Debit

cjavilla-stripe: ah dang! good catch.

PR #2062, opened 2020-06-03
Add support for Bancontact, EPS, Giropay and P24 on `PaymentMethod`

remi-stripe:
r? @cjavilla-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/PaymentMethods/PaymentMethod.cs
+++ b/src/Stripe.net/Entities/PaymentMethods/PaymentMethod.cs
@@ -7,27 +7,57 @@ namespace Stripe
 
     public class PaymentMethod : StripeEntity<PaymentMethod>, IHasId, IHasMetadata, IHasObject
     {
+        /// <summary>
+        /// Unique identifier for the object.
+        /// </summary>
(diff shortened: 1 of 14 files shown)


src/Stripe.net/Entities/PaymentMethods/PaymentMethod.cs
```diff
@@ -56,24 +89,64 @@ public Customer Customer
         internal ExpandableField<Customer> InternalCustomer { get; set; }
         #endregion
 
+        /// <summary>
+        /// Properties specific to Eps PaymentMethod.
```
cjavilla-stripe, line 93: non blocking nit, but it looks like we call this EPS in the docs rather than Eps?

cjavilla-stripe: We'll fix all of these with autogen anyways.

remi-stripe: yeah I think it doesn't really matter though I tried to use the right writing everywhere but I typed that string a lot by hand :p

PR #2063, opened 2020-06-03
Add support for BACS Debit as a `Capability` on `Account`

remi-stripe:
r? @cjavilla-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/Accounts/AccountCapabilities.cs
+++ b/src/Stripe.net/Entities/Accounts/AccountCapabilities.cs
@@ -13,6 +13,13 @@ namespace Stripe
         public string AuBecsDebitPayments { get; set; }
 
         /// <summary>
+        /// The status of the Bacs Direct Debits payments capability of the account, or whether the
+        /// account can directly process Bacs Direct Debits charges.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/Stripe.net/Entities/Accounts/AccountCapabilities.cs
```diff
@@ -12,6 +12,13 @@ public class AccountCapabilities : StripeEntity<AccountCapabilities>
         [JsonProperty("au_becs_debit_payments")]
         public string AuBecsDebitPayments { get; set; }
 
+        /// <summary>
+        /// The status of the Bacs Direct Debits payments capability of the account, or whether the
```
cjavilla-stripe, line 16: wonder if we want to add the values in the comment here? `active`, `inactive`, or `pending`? non blocking.

cjavilla-stripe: I guess that's the same for all of these so maybe doesn't make sense to add to all.

remi-stripe: Yeah it'd be repeated and the alignment of the comments is often different based on the previous sentence so it's a pain to do by hand :p

PR #2094, opened 2020-07-01
More no-op changes to match what is autogennerated

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

I'm some how generating 800+ files, but still not touching like 100ish so trying to track down where those live in the open api spec tree that I'm not hitting. This is fixing up the imports and alphabetizing the ones I recently generated.


remi-stripe:
Thanks for all the fixes!

PR #2104, opened 2020-07-01
Prepare future migration to Newtonsoft.Json's UnixDateTimeConverter

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Also cc @cjavilla-stripe / @richardm-stripe  since this will require a (trivial) change to codegen.

As of v11.0, Newtonsoft.Json provides a [`UnixDateTimeConverter`](https://github.com/JamesNK/Newtonsoft.Json/blob/666d9760719e5ec5b2a50046f7dbd6a1267c01c6/Src/Newtonsoft.Json/Converters/UnixDateTimeConverter.cs) class that does what our current custom `DateTimeConverter` does (with a small behavior difference, more on that below).

Once we make 11.0 the minimum required version for all runtimes, we will be able to remove our custom converter and use the prebuilt class. This PR paves the way for that future migration by:
1. renaming the converter from `DateTimeConverter` to `UnixDateTimeConverter`
2. changing the implementation to be the same as Newtonsoft.Json's

For normal usage of the library, this should be completely transparent.

The behavior change I mentioned is as follows: the default value for `DateTime` is January 1, 0001 00:00:00. This date cannot be encoded as a Unix timestamp, as it's 1969 years too early 😂  (the Unix epoch is January 1, 1970, 00:00:00).

So previously, if you did this:

```c#
var customer = new Stripe.Customer {Id = "cus_123"};
var json = customer.ToJson();
```

`json` would contain something like this: `{"id": "cus_123", ..., "created": 0, ...}` (because of [this](https://github.com/stripe/stripe-dotnet/blob/b62c975a0ae085d89be49354eb1dd71bbcbc52ad/src/Stripe.net/Infrastructure/EpochTime.cs#L16-L19)).

With this change, calling `ToJson()` will raise a `JsonSerializationException` with the message "Cannot convert date value that is before Unix epoch of 00:00:00 UTC on 1 January 1970." This is more correct, as the current behavior means that a serialization/deserialization roundtrip would return different values (January 1, 0001 -> `0` -> January 1, 1970).

This is a breaking change, but it would only affect users that are 1/ manually creating `StripeEntity` instances with `DateTime` props, 2/ without providing valid values for the `DateTime` props` and 3/ serializing the instance to JSON.


cjavilla-stripe:
Ack. Thanks for the heads up for autogen. Simple change once this is in.

ob-stripe:
@richardm-stripe Gentle bump

richardm-stripe:
This LGTM, but
r? @cjavilla-stripe 
for the codegen stuff
(and the present merge conflicts)

cjavilla-stripe:
LGTM codegen will be a super quick fix that I can run once this is in. in terms of merge conflicts, lmk if you want me to help with those, @ob-stripe :)

ob-stripe:
I rebased to fix the conflicts and merged this to the integration branch.

One thought, if users find it annoying to have to explicitly initialize all `DateTime` fields when manually constructing Stripe models: we could initialize the fields with the Unix epoch ourselves, e.g. replace all:

```c#
public DateTime Created { get; set; }
```

with:

```c#
public DateTime Created { get; set; } = DateTime.UnixEpoch;
```

cjavilla-stripe:
I like that idea, @ob-stripe. What would we do in the case of something like `AnyOf<DateTime?, SubscriptionTrialEnd> TrialEnd { get; set; }`?

cjavilla-stripe:
or maybe that only applies to entities and not options?

ob-stripe:
Users might want to serialize options classes to JSON as well, but it's not going to be an issue there, as all properties in options classes are always nullable. The issue about `DateTime` not being encodable to a Unix timestamp only applies to non-nullable `DateTime` fields (nullable `DateTime` fields would just be initialized to `null`, not January 1st, 0001).

PR #2113, opened 2020-07-12
Adds several breaking changes to bring inline with the spec

cjavilla-stripe:
- [x] ⚠️ Removes metadata from `DisputeEvidenceOptions` (not supported)
- [x] ⚠️ Updates `InvoiceListOptions` to use `ListOptionsWithCreated` base and uses created options from base rather than concrete class. Also removes `paid` from `InvoiceListOptions` (use status instead)
- [x] ⚠️ Moves `notification_method` from `SourceMandateAcceptanceOptions` to `SourceMandateOptions`

r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Services/Disputes/DisputeEvidenceOptions.cs
+++ b/src/Stripe.net/Services/Disputes/DisputeEvidenceOptions.cs
@@ -1,9 +1,8 @@
 namespace Stripe
 {
-    using System.Collections.Generic;
     using Newtonsoft.Json;
 
-    public class DisputeEvidenceOptions : INestedOptions, IHasMetadata
+    public class DisputeEvidenceOptions : INestedOptions
     {
         [JsonProperty("access_activity_log")]
         public string AccessActivityLog { get; set; }
@@ -43,9 +42,6 @@ namespace Stripe
 
         [JsonProperty("duplicate_charge_id")]
         public string DuplicateChargeId { get; set; }
-
-        [JsonProperty("metadata")]
-        public Dictionary<string, string> Metadata { get; set; }
 
         [JsonProperty("product_description")]
         public string ProductDescription { get; set; }
--- a/src/Stripe.net/Services/Invoices/InvoiceListOptions.cs
+++ b/src/Stripe.net/Services/Invoices/InvoiceListOptions.cs
@@ -4,21 +4,13 @@ namespace Stripe
     using Newtonsoft.Json;
     using Stripe.Infrastructure;
 
-    public class InvoiceListOptions : ListOptions
+    public class InvoiceListOptions : ListOptionsWithCreated
     {
         /// <summary>
         /// Either <c>charge_automatically</c>, or <c>send_invoice</c>.
         /// </summary>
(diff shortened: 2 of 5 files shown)


src/Stripe.net/Services/Orders/OrderReturnOrderOptions.cs
```diff
@@ -3,7 +3,7 @@ namespace Stripe
     using System.Collections.Generic;
     using Newtonsoft.Json;
 
-    public class OrderReturnOptions : BaseOptions
```
remi-stripe, line 6: Why not just make an override here instead of costing a rename?

src/Stripe.net/Services/Orders/OrderService.cs
```diff
@@ -73,12 +73,12 @@ public virtual Task<Order> PayAsync(string id, OrderPayOptions options = null, R
             return this.RequestAsync(HttpMethod.Post, $"{this.InstanceUrl(id)}/pay", options, requestOptions, cancellationToken);
         }
 
-        public virtual OrderReturn Return(string id, OrderReturnOptions options = null, RequestOptions requestOptions = null)
```
remi-stripe, line 76: Same as above, why not an override versus costing a rename to users?

src/Stripe.net/Services/SetupIntents/SetupIntentConfirmOptions.cs
```diff
@@ -4,13 +4,6 @@ namespace Stripe
 
     public class SetupIntentConfirmOptions : BaseOptions
     {
-        /// <summary>
-        /// The client secret of the SetupIntent. Required if a publishable key is used to
-        /// confirm the setup intent.
-        /// </summary>
-        [JsonProperty("client_secret")]
-        public string ClientSecret { get; set; }
```
remi-stripe, line 12: This needs to stay here. This is how you use stripe-dotnet to confirm a PI or SI in a mobile app right?

cjavilla-stripe: As in using stripe-dotnet in the mobile app with xamarin or something?

remi-stripe: Yes

cjavilla-stripe: ack will add with an override :)

cjavilla-stripe:
Thanks, I reverted the order return and setup intent changes. 

ptal @remi-stripe

src/Stripe.net/Services/SetupIntents/SetupIntentConfirmOptions.cs
remi-stripe: It's still removed incorrect right?

cjavilla-stripe: woops yeah didn't push the right thing. should be fixed now.

PR #2115, opened 2020-07-13
Use string `failure_reason` for `SourceRedirect` and remove several fields not in the OpenAPI spec

cjavilla-stripe:
Several breaking API changes to bring inline with OpenAPI spec:
  * ⚠️ Removes `bitcoin` from `Source`
  * ⚠️ Deletes `SourceBitcoin`
  * ⚠️ Changes `failure_reason` on `SourceRedirect` to return `string` rather than `RedirectFailureReason` enum
  * ⚠️ Deletes `RedirectFailureReason`
  * ⚠️ Removes `dashboard ` from `AccountSettingsOptions` 
  * ⚠️ Removes `usage` from `SetupIntentUpdateOptions`
  * ⚠️ Removes `invoice_settings` from `SubscriptionScheduleCreateOptions` and `SubscriptionScheduleUpdateOptions`

r? @remi-stripe 
cc @stripe/api-libraries


src/Stripe.net/Services/Accounts/AccountSettingsOptions.cs
```diff
@@ -10,9 +10,6 @@ public class AccountSettingsOptions : INestedOptions
         [JsonProperty("card_payments")]
         public AccountSettingsCardPaymentsOptions CardPayments { get; set; }
 
-        [JsonProperty("dashboard")]
-        public AccountSettingsDashboardOptions Dashboard { get; set; }
-
```
remi-stripe, line 15: you forgot to remove that class from the library right?

cjavilla-stripe: Yep. I forgot that one. great catch! ptal

src/Stripe.net/Services/SubscriptionSchedules/SubscriptionScheduleUpdateOptions.cs
```diff
@@ -19,12 +19,6 @@ public class SubscriptionScheduleUpdateOptions : BaseOptions, IHasMetadata
         [JsonProperty("end_behavior")]
         public string EndBehavior { get; set; }
 
-        /// <summary>
-        /// Define the default settings applied to invoices created by this subscription schedule.
-        /// </summary>
-        [JsonProperty("invoice_settings")]
-        public SubscriptionScheduleInvoiceSettingsOptions InvoiceSettings { get; set; }
-
```
remi-stripe, line 27: same here you forgot to remove the class too right?

cjavilla-stripe: So that one is used in `SubscriptionScheduleDefaultSettingsOptions` and `SubscriptionSchedulePhaseOptions` should that name be different?

remi-stripe: Yeah it should be `SubscriptionSchedulePhaseInvoiceSettingsOptions` in that case. But not worth renaming for now I'd say

PR #2116, opened 2020-07-14
Multiple API changes

cjavilla-stripe:
Multiple API changes:
  * Makes `Obsolete` helper text consistent
  * ⚠️ Removes several unsupported properties
  * Adds empty Options classes to make service methods more consistent

r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/Charges/ChargePaymentMethodDetailsInteracPresentReceipt.cs
+++ b/src/Stripe.net/Entities/Charges/ChargePaymentMethodDetailsInteracPresentReceipt.cs
@@ -1,4 +1,4 @@
-﻿namespace Stripe
+namespace Stripe
 {
     using Newtonsoft.Json;
 
--- a/src/Stripe.net/Services/Disputes/DisputeCloseOptions.cs
+++ b/src/Stripe.net/Services/Disputes/DisputeCloseOptions.cs
@@ -1,0 +1,6 @@
+namespace Stripe
+{
+    public class DisputeCloseOptions : BaseOptions
+    {
+    }
+}
--- a/src/Stripe.net/Services/Disputes/DisputeService.cs
+++ b/src/Stripe.net/Services/Disputes/DisputeService.cs
@@ -22,14 +22,14 @@ namespace Stripe
 
         public override string BasePath => "/v1/disputes";
 
-        public virtual Dispute Close(string id, RequestOptions requestOptions = null)
+        public virtual Dispute Close(string id, DisputeCloseOptions options = null, RequestOptions requestOptions = null)
         {
-            return this.Request(HttpMethod.Post, $"{this.InstanceUrl(id)}/close", null, requestOptions);
+            return this.Request(HttpMethod.Post, $"{this.InstanceUrl(id)}/close", options, requestOptions);
         }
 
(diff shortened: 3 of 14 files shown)


src/Stripe.net/Services/Customers/CustomerCreateOptions.cs
```diff
@@ -108,9 +108,6 @@ public class CustomerCreateOptions : BaseOptions, IHasMetadata
         [JsonProperty("tax_id_data")]
         public List<CustomerTaxIdDataOptions> TaxIdData { get; set; }
 
-        [JsonProperty("tax_percent")]
```
remi-stripe, line 111: That one is not a minor change, we should be really careful shipping this changed without a clear understanding of how many users on the latest stripe-dotnet major are passing it. Can you investigate this first?

src/Stripe.net/Services/Disputes/DisputeCloseOptions.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Stripe
```
remi-stripe, line 1: Try to separate your PRs by conceptual changes. Adding those new classes is wildly different from removing `tax_percent` and it makes it hard to pinpoint the root cause of the changes

src/Stripe.net/Services/Disputes/DisputeService.cs
```diff
@@ -22,14 +22,14 @@ public DisputeService(IStripeClient client)
 
         public override string BasePath => "/v1/disputes";
 
-        public virtual Dispute Close(string id, RequestOptions requestOptions = null)
+        public virtual Dispute Close(string id, DisputeCloseOptions options = null, RequestOptions requestOptions = null)
```
remi-stripe, line 25: @ob-stripe I never know if this change is a breaking change

src/Stripe.net/Services/Customers/CustomerCreateOptions.cs
cjavilla-stripe: I was only able to find 5 requests in the last 2 weeks for anyone using the past 3 majors.

remi-stripe: Ack so we should remove this one on all resources/params for good in that case :)

src/Stripe.net/Entities/Charges/ChargePaymentMethodDetailsInteracPresentReceipt.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace Stripe
+namespace Stripe
```
cjavilla-stripe, line 1: there was a bom at the top of this file that I removed

src/Stripe.net/Services/Disputes/DisputeService.cs
cjavilla-stripe: I think this is a breaking change since we're adding a new argument. If someone was calling this before with request options as the second arg, it'll fail because we're changing the type for the second arg.

cjavilla-stripe:
I think that is the only outstanding question and it's a breaking change, but should only be breaking if people are already using request options for those calls.

PR #2121, opened 2020-07-18
Add support for multiple coupons on Billing APIs

remi-stripe:
Add support for multiple coupons on Billing APIs
  * Add support for arrays of expandable API resources otherwise returning an array of strings by default
  * Add support for `Id`, `Invoice` and `InvoiceItem` on `Discount`.
  * Add support for `Discounts` on `Invoice`, `InvoiceItem`, `InvoiceLineItem`
  * Add support for `DiscountAmounts` on `CreditNote`, `CreditNoteLineItem`, `InvoiceLineItem`
  * Add support for `TotalDiscountAmounts` on `Invoice`

This PR supersedes https://github.com/stripe/stripe-dotnet/pull/2016 but used @ob-stripe's original commit as a base
It also mirrors https://github.com/stripe/stripe-java/pull/1070 in java

r? @richardm-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/StripeTests/Services/Accounts/AccountServiceTest.cs
+++ b/src/StripeTests/Services/Accounts/AccountServiceTest.cs
@@ -34,6 +34,17 @@ namespace StripeTests
                     Name = "business name",
                 },
                 BusinessType = "company",
+                Capabilities = new AccountCapabilitiesOptions
+                {
+                    CardPayments = new AccountCapabilitiesCardPaymentsOptions
+                    {
+                        Requested = true,
+                    },
+                    Transfers = new AccountCapabilitiesTransfersOptions
+                    {
+                        Requested = true,
+                    },
+                },
                 Company = new AccountCompanyOptions
                 {
                     Address = new AddressOptions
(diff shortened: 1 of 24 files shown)


src/StripeTests/Services/Accounts/AccountServiceTest.cs
```diff
@@ -34,6 +34,17 @@ public class AccountServiceTest : BaseStripeTest
                     Name = "business name",
                 },
                 BusinessType = "company",
+                Capabilities = new AccountCapabilitiesOptions
```
richardm-stripe, line 37: Is this change necessary with the bump of stripe-mock version? (Just wondering why it's present in this PR)

remi-stripe: Yes! `RequestedCapabilities` is deprecated and was replaced with `Capabilities` last week so when we bumped stripe-mock` we had to make that change and change the `Type` on `AccountLink` too

PR #2125, opened 2020-07-21
Adds `TokenPiiOptions` for `TokenCreateOptions`

cjavilla-stripe:
* Adds `TokenPiiOptions` for `TokenCreateOptions`

r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Services/Tokens/TokenPiiOptions.cs
+++ b/src/Stripe.net/Services/Tokens/TokenPiiOptions.cs
@@ -1,0 +1,13 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+
+    public class TokenPiiOptions : INestedOptions
+    {
+        /// <summary>
+        /// The <c>IdNumber</c> for the PII, in string form.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


src/Stripe.net/Services/Tokens/TokenPiiOptions.cs
```diff
@@ -0,0 +1,13 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+
+    public class TokenPiiOptions : INestedOptions
+    {
+        /// <summary>
+        /// The <c>IdNumber</c> for the PII, in string form.
```
remi-stripe, line 8: why the `<c>` tags here? Shouldn't this just be `The personal id number for the PII` or similar? Feels like this is trying to reference another property elsewhere. Not blocking though

cjavilla-stripe: It was `id_number` so I pascal cased it, but again I'm not sure if we'll be able to do this with autogen anyways

PR #2132, opened 2020-07-22
Remove Bitcoin and ThreeDSecure

cjavilla-stripe:
* ⚠️ Remove Bitcoin and ThreeDSecure

r? @remi-stripe 
cc @stripe/api-libraries 

Only found 1 account making 4 requests to the /v1/3d_secure endpoint, with one of the last 3 majors, in livemode since the start of May. Didn't find any bitcoin requests. I _believe_ I removed the right ThreeDSecure classes and left the correct ones for Sources and PaymentMethod types, but if you could check that it would be helpful too, Remi.


src/Stripe.net/Constants/SourceType.cs
```diff
@@ -10,8 +10,6 @@ public static class SourceType
 
         public const string Bancontact = "bancontact";
 
-        public const string Bitcoin = "bitcoin";
```
remi-stripe, line 13: Flagging so that it's explicit and we should write down our thoughts here. Basically it means if you look at an old Source for example on an old Charge, then you can not get the source type or its properties.
I still think it's the right choice, but that's what has blocked us for so long.

src/Stripe.net/Entities/Charges/ChargePaymentMethodDetails.cs
```diff
@@ -25,9 +25,6 @@ public class ChargePaymentMethodDetails : StripeEntity<ChargePaymentMethodDetail
         [JsonProperty("bancontact")]
         public ChargePaymentMethodDetailsBancontact Bancontact { get; set; }
 
-        [JsonProperty("bitcoin")]
-        public ChargePaymentMethodDetailsBitcoin Bitcoin { get; set; }
```
remi-stripe, line 29: That one bothers me more, I think this should stay for now, can you fix it?

cjavilla-stripe:
Okay removed the PMD one. ptal @remi-stripe

ob-stripe:
@cjavilla-stripe Can you rebase the PR on the current integration branch?

PR #2137, opened 2020-07-24
Use Request rather than DeleteEntity for custom Cancel method

cjavilla-stripe:
* Use Request rather than DeleteEntity for custom Cancel method

r? @ob-stripe 
cc @stripe/api-libraries 

I don't think this is a breaking change, but since Cancel is a custom method, I want to render it like this for autogen :)


ob-stripe:
@cjavilla-stripe Can you rebase the PR on the current integration branch?

cjavilla-stripe:
done. this one should be a quickie.

PR #2139, opened 2020-07-24
Make MigratedTo and RolledBackFrom expandable on Recipient

cjavilla-stripe:
* Make `MigratedTo` and `RolledBackFrom` expandable on `Recipient`

r? @remi-stripe 
cc @stripe/api-libraries 

I'm not 100% sure if these are breaking or not. We're going from a string field to an expandable field. I don't believe it's breaking, but putting in integration-v38 can can label appropriately if so.


cjavilla-stripe:
closing this PR to instead remove recipient.

PR #2142, opened 2020-07-25
Cjavilla/unify pmd

cjavilla-stripe:
- [x] ⚠️ Stop sharing from `PaymentIntentPaymentMethodDataOptions`
- [x] ⚠️ Stop sharing sub hashes from `PaymentMethodCreateOptions` and `PaymentMethodUpdateOptions`

I did this in two commits to try to make the review a little easier. I think this one likely needs a bit bigger blurb in the change log.

r? @remi-stripe 
cc @stripe/api-libraries


src/Stripe.net/Services/PaymentMethods/PaymentMethodBancontactOptions.cs
```diff
@@ -0,0 +1,6 @@
+namespace Stripe
```
remi-stripe, line 1: it's weird it doesn't see this one as a rename

src/Stripe.net/Services/PaymentMethods/PaymentMethodCardOptions.cs
```diff
@@ -2,7 +2,7 @@ namespace Stripe
 {
     using Newtonsoft.Json;
 
-    public class PaymentMethodCardCreateOptions : INestedOptions
+    public class PaymentMethodCardOptions : INestedOptions
```
remi-stripe, line 5: For later it'd be need if we flagged params that are only on create (or vice versa) in the docstring. Not worth it but it'd be a good todo to track

cjavilla-stripe: yeah I think we could likely do that in the params somehow. Right now, I'm merging the params between create and update pretty far up-stream, so we'd likely need to pass down the state from pretty early on.

src/Stripe.net/Services/PaymentMethods/PaymentMethodBancontactOptions.cs
cjavilla-stripe: yeah I think between the class name changing, and removing the unneeded json using statement, it's different enough to git to think it's a new file 🤷

src/Stripe.net/Services/PaymentIntents/PaymentIntentPaymentMethodDataAuBecsDebitOptions.cs
```diff
@@ -0,0 +1,13 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+
+    public class PaymentIntentPaymentMethodDataAuBecsDebitOptions : INestedOptions
+    {
+        [JsonProperty("account_number")]
```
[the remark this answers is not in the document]
cjavilla-stripe: Added :)

PR #2145, opened 2020-07-25
Add support for `BillingThresholds` on `SubscriptionItemOptions`

cjavilla-stripe:
* Add support for `BillingThresholds` on `SubscriptionItemOptions`

r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Services/Subscriptions/SubscriptionItemOptions.cs
+++ b/src/Stripe.net/Services/Subscriptions/SubscriptionItemOptions.cs
@@ -6,6 +6,14 @@ namespace Stripe
     public class SubscriptionItemOptions : IHasId, INestedOptions, IHasMetadata
     {
         /// <summary>
+        /// Define thresholds at which an invoice will be sent, and the subscription advanced to a
+        /// new billing period. When updating, pass an empty string to remove previously-defined
+        /// thresholds.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/Stripe.net/Services/Subscriptions/SubscriptionItemOptions.cs
```diff
@@ -5,6 +5,14 @@ namespace Stripe
 
     public class SubscriptionItemOptions : IHasId, INestedOptions, IHasMetadata
     {
+        /// <summary>
+        /// Define thresholds at which an invoice will be sent, and the subscription advanced to a
+        /// new billing period. When updating, pass an empty string to remove previously-defined
+        /// thresholds.
+        /// </summary>
+        [JsonProperty("billing_thresholds")]
+        public SubscriptionItemBillingThresholdsOptions BillingThresholds { get; set; }
```
remi-stripe, line 14: technically you're introducing something you know you will break later with "no sharing" right?

cjavilla-stripe: yeah this one is really odd because the unshared and shared name is the same

cjavilla-stripe:
thanks yeah lets merge and we can figure out how to unshare later with some custom name, at least this matches the current naming.

PR #2154, opened 2020-08-06
Add default values for non-nullable `DateTime` properties

cjavilla-stripe:
* Add default values for non-nullable `DateTime` properties

r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/src/Stripe.net/Entities/AccountLinks/AccountLink.cs
+++ b/src/Stripe.net/Entities/AccountLinks/AccountLink.cs
@@ -17,14 +17,14 @@ namespace Stripe
         /// </summary>
         [JsonProperty("created")]
         [JsonConverter(typeof(UnixDateTimeConverter))]
-        public DateTime Created { get; set; }
+        public DateTime Created { get; set; } = Stripe.Infrastructure.DateTimeUtils.UnixEpoch;
 
         /// <summary>
         /// The timestamp at which this account link will expire.
         /// </summary>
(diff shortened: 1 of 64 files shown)


src/Stripe.net/Entities/AccountLinks/AccountLink.cs
```diff
@@ -17,14 +17,14 @@ public class AccountLink : StripeEntity<AccountLink>, IHasObject
         /// </summary>
         [JsonProperty("created")]
         [JsonConverter(typeof(UnixDateTimeConverter))]
-        public DateTime Created { get; set; }
+        public DateTime Created { get; set; } = new DateTime(1970, 1, 1, 0, 0, 0);
```
ob-stripe, line 20: We should not pass a value directly. Instead, we should define our own constant somewhere, e.g. in a `Stripe.Infrastructure.DateTimeUtils` class:
```c#
public static readonly DateTime UnixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```

then initialize values with `= Stripe.Infrastructure.DateTimeUtils.UnixEpoch;`

(Also maybe a leave a comment to just use `DateTime.UnixEpoch` once we bump the minimum supported version to .NET Standard 2.1`)

cjavilla-stripe:
hows that @ob-stripe ? :) Thanks again for looking!

PR #2157, opened 2020-08-06
Moves expandable discounts to bottom of class because they are overridden

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries


cjavilla-stripe:
Should not be overrides

PR #2172, opened 2020-08-09
Unshare and rename invoice objects

cjavilla-stripe:
- [x] ⚠️ Unshare and rename `Invoice` options classes and service to align with spec.

Service method name for nested methods like /v1/invoices/{id}/lines are following the pattern:

ListInvoiceLineItems
`{action}{returnType}`

and in the case of `/v1/invoices/upcoming/lines` we use the following pattern

UpcomingListInvoiceLineItems
`{action}{returnType}`

The options classes for these methods are:

InvoiceListInvoiceLineItemsOptions
`{parent}{action}{returnType}Options`

InvoiceUpcomingListInvoiceLineItemsOptions
`{parent}{action}{returnType}Options`

r? @remi-stripe 
cc @stripe/api-libraries


src/Stripe.net/Services/Invoices/InvoiceFinalizeInvoiceOptions.cs
```diff
@@ -1,11 +1,8 @@
 namespace Stripe
 {
-    using System;
-    using System.Collections.Generic;
     using Newtonsoft.Json;
-    using Stripe.Infrastructure;
 
-    public class InvoiceFinalizeOptions : BaseOptions
+    public class InvoiceFinalizeInvoiceOptions : BaseOptions
```
remi-stripe, line 5: I'm still on the fence about that one. I wonder if we shouldn't just go back to `InvoiceFinalizeOptions` and rename the method `Finalize` (and same for all the others)

src/Stripe.net/Services/Invoices/InvoiceSendInvoiceOptions.cs
```diff
@@ -0,0 +1,6 @@
+namespace Stripe
```
remi-stripe, line 1: Same as `Finalize` (won't flag the others)

src/Stripe.net/Services/Invoices/InvoiceUpcomingListInvoiceLineItemsOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+    public class InvoiceUpcomingListInvoiceLineItemsOptions : ListOptions
+    {
+        [JsonProperty("coupon")]
+        public string Coupon { get; set; }
+
+        [JsonProperty("customer")]
+        public string Customer { get; set; }
+
+        [JsonProperty("discounts")]
+        public List<InvoiceLineItemDiscountOptions> Discounts { get; set; }
```
remi-stripe, line 17: So this name bothers me for two reasons.

First, it could conflict with other options on the Create/Update path. That's the "problem" with upcoming because it has its own set of parameters/hashes but because we "share" everything that is a nested hash they can conflict. It's definitely not the end of the world and likely the right decision but damn this is definitely confusing.

Second, this is an option on Upcoming invoice, but you name it as if it's an option on InvoiceLineItem which _feels_ incorrect. Shouldn't it be `<parent><child>Options`? Like if we shipped a `createPerson` on `AccountService`, wouldn't you call the option `AccountPersonCreateOptions` or `AccountCreatePersonOptions`?

cjavilla-stripe:
Punting renaming this stuff until we decide what to do long term with nested service methods. Going to override code to match what we already have.

PR #2174, opened 2020-08-10
Rename custom methods on InvoiceService

cjavilla-stripe:
This is one of the proposed approaches for the naming of custom methods on the InvoiceService.

We need to decide whether to

:one: Keep the method named `FinalizeInvoice` and change the options class to `InvoiceFinalizeInvoiceOptions`
:two: Keep the options class as `InvoiceFinalizeOptions` and  rename the method `FinalizeInvoice` to `Finalize`.

r? @remi-stripe 
cc @stripe/api-libraries 
cc @richardm-stripe @ob-stripe


cjavilla-stripe:
Punting renaming this stuff until we decide what to do long term with nested service methods. Going to override code to match what we already have.

PR #2177, opened 2020-08-10
Adding fields removed from SourceCardOptions in merge

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

When I merged SourceCard with CreditCard options some of this was dropped, then picked back up by a bug fix.

### DIFF
--- a/src/Stripe.net/Services/Sources/SourceCardOptions.cs
+++ b/src/Stripe.net/Services/Sources/SourceCardOptions.cs
@@ -1,8 +1,9 @@
 namespace Stripe
 {
+    using System.Collections.Generic;
     using Newtonsoft.Json;
 
-    public class SourceCardOptions : INestedOptions
+    public class SourceCardOptions : INestedOptions, IHasMetadata
     {
         [JsonProperty("address_city")]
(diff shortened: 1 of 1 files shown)


src/Stripe.net/Services/Sources/SourceCardOptions.cs
```diff
@@ -25,6 +26,9 @@ public class SourceCardOptions : INestedOptions
         [JsonProperty("currency")]
         public string Currency { get; set; }
 
+        [JsonProperty("default_for_currency")]
+        public bool? DefaultForCurrency { get; set; }
+
```
remi-stripe, line 31: Why is this here? `default_for_currency` is for *external accounts* so it doesn't match a class named `SourceCardOptions`

cjavilla-stripe: Ah, added that as an override because it was showing as removed from something but I can't find which so will remove.

cjavilla-stripe: Removed

PR #2212, opened 2020-09-30
`subscription_prorate` is no longer supported on UpcomingInvoiceListL…

topherdavis:
`subscription_prorate` is no longer supported on UpcomingInvoiceListLineItemsOptions

This has been replaced with `subscription_proration_behavior`. This is updated here.

### DIFF
--- a/src/Stripe.net/Services/Invoices/UpcomingInvoiceListLineItemsOptions.cs
+++ b/src/Stripe.net/Services/Invoices/UpcomingInvoiceListLineItemsOptions.cs
@@ -42,8 +42,8 @@ namespace Stripe
         [JsonProperty("subscription_items")]
         public List<InvoiceSubscriptionItemOptions> SubscriptionItems { get; set; }
 
-        [JsonProperty("subscription_prorate")]
-        public bool? SubscriptionProrate { get; set; }
+        [JsonProperty("subscription_proration_behavior")]
+        public string SubscriptionProrationBehavior { get; set; }
 
         [JsonProperty("subscription_proration_date")]
         [JsonConverter(typeof(UnixDateTimeConverter))]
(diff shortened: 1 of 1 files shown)


src/Stripe.net/Services/Invoices/UpcomingInvoiceListLineItemsOptions.cs
```diff
@@ -42,8 +42,8 @@ public class UpcomingInvoiceListLineItemsOptions : ListOptions
         [JsonProperty("subscription_items")]
         public List<InvoiceSubscriptionItemOptions> SubscriptionItems { get; set; }
 
-        [JsonProperty("subscription_prorate")]
-        public bool? SubscriptionProrate { get; set; }
+        [JsonProperty("subscription_proration_behavior")]
+        public string SubscriptionProrationBehavior { get; set; }
```
remi-stripe, line 46: @topherdavis Could you keep the original one and add the new one instead? Otherwise this is a breaking change for developers that still use the old parameter and it would need a major version.

The best option is to do this:

```
        [Obsolete("Use SubscriptionProrationBehavior instead.")]
        [JsonProperty("subscription_prorate")]
        public bool? SubscriptionProrate { get; set; }

        [JsonProperty("subscription_proration_behavior")]
        public string SubscriptionProrationBehavior { get; set; }
```

remi-stripe (commented):
@topherdavis Thanks a lot for doing the work here. While most the library is automatically generated from our openapi spec, this file (and a few others) aren't which is why it slipped through the cracks.

We would need some change on your PR to merge this. Let me know if you want to take of if you prefer I handle directly.

src/Stripe.net/Services/Invoices/UpcomingInvoiceListLineItemsOptions.cs
topherdavis: Thanks, makes sense. I wondered why SubscriptionProrate was still on `UpcomingInvoiceOptions`. Updated.

remi-stripe (approved):
Perfect thank you so much for the PR and for taking the time to edit it!

remi-stripe:
I just released this as 39.10.0, thanks again for taking the work yourself!

topherdavis:
Thanks for releasing it!

PR #2239, opened 2020-10-29
Avoid async in autopagination when NET461

### LINKED ISSUE #2223: ListAutoPaging requires async capabilities?
Using webforms (I know.. I know...), our pages are setup to post back to the server and we have not enabled any async capabilities on the page. The page's code-behind is calling 

```
var invoiceService = new InvoiceService(_client);
var options = new InvoiceListOptions()
{
       Limit = 50,
       Customer = customerId,
};

var invoices = invoiceService.ListAutoPaging(options).ToList();
return invoices;
```

Using Fidler I can see that the service is contacting Stripe and returning 50 records. 
The system just spins and does nothing until the thread expires if there are more than 50 records, despite the returned JSON property "has_more" = true. 

I believe this is because the underlying/base "ListAutoPaging" feature has async dependencies, as found here,
https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/_base/Service.cs#L260

Thus, the service works in .net web-forms, but never completes because the async services stall out when web-forms are not setup to support async. 

I will work around this with our own do/while loop, but wanted to raise this for others and verify I'm not missing anything?

Using Visual Studio Version 16.7.3
.NET v4.7.2, C#

richardm-stripe:
Fixes #2223.

Uses a preprocessor directive so that `ListRequestAutoPaging` avoids trying to be async in NET461 (which breaks webforms), and uses something identical to the previous (sync) iteration: https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/_base/Service.cs#L260

Thanks @ob-stripe for pairing with me on the implementation of this.

r? @ob-stripe 
cc @remi-stripe

### DIFF
--- a/src/Stripe.net/Services/_base/Service.cs
+++ b/src/Stripe.net/Services/_base/Service.cs
@@ -263,10 +263,84 @@ namespace Stripe
             RequestOptions requestOptions)
             where T : IStripeEntity
         {
+#if NET461
+            return
+                this.ListRequestAutoPagingSync<T>(url, options, requestOptions);
+#else
             return AsyncUtils.ToEnumerable(
                 this.ListRequestAutoPagingAsync<T>(url, options, requestOptions));
+#endif
         }
 
+#if NET461
+        protected IEnumerable<T> ListRequestAutoPagingSync<T>(
+            string url,
+            ListOptions options,
+            RequestOptions requestOptions)
+            where T : IStripeEntity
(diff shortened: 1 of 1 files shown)


ob-stripe (approved):

mexner:
Thanks @richardlawley @ob-stripe !

PR #2246, opened 2020-11-10
Add .NET 5.0 target

ob-stripe:
r? @richardm-stripe 

[.NET 5.0](https://devblogs.microsoft.com/dotnet/announcing-net-5-0/) was released today. Microsoft [recommends](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/#what-you-should-target) that reusable libraries add it as a build target but still keep targeting .NET Standard 2.0 as well for backwards compatibility.

### DIFF
--- a/src/Stripe.net/Infrastructure/RuntimeInformation.cs
+++ b/src/Stripe.net/Infrastructure/RuntimeInformation.cs
@@ -21,13 +21,15 @@ namespace Stripe.Infrastructure
 
         internal static bool IsFullFramework => FrameworkDescription.StartsWith(".NET Framework", StringComparison.OrdinalIgnoreCase);
 
-        internal static bool IsNetCore => FrameworkDescription.StartsWith(".NET Core", StringComparison.OrdinalIgnoreCase) && !string.IsNullOrEmpty(typeof(object).Assembly.Location);
+        internal static bool IsNetCore
+            => ((Environment.Version.Major >= 5) || FrameworkDescription.StartsWith(".NET Core", StringComparison.OrdinalIgnoreCase))
+                && !string.IsNullOrEmpty(typeof(object).Assembly.Location);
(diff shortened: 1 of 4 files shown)


richardm-stripe (dismissed):
Thank you!

ob-stripe:
It looks like the 5.0 SDK included in Appveyor's "Visual Studio 2019 Preview" image is still the RC1 version and not the final version, so I'll wait until they update it to the final version before merging.

ob-stripe:
Appveyor added .NET 5.0 to the Visual Studio 2019 image (cf. https://www.appveyor.com/updates/2020/11/14/).

I've also updated the `RuntimeInformation` code to properly report .NET 5 as `".NET 5.0.0"` rather than `".NET Core 5.0.0"`. Like the initial code, the updated code is basically lifted straight from the BenchmarkDotnet project (https://github.com/dotnet/BenchmarkDotNet/blob/0de41c2864de41fef983e55790490f1b3364ccbf/src/BenchmarkDotNet/Portability/RuntimeInformation.cs)

ptal @richardm-stripe

src/Stripe.net/Infrastructure/RuntimeInformation.cs
```diff
@@ -110,27 +112,7 @@ internal static string GetFullFrameworkVersion()
...
-            {
-                return "4.5.2";
-            }
-
-            if (string.Compare(servicingVersion, "4.6.1") < 0)
-            {
-                return "4.6";
-            }
-
+            // the following code assumes that .NET 4.6.1 is the oldest supported version
```
ob-stripe, line 115: Unrelated, but since 4.6.1 is the minimum required version now we can remove a bunch of code.

richardm-stripe (approved):

PR #2258, opened 2020-12-02
StripeEntityConverter to set RawJObject on all Stripe Entities

clement911:
The current `StripeEntity.RawJObject` is very hard to work with because it only works on the top level entity returned by the API call that was used.

`new CustomerService().Get(id).RawJObject["unsupported_field"]` => OK

`new CustomerService().List().First().RawJObject["unsupported_field"]` => NullReference exception

This PR aims to fix this issue by setting the RawJObject on all Stripe entities in the object graph returned.


remi-stripe:
@clement911 Thanks for the PR! The `RawJObject` is seen as an escape hatch on our end to access something secret/gated that is not officially supported by the library. While the change makes it a bit easier to access it per resource, it's not that much different from `listCustomers.RawJObject["data"][0]["unsupported_field"];`

We're discussing internally how to approach this but in the meantime could I ask you to ask some detailed tests for the changes? You are adding a brand new converter internal which can easily break during refactors so having a good test coverage would be helpful!

clement911:
Thank you @remi-stripe for the feedback.
I added some tests, let me know what you think.

I personally feel that this can make quite a big difference depending on how your code is structure.
Consider the following:

```
void ProcessOneEntity(MyEntity e)
{
   var unsupportedField = e.RawObject["unsupported_field"];
   //DO SOMETHING
}

void ProcessAllEntities(StripeList<MyEntity> list)
{
   foreach (var e in list)
      ProcessOneEntity(e);
}

void ProcessEntityUpdated(Event evt)
{
   ProcessOneEntity((MyEntity)e.Data.Object);
}
```

With the current serializer, that won't work because only the root object has a `RawObject`. 

With the new serializer, it works as expected. 

Now imagine that the code that makes the call to the API and the method `ProcessOneEntity` are "far" from each other.  
It would be very hard to cleanly pass the RawObject of the root API object and track the correct path to use in the `ProcessOneEntity` method. 

Is it :

- `RawJObject["unsupported_field"]`? (Get)
- `RawJObject["data"][0]["unsupported_field"]`? (List)
- `RawJObject["data"][5]["unsupported_field"]`? (List)
- `RawJObject["data"]["object]["unsupported_field"]`? (Event)

clement911:
Also note that I'm aware of the `event.EventData.RawObject` dynamic property but that is yet another way of accessing the unsupported field so it doesn't solve much. Actually, I think if you accept this PR, this property might be redundant.

remi-stripe:
I definitely agree that what you have makes this easier for sure! Someone on the team will look at it next week and circle back with feedback (if any)

cc @brandur-stripe to have a look too because we should try to be consistent with how we added `RawJSON` in stripe-go too.

clement911:
Hi again,
We're still bumping again this problem.
Can anyone look at this PR?

clement911:
Anyone?

pakrym-stripe:
@clement911, I'm sorry for the very long response. 

I think this fix is a good idea and keeps things working consistently.

The issue I have with it is more on the implementation/performance side. 

Calling `JObject.Load(reader);` for every object keeps re-materializing JObject trees again and again. Which for deep trees severely increases the memory usage.

Do you think this can be achieved my somehow re-using the root tree we already have parsed.

clement911:
@pakrym-stripe good point.
I've amended the logic so that the JSON is parsed and loaded only once for the top level entity. Nested entities will directly access the current token from the existing tree.

src/Stripe.net/Infrastructure/JsonConverters/StripeEntityConverter.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <returns>The object value.</returns>
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            if (reader.TokenType == JsonToken.Null)
+            {
+                return null;
+            }
+
+            JToken jToken;
+            if (reader is JTokenReader jTokenReader)
```
pakrym-stripe, line 62: Hm, I don't think we ever create a `JTokenReader` and the default implementation also doesn't use it: https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/JsonConvert.cs#L829

So we never hit the optimized codepath.

clement911: I guess I should have added some comments.

When calling `jToken.CreateReader()`, it actually returns an `JTokenReader` internally. `JTokenReader` is more efficient than `JsonReader` because instead of parsing the raw text, it just visits the existing json tree.

What happens is that the top level entity is parsed normally when calling `JToken.Load(reader)`. However, nested entities will detect that the reader is a `JTokenReader` and will create a new sub reader using the existing jToken.
As a result, the json is parsed once only. In addition, all of the `RawObject` properties of all the nested entities reference a node from the single top level tree so the json is loaded in memory once only.

I hope that makes sense.

If you want proof that this works, I would suggest that you merge this in (or check out my branch), run the test below and put a breakpoint in the `StripeEntityConverter`.

https://github.com/stripe/stripe-dotnet/blob/7b62c461d7c0cf2c9e06dce5e564b374a9d232e0/src/StripeTests/Entities/Accounts/AccountTest.cs#L14-L18

Also, you can see that the other converters actually do already create `JTokenReader` whenever they create a subreader (with the same pattern as here).
https://github.com/stripe/stripe-dotnet/blob/7b62c461d7c0cf2c9e06dce5e564b374a9d232e0/src/Stripe.net/Infrastructure/JsonConverters/AnyOfConverter.cs#L70
https://github.com/stripe/stripe-dotnet/blob/7b62c461d7c0cf2c9e06dce5e564b374a9d232e0/src/Stripe.net/Infrastructure/JsonConverters/StripeObjectConverter.cs#L63
https://github.com/stripe/stripe-dotnet/blob/7b62c461d7c0cf2c9e06dce5e564b374a9d232e0/src/Stripe.net/Infrastructure/JsonConverters/ExpandableFieldConverter.cs#L76
https://github.com/stripe/stripe-dotnet/blob/7b62c461d7c0cf2c9e06dce5e564b374a9d232e0/src/Stripe.net/Infrastructure/JsonConverters/EventConverter.cs#L72

pakrym-stripe: Ah, you are right. I'm sorry I missed that.

Do you think there is any way to get rid of the Reader allocation that we now have for every entity?

clement911:
Hi @pakrym-stripe , I think it's good to merge now?

src/Stripe.net/Entities/_base/StripeEntity.cs
```diff
@@ -27,29 +25,7 @@ public abstract class StripeEntity : IStripeEntity
...
-
-                return this.rawJObject;
-            }
-
-            protected set
-            {
-                this.rawJObject = value;
-            }
-        }
+        public JObject RawJObject { get; set; }
```
pakrym-stripe, line 28: I'm sorry I missed this in the original review. Can we keep the setter accessibility the same as before by defining an internal method that `StripeEntityConverter` will call to set the value?

pakrym-stripe (approved):
Looks great ! One small commend I missed in the original review.

clement911:
@pakrym-stripe all done.

clement911:
@pakrym-stripe are we all good to merge now?

PR #2276, opened 2021-01-21
Add missing transfer event types

liamwhan:
Working on integrating Stripe webhooks and I noticed the below events were missing from the constants defined in the Events class, not sure if this is intentional but if it isn't this PR rectifies this.

Events added:
- `transfer.paid`
- `transfer.failed`

### DIFF
--- a/src/Stripe.net/Constants/Events.cs
+++ b/src/Stripe.net/Constants/Events.cs
@@ -847,6 +847,16 @@ namespace Stripe
         public const string TransferUpdated = "transfer.updated";
 
         /// <summary>
+        /// Occurs after a transfer is paid. For Instant Payouts, the event will typically be sent within 30 minutes.
+        /// </summary>
+        public const string TransferPaid = "transfer.paid";
+
+        /// <summary>
+        /// Occurs whenever a transfer failed.
+        /// </summary>
+        public const string TransferFailed = "transfer.failed";
+
+        /// <summary>
         /// May be sent by Stripe at any time to see if a provided webhook URL is working.
         /// </summary>
         public const string Ping = "ping";


remi-stripe:
@liamwhan Thanks for opening a PR. Those events are not missing from stripe-dotnet at the moment though because they have been actively deprecated. 

Back in 2017 we released the API version [2017-04-06](https://stripe.com/docs/upgrades#2017-04-06) which split the `Transfer` resource in two separate concepts: The `Transfer` object stayed to represent money movement between Stripe accounts while the `Payout` object was introduced to represent money movements from Stripe to your bank account outside of Stripe.

Transfers between Stripe accounts are instant and succeed or fail synchronously. This means that either the Transfer succeeds immediately or you get an API error and it doesn't get created. For that reason, the `transfer.failed` and `transfer.paid` events were deprecated as they could not happen anymore.

Separately, Payouts could fail asynchronously, for example after we sent the funds to the bank and the bank returned the funds. For that reason, we introduced new events like `payout.created`, `payout.paid` and `payout.failed`.

This means if your account is on an API version that is on or after `2017-04-06`, you can never get the `transfer.paid` or `transfer.failed` events, as they don't make sense. Those events only exists for older API versions, for integrations built before that date.

The stripe-dotnet library is pinned to the most recent API version in the code. This means that the `transfer.paid` and `transfer.failed` events were removed a few years ago from the library as they can not happen with an integration on that API version.

Does that make sense?

liamwhan:
Hey @remi-stripe 

Gotcha, makes perfect sense cheers. which would also explain why the events haven't appeared in the webhook Event log, 

I'll abandon this PR, but it might be worth considering updating [this page](https://stripe.com/docs/api/events/types#event_types-transfer.failed) , anyone like myself who looks for these definitive lists to ensure they are covering all bases might benefit, 

Although it quite possible I've just missed it elsewhere in the documentation

remi-stripe:
Ah good point about the docs. Those shouldn't appear in the docs since they are only for legacy integrations. I'll get it fixed!

PR #2282, opened 2021-02-09
API Updates

richardm-stripe:
Codegen for openapi 04fba1e.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `afterpay_clearpay_payments` on `Account.capabilities`
* Add support for `payment_settings` on `Invoice`

### DIFF
--- a/src/Stripe.net/Entities/Invoices/Invoice.cs
+++ b/src/Stripe.net/Entities/Invoices/Invoice.cs
@@ -541,6 +541,9 @@ namespace Stripe
         internal ExpandableField<PaymentIntent> InternalPaymentIntent { get; set; }
         #endregion
 
+        [JsonProperty("payment_settings")]
+        public InvoicePaymentSettings PaymentSettings { get; set; }
+
         /// <summary>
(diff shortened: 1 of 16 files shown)


src/Stripe.net/Entities/Invoices/Invoice.cs
```diff
@@ -541,6 +541,9 @@ public PaymentIntent PaymentIntent
         internal ExpandableField<PaymentIntent> InternalPaymentIntent { get; set; }
         #endregion
 
+        [JsonProperty("payment_settings")]
```
remi-stripe, line 544: No docstring is that expected?

remi-stripe (approved):
Approved with a missing docstring though

src/Stripe.net/Entities/Invoices/Invoice.cs
richardm-stripe: No. I think this problem might extend to more expandable properties, too. Will open an internal ticket to investigate.

PR #2290, opened 2021-02-19
API Updates

richardm-stripe:
Codegen for openapi 403fdb8.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Added support for new resource `BillingPortal.Configuration`
* Added support for `configuration` and `on_behalf_of` on `Session#create` and `BillingPortal.Session`
* `WebhookEndpoint#create.enabled_events[]` and `WebhookEndpoint#update.enabled_events[]` added new enum members: `billing_portal.configuration.created and billing_portal.configuration.updated`

### DIFF
--- a/src/StripeTests/Entities/BillingPortal/Sessions/ConfigurationTest.cs
+++ b/src/StripeTests/Entities/BillingPortal/Sessions/ConfigurationTest.cs
@@ -1,0 +1,26 @@
+namespace StripeTests.BillingPortal
+{
+    using Newtonsoft.Json;
+    using Stripe;
+    using Stripe.BillingPortal;
+    using Xunit;
+
+    public class ConfigurationTest : BaseStripeTest
+    {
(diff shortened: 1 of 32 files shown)


src/StripeTests/Entities/BillingPortal/Sessions/ConfigurationTest.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        public ConfigurationTest(StripeMockFixture stripeMockFixture)
+            : base(stripeMockFixture)
+        {
+        }
+
+        [Fact]
+        public void Deserialize()
+        {
+            var json = GetResourceAsString("api_fixtures.billing_portal.configuration.json");
```
remi-stripe, line 18: Same as java, stripe-mock can return the fixture, so we should not be using a fixture here

remi-stripe (approved):
Approving but I'd prefer if we fixed the test instead to rely on stripe-mock and not a local fixture

PR #2329, opened 2021-06-15
Create PaymentIntentPaymentMethodCard.cs

GigiAkamala:
A token may not be passed in as a PaymentMethod. Instead, use payment_method_data with type=card and card[token]=tok_xxxxxxxxxx

### DIFF
--- a/src/Stripe.net/Services/PaymentIntents/PaymentIntentPaymentMethodCard.cs
+++ b/src/Stripe.net/Services/PaymentIntents/PaymentIntentPaymentMethodCard.cs
@@ -1,0 +1,10 @@
+namespace Stripe
+{
+    using Newtonsoft.Json;
+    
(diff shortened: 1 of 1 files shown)


remi-stripe:
@GigiAkamala The change you have wouldn't really work as is since you're not using that class anywhere. You'l also trying to add support for a parameter but the naming of the class implies it's part of the resource.

We don't want to add support for the `payment_method_data[card]` parameter in stripe-dotnet as it could encourage developers to pass raw PANs. We could support just `payment_method_data[card][token]` but it'd be more confusing to new developers as passing a Token is mostly an escape hatch for old integrations.

A better solution here is to use `AddExtraParam` to pass the parameter even though it's not officially in the library.

```
options.AddExtraParam("payment_method_data[card][token]", "tok_12345");
```

I'm going to close this out, I hope it's okay!

PR #2333, opened 2021-06-21
Add infrastructure for handling streaming binary responses

ramon-stripe:
In support of upcoming API methods that return binary data, this PR adds versions of the request methods that return a `Stream` instead of assuming a JSON response.

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/StripeStreamedResponse.cs
+++ b/src/Stripe.net/Infrastructure/Public/StripeStreamedResponse.cs
@@ -1,0 +1,51 @@
+namespace Stripe
+{
+    using System.IO;
+    using System.Net;
+    using System.Net.Http.Headers;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// A streaming binary response. Body is not assumed to be text for successful responses.
+    /// </summary>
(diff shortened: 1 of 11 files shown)


src/Stripe.net/Infrastructure/Public/StripeStreamedResponse.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// Converts this response into a regular <see cref="StripeResponse"/> by
+        /// reading the body in full.
+        /// </summary>
+        /// This assumes that the response body is textual, which will be the case for
+        /// most API responses and for errors.
+        /// <returns>The response with body fully read.</returns>
+        public async Task<StripeResponse> ToStripeResponseAsync()
+        {
+            // We keep a reference to the task because we can only read the body once.
+            if (this.fetchFullyTask == null)
```
richardm-stripe, line 36: Hmm, I didn't think of this in [stripe-java](https://github.com/stripe/stripe-java/pull/1207/files#diff-f748ba076a0736ad824fca475e0594a7d8d118fd6d241ad0a0e047db81e43e49R25). Do you think it's important permit multiple calls there, too?

richardm-stripe (approved):
This looks good to me! Just had one question about "fetchFully"

Issue #2341, opened 2021-06-30
39.56.0 causes Request Timed Out errors


TomGullen:
Was using 39.55.0, updated to 39.56.0 and there appears to be issues with creating/updating PaymentIntents (and other actions I beleive?) where it just freezes with `Request Timed Out` errors.  Rolling back to 39.55.0 seems to fix the issue.  I know this isn't the best bug report but will update with more information when I run some tests.

It's hard to pinpoint the exact problem as it freezes our application in a waiting state and it's hard to catch or pinpoint any exceptions.

remi-stripe:
@TomGullen tried to reproduce on my end on the latest version but PaymentIntent creation works fine. Getting more details about your exact code, example request ids (req_123) and details about your set up such as exact .NET framework and operating system you're using?

dcr-stripe:
Hi @TomGullen, thanks for the report! We've rolled out a new patch version, https://github.com/stripe/stripe-dotnet/releases/tag/v39.56.1, to try and address this. Could you try with this and see if the issue is resolved for you?

We believe it has to do with a `SynchronizationContext` configuration which differs from our testing environment and could have encountered a deadlock due to missing `ConfigureAwait` calls.

TomGullen:
Thank you, updated to 39.57.0 and it appears to of fixed the issue!

ramon-stripe:
@tomer-stripe Thank you again for reporting this so quickly, glad to hear the issue is resolved for you. I am trying to add unit tests to reproduce this issue. Would it be possible to give any more details on your setup e.g. ASP.NET version or more specifically do you know what `SynchronizationContext` you are using?

TomGullen:
No problem!  Always keep everything as up to date as possible and check for updates on our nuget packages almost daily so do spot problems quickly usually.

Using ASP.net 4.8 with Webforms.  It's hard to locate specifically where the error was being thrown due to the nature of the exception (it crashes my dev environment in a non helpful way) but most of our calls take this form:

```
private static PaymentIntent CreatePaymentIntent(
        StripeClient stripeClient,
        PaymentIntentService paymentIntentService,
        PaymentMethod method,
        ModelsNew.AssetStore.Basket forBasket,
        Models.Currency inCurrency,
        long totalCents,
        long voucherBalanceToUseCents)
    {
        // Basket item types
        var basketItemTypes = forBasket.GetBasketItemTypes();

        // Get relevant Stripe customer ID
        string stripeCustomerID = null;
[...]

PR #2342, opened 2021-06-30
Add missing `ConfigureAwait` call to `SystemNetHttpClient.SendHttpRequest`

ramon-stripe:


ramon-stripe:
Suspect this may be behind #2341 
Asynchronously will investigate if there is some way to detect this in tests in the future.

dcr-stripe (approved):
Thanks Ramon, looks good!

Issue #2353, opened 2021-07-27
Implement UpcomingInvoiceOptions.CustomerDetails


jon-armen:
https://stripe.com/docs/api/invoices/upcoming?lang=dotnet

Indicates that customer details can be specified when customer id is omitted.

PR #2355, opened 2021-07-27
Adding customer details to Upcoming Invoice Options

jon-armen:
See #2353


remi-stripe (approved):
Thank you for opening a PR. This file is not yet automatically generated from our openapi spec which is why we missed it.

PR #2359, opened 2021-08-19
API Updates

dcr-stripe:
Codegen for openapi 10ebdb9.
r? @richardm-stripe
cc @stripe/api-libraries


remi-stripe (approved):
Approved but it's purely a docs changes

PR #2366, opened 2021-08-31
API Updates

richardm-stripe:
Codegen for openapi 944e56f.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `AfterExpiration`, `ConsentCollection`, and `ExpiresAt` on `CheckoutSessionCreateOptions` and `CheckoutSession`
* Add support for `Consent` and `RecoveredFrom` on `CheckoutSession`

### DIFF
--- a/src/Stripe.net/Entities/Checkout/Sessions/SessionAfterExpirationRecovery.cs
+++ b/src/Stripe.net/Entities/Checkout/Sessions/SessionAfterExpirationRecovery.cs
@@ -1,0 +1,39 @@
+// File generated from our OpenAPI spec
+namespace Stripe.Checkout
+{
+    using System;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
+    public class SessionAfterExpirationRecovery : StripeEntity<SessionAfterExpirationRecovery>
+    {
+        /// <summary>
+        /// Enables user redeemable promotion codes on the recovered Checkout Sessions. Defaults to
+        /// <c>false</c>.
+        /// </summary>
+        [JsonProperty("allow_promotion_codes")]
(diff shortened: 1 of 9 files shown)


src/Stripe.net/Entities/Checkout/Sessions/SessionAfterExpirationRecovery.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// </summary>
+        [JsonProperty("enabled")]
+        public bool? Enabled { get; set; }
+
+        /// <summary>
+        /// The timestamp at which the recovery URL will expire.
+        /// </summary>
+        [JsonProperty("expires_at")]
+        [JsonConverter(typeof(UnixDateTimeConverter))]
+        public DateTime? ExpiresAt { get; set; }
```
remi-stripe, line 30: Why is this nullable? Shouldn't it be either always set or the entire hash is null?

remi-stripe (approved):

src/Stripe.net/Entities/Checkout/Sessions/SessionAfterExpirationRecovery.cs
richardm-stripe: Good catch - from the checkout team:
> `after_expiration.recovery.enabled` should be required - `after_expiration.recovery.allow_promotion_codes` is optional on the write side with a default of false - so the resource [should] have that field as required as well. `after_expiration.recovery.url` and `after_expiration.recovery.expires_at` will be null until the Checkout Session expires, so those are correct

richardm-stripe:
re-r? @remi-stripe with the latest changes

remi-stripe (approved):

PR #2368, opened 2021-09-07
API Updates

richardm-stripe:
Codegen for openapi ad7d382.
r? @dcr-stripe
cc @stripe/api-libraries


richardm-stripe:
docstring-only, closing

PR #2375, opened 2021-10-07
API Updates

richardm-stripe:
Codegen for openapi d0904af.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `PhoneNumberCollection` on `CheckoutSessionCreateOptions` and `CheckoutSession`
* Add support for `Phone` on `CheckoutSessionCustomerDetails`


richardm-stripe:
@dcr-stripe I don't think so, because it's just a change to the docstring and doesn't materially affect the library. In `stripe-php` the PHPDoc is used for typechecking/static analysis so it's a little different.

PR #2376, opened 2021-10-08
API Updates

richardm-stripe:
Codegen for openapi c2caca4.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `ListPaymentMethods` method on resource `Customer`


src/Stripe.net/Services/Customers/CustomerListPaymentMethodsOptions.cs
```diff
@@ -0,0 +1,11 @@
+// File generated from our OpenAPI spec
+namespace Stripe
+{
+    using Newtonsoft.Json;
+
+    public class CustomerListPaymentMethodsOptions : ListOptions
+    {
+        [JsonProperty("type")]
```
dcr-stripe, line 8: Should there be documentation here of the various enum values?

PR #2415, opened 2022-01-19
API Updates

richardm-stripe:
Codegen for openapi 4d2f301.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new resource `PaymentLink`
* Add support for `PaymentLink` on `CheckoutSession`

### DIFF
--- a/src/StripeTests/Services/GeneratedExamplesTest.cs
+++ b/src/StripeTests/Services/GeneratedExamplesTest.cs
@@ -1358,6 +1358,20 @@ namespace StripeTests
         }
 
         [Fact]
+        public void TestPaymentLinkServiceListLineItems()
+        {
+            var service = new PaymentLinkService(this.StripeClient);
+            service.ListLineItems("pl_xyz");
+        }
+
+        [Fact]
+        public void TestPaymentLinkServiceRetrieve()
+        {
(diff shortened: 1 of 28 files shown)


src/StripeTests/Services/GeneratedExamplesTest.cs
```diff
@@ -1339,6 +1339,38 @@ public void TestPaymentIntentServiceUpdate()
...
+            };
+            var service = new PaymentLinkService(this.StripeClient);
+            service.Create(options);
+        }
+
+        [Fact]
+        public void TestPaymentLinkServiceListLineItems()
+        {
+            var service = new PaymentLinkService(this.StripeClient);
+            service.ListLineItems("pl_xyz");
```
remi-stripe, line 1364: Those tests are quite lacking. Old tests checked the return class and its object property and all that no?

remi-stripe (approved):
nitpick on tests generated but looks great

src/StripeTests/Services/GeneratedExamplesTest.cs
richardm-stripe: I agree. Checking the return class seems a little redundant since that is already enforced by the static type, but I'll look into seeing if we can add object property checks to these.

PR #2419, opened 2022-02-03
API Updates

yejia-stripe:
Codegen for openapi 10fe9f5.
r? @dcr-stripe
cc @stripe/api-libraries


dcr-stripe (commented):
I think we can skip this one as well since it's technically all comments and has no functionality change in .NET.

PR #2427, opened 2022-02-25
Add Source Link support for enhanced debugging experience

jon-armen:
Currently, a symbol package is generated, but never pushed onto a nuget feed, thus Source Link support is not functioning.  This update adds the Reproducible Builds package (https://github.com/dotnet/reproducible-builds), which includes Source Link support and sets properties required to embed the symbols into the executable, removing the need for a symbol package.


Directory.Build.props
```diff
@@ -0,0 +1,5 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <PackageReference Include="DotNet.ReproducibleBuilds" Version="1.1.1" PrivateAssets="All"/>
```
pakrym-stripe, line 3: We already set most of the properties that DotNet.ReproducibleBuilds does. I think the only one that's missing is `    <DebugType>embedded</DebugType>` can we add just it?

pakrym-stripe:
Thank you for your contribution! I think we can enable a similar set of functionality without taking an extra package dependency.

pakrym-stripe:
Btw, we now also publish the symbol package: https://github.com/stripe/stripe-dotnet/runs/6120550646?check_suite_focus=true

pakrym-stripe:
Hi, we've enabled embedded PDBs in https://github.com/stripe/stripe-dotnet/pull/2484.

PR #2437, opened 2022-03-15
Add infrastructure for search methods

pakrym-stripe:
https://jira.corp.stripe.com/browse/DEVSDK-526

r? @dcr-stripe 

Adds infrastructure required to generate Search methods.

Adds mock tests we can add stripe-mock coverage when it's updated.


src/Stripe.net/Entities/StripeSearchResult.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// The URL for accessing this list.
+        /// </summary>
+        [JsonProperty("url")]
+        public string Url { get; set; }
+
+        /// <summary>
+        /// The URL for accessing the next page in search results.
+        /// </summary>
+        [JsonProperty("page")]
+        public string Page { get; set; }
```
dcr-stripe, line 40: Results will continue to return `next_page`. Only the request uses `page`.

src/Stripe.net/Services/_base/SearchOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+    public class SearchOptions : BaseOptions
+    {
+        /// <summary>
+        /// A limit on the number of objects to be returned, between 1 and 100.
+        /// </summary>
+        [JsonProperty("limit")]
+        public long? Limit { get; set; }
+
+        [JsonProperty("next_page")]
+        public string NextPage { get; set; }
```
dcr-stripe, line 14: This should be `page`.

pakrym-stripe:
>Have you done an end to end test with this against an actual end point? 

I did. I'm looking into why I didn't catch this.

pakrym-stripe:
re-r? @dcr-stripe

Issue #2448, opened 2022-04-02
Encapsulate Signature Validation


bobbyangers:
`EventUtility.ConstructEvent` rely on a few `static` methods to validate the signature, making the unit testing a little cumbersome.

Using the following technique, we could greatly enhance the testability by encapsulating the validation logic in an `ISignatureValidator` and use `DefaultSignatureValidator`

And exposing a:
```
 public static Event ConstructEvent(string json, string stripeSignatureHeader,
            ISignatureValidator validator,
            long tolerance,
            long utcNow,
            bool throwOnApiVersionMismatch = true)
```

Here is my proposed changes, but I could not get the unit tests to work....
```
public interface ISignatureValidator {
    void Validate(string json, string stripeSignatureHeader, long tolerance, long utcNow);
}

public class DefaultSignatureValidator : ISignatureValidator
{
    public DefaultSignatureValidator(string secret) { /* ... */ }
    /* code redacted */

  private static ILookup<string, string> ParseStripeSignature(string stripeSignatureHeader) { }
  private static bool IsSignaturePresent(string signature, IEnumerable<string> signatures) { }
  private static string ComputeSignature(string secret, string timestamp, string payload) { }
}

//////////////////////////////////////////////////////
public class EventUtility
{
          /*  other code redacted */
           public static Event ConstructEvent(string json, string stripeSignatureHeader,
            string secret,
            long tolerance,
            long utcNow,
            bool throwOnApiVersionMismatch = true)
        {
			var validator = new DefaultSignatureValidator(secret)
            return ConstructEvent(json, stripeSignatureHeader, validator, tolerance, utcNow,  throwOnApiVersionMismatch);
        }

           public static Event ConstructEvent(string json, string stripeSignatureHeader,
            ISignatureValidator validator,
            long tolerance,
            long utcNow,
            bool throwOnApiVersionMismatch = true)
        {
            validator.Validate(json,  stripeSignatureHeader, tolerance, utcNow);
            return ParseEvent(json, throwOnApiVersionMismatch);
        }
}
```

I forked the repo and tried it, and created a draft PR;

pakrym-stripe:
Hi, @bobbyangers. I'm going to close the issue and PR for now, but feel free to reopen them if you still think there is something to be fixed.

PR #2449, opened 2022-04-02
Encapsulate the signature validator logic

bobbyangers:
#2448


pakrym-stripe:
Closing, let's chat in the issue about the problem you are trying to solve and reopen when we come to the right design.

PR #2452, opened 2022-04-11
Add support for updating the version of the repo

pakrym-stripe:
Encodes the logic for version updating as a Makefile. Simplifies what the generation tools needs to know about the repo.
The plan is to have the same interface across all the repos.

Sample usage:

```
make update-version VERSION=1.2.5
```

Also adds a target that codegen will use to format all our repos:

```
make codegen-format
```


richardm-stripe:
LGTM, weak preference for `sed` over `perl` for regex replace because there's [slightly](https://github.com/stripe/stripe-java/blob/432ee5b1/.github/workflows/ci.yml#L75) [more](https://github.com/stripe/stripe-cli/blob/8ee30a5b/scripts/publish-deb-to-bintray.sh#L28) [precedent](https://github.com/stripe/stripe-firebase-extensions/blob/533adbc6/firestore-stripe-web-sdk/deploy.sh#L28) in other Stripe OSS repos.

richardm-stripe (approved):

dcr-stripe (approved):
The goal is to add this interface to each SDK right? Just wanted to double check that you've confirmed we can do the version update similarly for each language.

Assuming that, LGTM!

PR #2454, opened 2022-04-14
Push NuGet package from Github Actions

pakrym-stripe:
Adds a manual-dispatch workflow step to release nuget packages to Nuget.org.

The testing plan:
1. Try to push an old tag (to see a version already exists message from the package manager).
2. Disable the appveyor and release a new package manually
3. Flip workflow to automatic on tag pushes

r? @dcr-stripe

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,13 +4,15 @@ on:
   push:
     branches:
       - master
+    tags:
+      - '*'
   pull_request:
     branches:
       - master
 
 jobs:
-  tests:
-    name: Tests
+  build:
+    name: Build and test
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -57,3 +59,27 @@ jobs:
         env:
           CI_NAME: github-actions
           COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+      - name: Pack
+        run: dotnet pack src -c Release --no-build --output nuget
+      - name: 'Upload Artifact'
+        uses: actions/upload-artifact@v2
```
dcr-stripe, line 65: WDYT about only conditionally uploading the artifact if we know we're publishing to avoid unnecessary storage?

.github/workflows/ci.yml
```diff
@@ -57,3 +59,27 @@ jobs:
...
+        with:
+          name: nuget
+          path: nuget/
+
+  publish:
+    if: (github.event_name == 'workflow_dispatch') && (github.ref == 'refs/heads/master') && (endsWith(github.actor, '-stripe')
```
dcr-stripe, line 71: Do we want to set up a CODEOWNERS file in .github as well to limit modification to publishing?

.github/workflows/ci.yml
```diff
@@ -57,3 +59,27 @@ jobs:
...
+      uses: actions/download-artifact@v2
+      with:
+        name: nuget
+        path: nuget
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 6.0.x
+    - name: Publish NuGet packages to NuGet
+      run: dotnet nuget push nuget/*.nupkg --api-key ${{ secrets.NUGET_KEY }} --source "nuget.org"
```
dcr-stripe, line 85: Do we need to filter explicitly for `src\Stripe.net\bin\Release\*.nupkg` as we were doing in appveyor.yml? Or is that no longer necessary here?

dcr-stripe (commented):
Small questions here and then I think we're good to go! 🎉

.github/workflows/ci.yml
pakrym-stripe: Because files are coming from the artifact they land in the nuget directory we don't need extra filtering.

.github/workflows/ci.yml
pakrym-stripe: Artifacts are cleaned after 90 days by github. I've never seen artifact storage becoming a problem.

dcr-stripe (approved):
LGTM!

PR #2462, opened 2022-04-21
Enable workflow dispatch and codeowners

pakrym-stripe:


.github/workflows/ci.yml
```diff
@@ -1,6 +1,7 @@
 name: CI
 
 on:
+  workflow_dispatch: {}
```
dcr-stripe, line 4: Whoops - I missed this too :)

dcr-stripe (commented):
Can we use the admins group for .github CODEOWNERS as well?

pakrym-stripe:
> Can we use the admins group for .github CODEOWNERS as well?

I don't think we need the extra line for .github. `*` includes all the files.

dcr-stripe (approved):
LGTM!

PR #2465, opened 2022-04-22
Enable automated Github release

pakrym-stripe:
Turns the github workflow from manual to automatic on tag push.


src/Stripe.net/Stripe.net.csproj
```diff
@@ -4,8 +4,8 @@
   <PropertyGroup>
     <Description>Stripe.net is a sync/async .NET 4.6.1+ client, and a portable class library for the Stripe API.  (Official Library)</Description>
     <AssemblyTitle>Stripe.net</AssemblyTitle>
-    <VersionPrefix>39.107.0</VersionPrefix>
     <Version>39.107.0</Version>
+    <VersionPrefix>$(Version)</VersionPrefix>
```
pakrym-stripe, line 8: Fewer places to update.

dcr-stripe, line 8: Nice.

.github/workflows/ci.yml
```diff
@@ -1,12 +1,11 @@
 name: CI
 
 on:
-  workflow_dispatch: {}
```
dcr-stripe, line 4: Do we want to keep this in case we need to manually publish?

.github/workflows/ci.yml
```diff
@@ -1,12 +1,11 @@
...
   push:
     branches:
       - master
     tags:
-      - '*'
+      - v*
```
dcr-stripe, line 8: This can't be a more strict regex right?

dcr-stripe (commented):
🎉 awesome! Just some small comments and then good to 🚢 !

.github/workflows/ci.yml
pakrym-stripe: We can but I'm not sure if it will add anything.

.github/workflows/ci.yml
pakrym-stripe: Good idea.

.github/workflows/ci.yml
dcr-stripe: I think it'd be good just to make sure this doesn't accidentally get released for some other tag. Better to be overly cautious here.

.github/workflows/ci.yml
```diff
@@ -69,7 +69,8 @@ jobs:
           path: nuget/
 
   publish:
-    if: ((github.event_name == 'workflow_dispatch') && (github.ref == 'refs/heads/master') && endsWith(github.actor, '-stripe'))
+    name: Publish
+    if: (((github.event_name == 'push') || (github.event_name == 'workflow_dispatch')) && startsWith(github.ref, 'refs/tags/v') && endsWith(github.actor, '-stripe'))
```
dcr-stripe, line 73: Double checking here that we're OK with having workflow_dispatch only work on ref/tags/ rather than master

dcr-stripe (approved):
LGTM!

.github/workflows/ci.yml
pakrym-stripe: Yep, now that the workflow is going to be present in tag commit we should be good.

Issue #2424, opened 2022-02-20
Passing a header with Stripe-Signature not in the right format gives exception


aamirjamal:
In the version [39.91.0](https://github.com/stripe/stripe-dotnet/commit/624f3450cb5855ff2f6cd1db9310279d9a3c1489),
when we try to authenticate a request using the recommended way of using the library method EventUtility.ConstructEvent(), we get an Exception if we pass in a header with Stripe-Signature as key and a value which is in invalid format (eg "test").

This happens as the code tries to split the signature first on the basis of ',' and then on '=' and then pick the split items on the basis of index 0 and 1, thus giving an Array out of bound exception.

Recommended fix : 
We can easily fix this by putting the parsing logic (line linked shown below) into a try catch block.

https://github.com/stripe/stripe-dotnet/blob/7b62c461d7c0cf2c9e06dce5e564b374a9d232e0/src/Stripe.net/Services/Events/EventUtility.cs#L138

PR #2468, opened 2022-04-25
Harden ValidateSignature against invalid input

pakrym-stripe:
Fixes: https://github.com/stripe/stripe-dotnet/issues/2424

Adds checks and an error message before accessing array elements by index.


src/Stripe.net/Services/Events/EventUtility.cs
```diff
@@ -166,10 +166,22 @@ public static void ValidateSignature(string json, string stripeSignatureHeader,
 
         private static ILookup<string, string> ParseStripeSignature(string stripeSignatureHeader)
         {
+            (string Key, string Value) ParseItem(string item)
+            {
+                string[] parts = item.Trim().Split(new[] { '=' }, 2);
```
kamil-stripe, line 171: Should that '2' be removed? How does it work with the first test case? Wouldn't that case return two parts?

kamil-stripe (approved):

src/Stripe.net/Services/Events/EventUtility.cs
kamil-stripe: I see. Thanks!

pakrym-stripe: The first test case gets parsed into 2 elements with the second one being empty. ` ParseStripeSignature` succeeds but validation fails further down the line.

PR #2473, opened 2022-05-04
API Updates

dcr-stripe:
Codegen for openapi 158cf6c.
r? @kamil-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `RegisteredAddress` on `AccountIndividualOptions`, `PersonCreateOptions`, `PersonUpdateOptions`, `Person`, `TokenAccountIndividualOptions`, and `TokenPersonOptions`
* Change type of `PaymentIntentAmountDetailsTipAmount` from `nullable(integer)` to `integer`
* Add support for `PaymentMethodData` on `SetupIntentConfirmOptions`, `SetupIntentCreateOptions`, and `SetupIntentUpdateOptions`

### DIFF
--- a/src/Stripe.net/Entities/PaymentIntents/PaymentIntentAmountDetailsTip.cs
+++ b/src/Stripe.net/Entities/PaymentIntents/PaymentIntentAmountDetailsTip.cs
@@ -9,6 +9,6 @@ namespace Stripe
         /// Portion of the amount that corresponds to a tip.
         /// </summary>
         [JsonProperty("amount")]
-        public long Amount { get; set; }
+        public long? Amount { get; set; }
     }
 }


src/Stripe.net/Entities/PaymentIntents/PaymentIntentAmountDetailsTip.cs
```diff
@@ -9,6 +9,6 @@ public class PaymentIntentAmountDetailsTip : StripeEntity<PaymentIntentAmountDet
         /// Portion of the amount that corresponds to a tip.
         /// </summary>
         [JsonProperty("amount")]
-        public long? Amount { get; set; }
+        public long Amount { get; set; }
```
kamil-stripe, line 12: Have we discussed this?

kamil-stripe (commented):
It's possible there is more changes than API changes because of our changes to Options file generation. Let me double check.

src/Stripe.net/Entities/PaymentIntents/PaymentIntentAmountDetailsTip.cs
kamil-stripe: > It's possible there is more changes than API changes because of our changes to Options file generation. Let me double check.

Scratch that, only changes in src/Stripe.net/Services/Sources/SourceCardOptions.cs are related to that. It looks good outside of that Amount field.

dcr-stripe: Good catch. Fixed!

kamil-stripe (approved):

PR #2478, opened 2022-05-10
Make dotnet formatting faster

pakrym-stripe:
We are mostly interested in whitespace formatting. This avoids a full workspace load.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,8 +31,7 @@ jobs:
 
       - name: Install tools
         run: |
-          dotnet tool install dotnet-format --version 4.1.131201 --tool-path tools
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -31,8 +31,7 @@ jobs:
 
       - name: Install tools
         run: |
-          dotnet tool install dotnet-format --version 4.1.131201 --tool-path tools
```
dcr-stripe, line 34: Is the dotnet-format tool included by default so we don't need to have this line anymore?

pakrym-stripe: Yes. But we are also not using it in CI.

dcr-stripe (approved):
LGTM

PR #2479, opened 2022-05-10
API Updates

richardm-stripe:
Codegen for openapi 0f6948c.
r? @richardm-stripe
cc @stripe/api-libraries

## Summary
* API changes
* Removed tests for defunct orders endpoints that are changed in the OpenAPI spec.
## Changelog

* Add support for `AmountDiscount`, `AmountTax`, and `Product` on `LineItem`


src/StripeTests/Entities/Orders/OrderTest.cs
```diff
@@ -1,48 +0,0 @@
-namespace StripeTests
```
pakrym-stripe, line 1: Aren't these still funcitonal?

pakrym-stripe: Ah, I guess the openapi doesn't have endpoints anymore.

pakrym-stripe (approved):

PR #2484, opened 2022-05-17
Enable embedded symbols

pakrym-stripe:
Makes the debugging simpler. (https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/nuget#symbol-packages)

Similar to https://github.com/stripe/stripe-dotnet/pull/2427 but doesn't require an additional package reference.

### DIFF
--- a/src/Stripe.net/Stripe.net.csproj
+++ b/src/Stripe.net/Stripe.net.csproj
@@ -17,6 +17,7 @@
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <IncludeSymbols>true</IncludeSymbols>
+    <DebugType>embedded</DebugType>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


src/StripeTests/StripeTests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net5.0;net6.0;netcoreapp3.1;netcoreapp2.1;net461</TargetFrameworks>
```
dcr-stripe, line 4: We're dropping netcoreapp2.1 - here and below - is that expected?

src/Stripe.net/Stripe.net.csproj
```diff
@@ -17,6 +17,7 @@
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <IncludeSymbols>true</IncludeSymbols>
+    <DebugType>embedded</DebugType>
```
dcr-stripe, line 20: Does this get included in our release artifact/grow its size?

dcr-stripe (commented):
PTAL @pakrym-stripe

src/StripeTests/StripeTests.csproj
pakrym-stripe: Added it back here. Below is fine. We don't need a separate package list for every target framework.

src/Stripe.net/Stripe.net.csproj
pakrym-stripe: Yes, adds 100kb to a 3.9mb package.

PR #2487, opened 2022-05-18
Set CheckEolTargetFramework to false

pakrym-stripe:
We know that 2.1 is EOD but don't want to remove it yet. Suppress the warning.


dcr-stripe (approved):
LGTM!

PR #2490, opened 2022-05-20
Do not create a symbol package now that symbols are embedded

pakrym-stripe:
The symbol package is empty because symbols are embedded into the main package. 

Fixes the publishing error.


dcr-stripe (approved):
LGTM

PR #2501, opened 2022-06-06
Stop pointing to coveralls 

pakrym-stripe:
It's been outdated for a while.


dcr-stripe (approved):
LGTM!

Should we remove https://github.com/stripe/stripe-dotnet/blob/95e0f2eb/.github/workflows/ci.yml#L51-L61 and https://github.com/stripe/stripe-dotnet/blob/95e0f2eb/.config/dotnet-tools.json#L5?

Alternatively - do we want coveralls?

pakrym-stripe:
>Alternatively - do we want coveralls/coverage info?

I'll remove the code for now. If we want coverage consistently across repos we can add it back.

PR #2510, opened 2022-06-18
Update InvoiceItemPriceDataOptions.cs

todd-viziwyg:
Added TaxBehaviour as per: https://stripe.com/docs/api/invoiceitems/create


pakrym-stripe:
Thank you for your contribution! This file was not autogenerated before but we fixed it in https://github.com/stripe/stripe-dotnet/pull/2520 and the member you are looking for should be present now.

PR #2519, opened 2022-06-23
Add TreasuryReceivedCreditReturned event

kamil-stripe:
r? @dcr-stripe 

## Changelog
* Mark `treasury.check_deposit.reversed` and `treasury.received_credit.reversed` events as obsolete.
* Add `treasury.received_credit.returned` event.


dcr-stripe (commented):
We should also mark the following as obsolete right?

```
        public const string TreasuryCheckDepositReversed = "treasury.check_deposit.reversed";
```

Also let's add a proper PR description (similar to the changelogs from API updates) so that this can get pulled into the changelog for the next release.

kamil-stripe:
Not needed anymore.

PR #2520, opened 2022-06-24
Autogenerate more files

pakrym-stripe:
Switch line item-related files to be auto-generated.

r? @dcr-stripe


src/Stripe.net/Services/InvoiceItems/InvoiceItemPriceDataOptions.cs
```diff
@@ -0,0 +1,46 @@
...
+        public string Product { get; set; }
+
+        /// <summary>
+        /// Specifies whether the price is considered inclusive of taxes or exclusive of taxes. One
+        /// of <c>inclusive</c>, <c>exclusive</c>, or <c>unspecified</c>. Once specified as either
+        /// <c>inclusive</c> or <c>exclusive</c>, it cannot be changed.
+        /// One of: <c>exclusive</c>, <c>inclusive</c>, or <c>unspecified</c>.
+        /// </summary>
+        [JsonProperty("tax_behavior")]
+        public string TaxBehavior { get; set; }
```
dcr-stripe, line 29: Double checking: this was excluded due to being manually maintained and no other reason, right?

src/Stripe.net/Services/Plans/PlanProductOptions.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// statement in capital letters. Non-ASCII characters are automatically stripped.
+        /// </summary>
+        [JsonProperty("statement_descriptor")]
+        public string StatementDescriptor { get; set; }
+
+        /// <summary>
+        /// A <a href="https://stripe.com/docs/tax/tax-categories">tax code</a> ID.
+        /// </summary>
+        [JsonProperty("tax_code")]
+        public string TaxCode { get; set; }
```
dcr-stripe, line 53: Same check here.

src/Stripe.net/Services/Subscriptions/SubscriptionItemPriceDataOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+        public SubscriptionItemPriceDataRecurringOptions Recurring { get; set; }
+
+        /// <summary>
+        /// Specifies whether the price is considered inclusive of taxes or exclusive of taxes. One
+        /// of <c>inclusive</c>, <c>exclusive</c>, or <c>unspecified</c>. Once specified as either
+        /// <c>inclusive</c> or <c>exclusive</c>, it cannot be changed.
+        /// One of: <c>exclusive</c>, <c>inclusive</c>, or <c>unspecified</c>.
+        /// </summary>
+        [JsonProperty("tax_behavior")]
+        public string TaxBehavior { get; set; }
```
dcr-stripe, line 35: Same check here.

dcr-stripe (approved):
LGTM, thank you for doing this @pakrym-stripe ! Just a few things to double check and then good to merge.

I'm also assuming here that the file move itself isn't breaking since it's still all under the Stripe namespace.

src/Stripe.net/Services/InvoiceItems/InvoiceItemPriceDataOptions.cs
pakrym-stripe: I suspect so. We didn't suppress this property in any other generated file.

src/Stripe.net/Services/Plans/PlanProductOptions.cs
pakrym-stripe: Yep, we have `tax_code` in many other places.

src/Stripe.net/Services/Subscriptions/SubscriptionItemPriceDataOptions.cs
pakrym-stripe: Same as in `InvoiceItemPriceDataOptions.cs`

PR #2551, opened 2022-08-05
API Updates

pakrym-stripe:
Codegen for openapi v176.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `ExpiresAt` on `AppsSecretCreateOptions` and `AppsSecret`


src/Stripe.net/Services/Apps/Secrets/SecretCreateOptions.cs
```diff
@@ -1,10 +1,19 @@
 // File generated from our OpenAPI spec
 namespace Stripe.Apps
 {
+    using System;
```
yejia-stripe, line 4: ooc why did these get added?

yejia-stripe (approved):

src/Stripe.net/Services/Apps/Secrets/SecretCreateOptions.cs
pakrym-stripe: `DateTime` is in the `System` namespace and `UnixDateTimeConverter` is in `Stripe.Infrastructure`.

PR #2555, opened 2022-08-09
Set MaxDepth for serialization and deserialization

pakrym-stripe:
Limit resources that JSON parsing/writing can consume for untrusted inputs.

r? @dcr-stripe


dcr-stripe (approved):
LGTM, 128 feels safe to me!

PR #2561, opened 2022-08-19
Make APIVersion configuration settable

pakrym-stripe:
Allows StripeConfiguration.ApiVersion value to be modified. Trims beta headers before using them in WebHook event version check.

r? @dcr-stripe

### DIFF
--- a/src/Stripe.net/Infrastructure/StringUtils.cs
+++ b/src/Stripe.net/Infrastructure/StringUtils.cs
@@ -73,10 +73,10 @@ namespace Stripe.Infrastructure
 
         /// <summary>
         /// Trims the beta header part of an API Version string.
-        /// For example, "22-12-2022; orders_beta=v3" is converted to "22-12-2022".
+        /// For example, "12-22-2022; orders_beta=v3" is converted to "12-22-2022".
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/StripeTests/Infrastructure/TestsThatModifyApiVersion.cs
```diff
@@ -0,0 +1,9 @@
+namespace StripeTests
+{
+    using Xunit;
+
+    [CollectionDefinition(nameof(TestsThatModifyApiVersion), DisableParallelization = true)]
```
dcr-stripe, line 5: Neat :)

dcr-stripe (approved):
LGTM! Thanks Pavel!

src/Stripe.net/Infrastructure/StringUtils.cs
```diff
@@ -70,5 +70,22 @@ public static bool ContainsWhitespace(string str)
 
             return whitespaceRegex.IsMatch(str);
         }
+
+        /// <summary>
+        /// Trims the beta header part of an API Version string.
+        /// For example, "22-12-2022; orders_beta=v3" is converted to "22-12-2022".
```
dcr-stripe, line 76: Nit: The date format is inversed (should be 2022-12-22)

PR #2599, opened 2022-11-04
Merge master and update generated code

pakrym-stripe:


src/Stripe.net/Entities/LineItems/LineItem.cs
```diff
@@ -79,41 +78,6 @@ public class LineItem : StripeEntity<LineItem>, IHasId, IHasObject
         [JsonProperty("price")]
         public Price Price { get; set; }
 
-        #region Expandable Product
```
richardm-stripe, line 82: ?

pakrym-stripe, line 82: part of orders

pakrym-stripe: https://git.corp.stripe.com/stripe-internal/pay-server/pull/525592

PR #2619, opened 2022-12-13
Fixed Patching Improper Handling of Exceptional Conditions in Stripe

iot-defcon:
Newtonsoft.Json prior to version 13.0.1 is vulnerable to Insecure Defaults due to improper handling of expressions with high nesting level that lead to StackOverFlow exception or high CPU and RAM usage. Exploiting this vulnerability results in Denial Of Service (DoS).

The serialization and deserialization path have different properties regarding the issue.

Deserializing methods (like JsonConvert.DeserializeObject) will process the input that results in burning the CPU, allocating memory, and consuming a thread of execution. Quite high nesting level (>10kk, or 9.5MB of {a:{a:{... input) is needed to achieve the latency over 10 seconds, depending on the hardware.

Serializing methods (like JsonConvert.Serialize or JObject.ToString) will throw StackOverFlow exception with the nesting level of around 20k.

To mitigate the issue one either need to update Newtonsoft.Json to 13.0.1 or set MaxDepth parameter in the JsonSerializerSettings. This can be done globally with the following statement. After that the parsing of the nested input will fail fast with Newtonsoft.Json.JsonReaderException:
```
JsonConvert.DefaultSettings = () => new JsonSerializerSettings { MaxDepth = 128 };
```

[CWE-755](https://cwe.mitre.org/data/definitions/755.html)
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`

### DIFF
--- a/src/Stripe.net/Stripe.net.csproj
+++ b/src/Stripe.net/Stripe.net.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>Stripe.net is a sync/async .NET 4.6.1+ client, and a portable class library for the Stripe API.  (Official Library)</Description>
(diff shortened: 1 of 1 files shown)


pakrym-stripe:
Mitigation for this issue was done in https://github.com/stripe/stripe-dotnet/pull/2555. We'd prefer to stay on existing version of Newtonsoft.Json to avoid forcing breaking changes on our customers.

PR #2622, opened 2022-12-17
Fix missing IDisposable using

NoahStahl:
While running [InferSharp ](https://github.com/microsoft/infersharp)on a project using this library, a finding came up for a missing `using` for an `IDisposable`. After adding this, the warning is resolved. So, figured it would be nice to fix to avoid confusion for stripe-dotnet users who use this tool.

I attempted to run the tests but don't have the mock service in place, so I haven't verified that aspect. If needed, I can pursue.

The finding message, for reference:
```
/_/src/Stripe.net/Infrastructure/Public/StripeStreamedResponse.cs:47: error: Pulse Resource Leak
  Resource dynamically allocated by constructor System.IO.StreamReader() on line 46 is not closed after the last access at line 47, column 13.
```


pakrym-stripe (approved):

NoahStahl:
Is there anything else to be done for this one to merge?

richardm-stripe:
Thanks for following up @NoahStahl! Looks like this one fell through the cracks. Triggering the merge.

PR #2625, opened 2022-12-21
Stop `FileService` from mutating its `RequestOptions` argument

richardm-stripe:
## Notify
r? @anniel-stripe 

## Summary
A user reported getting an error like `Unrecognized request URL (GET: /v1/terminal/configurations)` after reusing RequestOptions in FileService and then the ConfigurationService, because FileService mutates RequestOptions and sets the base to files.stripe.com and `files.stripe.com/v1/terminal/configurations` doesn't exist.

This is unintuitive. Fixing it shouldn't be considered breaking. I can't imagine that any user is dependent on this behavior.

### DIFF
--- a/src/Stripe.net/Services/_common/RequestOptions.cs
+++ b/src/Stripe.net/Services/_common/RequestOptions.cs
@@ -2,7 +2,7 @@ namespace Stripe
 {
     using System;
 
-    public class RequestOptions
+    public class RequestOptions : ICloneable
     {
         /// <summary>
         /// Gets or sets the <a href="https://stripe.com/docs/api/authentication?lang=dotnet">API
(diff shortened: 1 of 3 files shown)


src/Stripe.net/Services/_common/RequestOptions.cs
```diff
@@ -2,7 +2,7 @@ namespace Stripe
 {
     using System;
 
-    public class RequestOptions
+    public class RequestOptions : ICloneable
```
cecilphillip, line 5: We can exclude this interface. You can exclude it.

src/Stripe.net/Services/_common/RequestOptions.cs
```diff
@@ -38,5 +38,10 @@ public class RequestOptions
         /// is only used for creating ephemeral keys, which require a specific API version.
         /// </remarks>
         internal string StripeVersion { get; set; }
+
+        public object Clone()
```
cecilphillip, line 42: - I don't think this method needs to be available to our users. We should be able to switch it from being public to it `internal`

cecilphillip: those changes shouldn't require any changes to your unit test(s)

anniel-stripe (approved):

PR #2636, opened 2023-01-24
Drop builds for out of support frameworks

iamcarbon:
.NET Framework 4.5.1, .NETCORE 3.1 & .NET 5.0 have gone out of support and are no longer updated by Microsoft.

This PR drops these targets from the main library build, and raises the minimum .NET Framework target to 4.6.2.

Note that the library can still be utilized by .NETCORE 3.1 & .NET 5.0 via the NETSTANDARD2.0 target. These targets also continued by tested by the test configuration.


pakrym-stripe:
Hi, thank you for your contribution!

Stripe libraries historically kept longer support for out of date runtimes. When choosing to deprecate runtime support we compare the benefits of deprecating with the cost of support. 

In this particular case supporting these .NET version doesn't incur extra cost so we would prefer to keep them.

PR #2690, opened 2023-05-04
StripeClient.RawRequest

anniel-stripe:


src/Stripe.net/Infrastructure/Public/StripeRequest.cs
```diff
@@ -5,15 +5,18 @@ namespace Stripe
...
     using System.Text;
-    using Stripe.Infrastructure;
     using Stripe.Infrastructure.FormEncoding;
 
     /// <summary>
     /// Represents a request to Stripe's API.
     /// </summary>
     public class StripeRequest
     {
+        // Either options os content can be set.
```
anniel-stripe, line 15:
```suggestion
        // Either options or content can be set.
```

src/Stripe.net/Services/PreviewService.cs
```diff
@@ -0,0 +1,38 @@
...
+    public class PreviewService
+    {
+        private readonly StripeClient client;
+
+        internal PreviewService(StripeClient client)
+        {
+            this.client = client;
+        }
+
+        public Task<StripeResponse> Get(string path, RawRequestOptions requestOptions = null, CancellationToken cancellationToken = default)
```
anniel-stripe, line 16: Do we need to call these `GetAsync`, etc. since it calls the async version of `RawRequestAsync`? Should we make sync versions too?

src/Stripe.net/Services/PreviewService.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public Task<StripeResponse> Delete(string path, RawRequestOptions requestOptions = null, CancellationToken cancellationToken = default)
+        {
+            return this.client.RawRequestAsync(HttpMethod.Delete, path, null, BuildPreviewOptions(requestOptions), cancellationToken);
+        }
+
+        private static RawRequestOptions BuildPreviewOptions(RawRequestOptions requestOptions)
+        {
+            RawRequestOptions cloned = requestOptions != null ? (RawRequestOptions)requestOptions.Clone() : new RawRequestOptions();
+            cloned.ApiMode = ApiMode.Preview;
```
anniel-stripe, line 34: Oh, I guess we don't need to hardcode the preview version when [there's already logic ](https://github.com/stripe/stripe-dotnet/pull/2690/files#diff-ac904c400fdd4a5fb6b51487f1d8b4f8b5cc1076c67c44b590c9b0ce53a5ecc3R186-R187)to set the Stripe-Version header if the API mode is preview.

src/Stripe.net/Infrastructure/Public/IStripeClient.cs
```diff
@@ -47,43 +47,6 @@ public interface IStripeClient
...
 
-        /// <summary>Sends a request to Stripe's API as a synchronous operation.</summary>
-        /// <param name="method">The HTTP method.</param>
-        /// <param name="path">The path of the request.</param>
-        /// <param name="options">The parameters of the request.</param>
-        /// <param name="requestOptions">The special modifiers of the request.</param>
-        /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
-        /// <returns>The response as a StripeResponse.</returns>
-        /// <exception cref="StripeException">Thrown if the request fails.</exception>
-        StripeResponse RawRequest(
```
anniel-stripe: Why don't we want to add this to `IStripeClient`? I get an error in my IDE when I try to use this in my playground project:
```
'IStripeClient' does not contain a definition for 'RawRequestAsync' and no accessible extension method 'RawRequestAsync' accepting a first argument of type 'IStripeClient' could be found (are you missing a using directive or an assembly reference?)
```

anniel-stripe (commented):
some nits / questions

src/Stripe.net/Infrastructure/Public/IStripeClient.cs
pakrym-stripe: I'm a bit bit of an opponent to .NET interfaces and their overuse. Interfaces are impossible to evolve in a backwards compatible way and most of the time don't serve a good purpose (testing is a popularly cited one but virtual methods on deal with that almost as well).

I expect we'll be putting more properties on `StripeClient` (service navigator) and we won't be able to change the interface every time we add a property (because that will be a breaking change).

I'd like to keep IStripeClient similar to ResponseGetter in Java or RequestSender in Node where it has minimum of methods to be able to send a request. Service implementations will still pass IStripeClient around and use it for their implementation but `IStripeClient` won't have everything that `StripeClient` itself does.

anniel-stripe: Got it, I trust you on this front. :) Thanks for the explanation!

anniel-stripe:
LGTM!

PR #2693, opened 2023-05-09
Allow for developers to pass null in for tolerance that disables the time checking.

AraHaan:
This would allow for developers who intend to have the entire timestamp checking to be disabled if they determine that time checking is more of a problem than what it is worth for them. This is for the webhook processing by the way.


AraHaan:
Oh, another issue ``(long - int) > int`` checking which can also cause problems as well. I just now noticed that as well.

src/Stripe.net/Services/Events/EventUtility.cs
```diff
@@ -156,12 +156,14 @@ public static void ValidateSignature(string json, string stripeSignatureHeader,
...
             }
 
-            var webhookUtc = Convert.ToInt32(signatureItems["t"].FirstOrDefault());
-
-            if (Math.Abs(utcNow - webhookUtc) > tolerance)
+            if (tolerance != null)
             {
-                throw new StripeException(
-                    "The webhook cannot be processed because the current timestamp is outside of the allowed tolerance.");
+                var webhookUtc = Convert.ToInt32(signatureItems["t"].FirstOrDefault());
```
AraHaan, line 161:
```suggestion
                var webhookUtc = Convert.ToInt64(signatureItems["t"].FirstOrDefault());
```

To convert this type to a ``long`` for the math that comes right afterwards.

clement911:
We recently started using the `EventUtility.ValidateSignature` to not only validate webhook signatures but to also verify our stripe app's signature. We follow [this doc](https://stripe.com/docs/stripe-apps/build-backend#send-a-signed-request) (although the docs are in node) .
It works well. We use the app signing secret instead of the webhook secret.
That's the right thing to do right?

If so, you might want to change the [exception message](https://github.com/stripe/stripe-dotnet/blob/a9dc0f4bcea8821193d35403a34d1497f3145f0c/src/Stripe.net/Services/Events/EventUtility.cs#L148) because they refer specifically to webhook and webhook secret, which would be a confusing message to get for a stripe app signature.

AraHaan:
With that logic all of the exception messages in that file would need to be edited, and I feel like the ``EventUtility`` class was made specifically with webhooks in mind only. However I do not disagree that perhaps the ``VerifySignature`` code could be put inside of a static class that can be shared with app signing secret code and the ``EventUtility`` code.

AraHaan:
I moved the ``ValidateSignature`` to a new class of ``AppSignature`` however to avoid an ABI break I duplicated the original methods that then calls into them.

src/Stripe.net/Services/Apps/Signatures/AppSignature.cs
```diff
@@ -0,0 +1,83 @@
...
+                throw new StripeException(
+                    "The webhook or app signature cannot be processed because the signature cannot be safely calculated.",
+                    ex);
+            }
+
+            if (!IsSignaturePresent(signature, signatureItems["v1"]))
+            {
+                throw new StripeException(
+                    "The expected signature was not found in the Stripe-Signature header. " +
+                    "Make sure you're using the correct webhook secret (whsec_) or app secret (apsec_) and confirm the incoming request came from Stripe.");
```
AraHaan, line 40: I am not sure if ``apsec_`` here is correct.

src/StripeTests/Services/Events/EventUtilityTest.cs
```diff
@@ -56,7 +56,7 @@ public void RejectIncorrectSignature()
 
             Assert.Equal(
                 "The expected signature was not found in the Stripe-Signature header. " +
-                "Make sure you're using the correct webhook secret (whsec_) and confirm the incoming request came from Stripe.",
+                "Make sure you're using the correct webhook secret (whsec_) or app secret (apsec_) and confirm the incoming request came from Stripe.",
```
AraHaan, line 59: I am not sure if ``apsec_`` here is correct.

pakrym-stripe:
Hi, @AraHaan thank you for your contribution. 

For the initial goal of suppressing the timestamp check, does passing the `Int64.MaxValue` help?

clement911:
My app signing secret starts with `absec_`, not `apsec_`. I'm not sure `ab` standards for...

AraHaan:
> Hi, @AraHaan thank you for your contribution.
> 
> For the initial goal of suppressing the timestamp check, does passing the `Int64.MaxValue` help?

I guess it could seemingly work as well. However, I prefer the nullable option more in a way though.

pakrym-stripe:
> However, I prefer the nullable option more in a way though.

I understand but changing the types parameter is a binary braking change. I don't think it's worth making a breaking change for something that's already possible to do with the library.

AraHaan:
Technically it's not breaking at all, all it does is just marks the type as nullable without changing the underlying type at all.

pakrym-stripe:
@AraHaan I do appreciate the effort you've put into this PR but we can't take it.

There are a few separate changes in this PR so to address them individually:
1. Ability to ignore timestamp - this is already possible. We don't recommend customers disable timestamp checks in signatures and are not looking to provide the easiest possible way to do it.
2. Separate entrypoint for app signature validation - we'd prefer to design a consistent API for all the languages before starting the implementation. For now customers are unblocked using existing Webhook methods.

PR #2700, opened 2023-05-23
Handle developer_message in preview error responses

pakrym-stripe:
The error message can be returned in a `developer_message` field, handle this case and throw a useful exception.


anniel-stripe (approved):
LGTM!

anniel-stripe (approved):
LGTM!

PR #2744, opened 2023-08-07
More consistent AnyOf<...> generation

richardm-stripe:
## Changelog
* Generate more accurate types for `AccountSettingsPayoutsScheduleOptions.DelayDays`, `SubscriptionSchedulePhaseOptions.EndDate`, `SubscriptionSchedulePhaseOptions.StartDate`,
`SubscriptionSchedulePhaseOptions.TrialEnd`,

### DIFF
--- a/src/Stripe.net/Services/Accounts/AccountBankAccountOptions.cs
+++ b/src/Stripe.net/Services/Accounts/AccountBankAccountOptions.cs
@@ -5,6 +5,9 @@ namespace Stripe
 
     public class AccountBankAccountOptions : INestedOptions
     {
+        [JsonProperty("object")]
+        internal string Object => "bank_account";
+
         /// <summary>
         /// The name of the person or business that owns the bank account.This field is required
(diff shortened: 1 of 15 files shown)


src/Stripe.net/Services/Accounts/AccountBankAccountOptions.cs
```diff
@@ -5,9 +5,6 @@ namespace Stripe
 
     public class AccountBankAccountOptions : INestedOptions
     {
-        [JsonProperty("object")]
-        internal string Object => "bank_account";
-
```
remi-stripe, line 10: This change looks like a potential bug, it's an implicit value that is required for the API to work and we don't want to force developers to set.

richardm-stripe: Great catch, brought it back!

pakrym-stripe (approved):

PR #2747, opened 2023-08-10
Import UnixDateTimeConverter from Newtonsoft

richardm-stripe:


richardm-stripe:
Oops, the time is not yet ripe. We use Newtonsoft.Json < 11 for net461 still: https://github.com/stripe/stripe-dotnet/blob/b9b4507d101b1ed3ba9cff63723f8d55ff67d60d/src/Stripe.net/Stripe.net.csproj#L66

PR #2794, opened 2023-11-06
Generate CardCreateNestedOptions

pakrym-stripe:
And make `Object` fields public.


src/Stripe.net/Services/Accounts/AccountBankAccountOptions.cs
```diff
@@ -3,11 +3,8 @@ namespace Stripe
 {
     using Newtonsoft.Json;
 
-    public class AccountBankAccountOptions : INestedOptions
+    public class AccountBankAccountOptions : INestedOptions, IHasObject
```
richardm-stripe, line 6: This really feels wrong, the docstring of IHasObject is 

"Interface that identifies entities returned by Stripe that have an `object` attribute." and this is not an entity returned by Stripe.

richardm-stripe (approved):

src/Stripe.net/Services/Accounts/AccountBankAccountOptions.cs
richardm-stripe: ah I guess the ship has sailed

PR #2804, opened 2023-11-29
Support passing in options to singleton retrieve methods

anniel-stripe:
## Changelog
* Add support for passing request parameters to `Get` on `Balance` and `Tax.Settings`.


src/Stripe.net/Services/_interfaces/ISingletonRetrievable.cs
```diff
@@ -3,6 +3,8 @@ namespace Stripe
     using System.Threading;
     using System.Threading.Tasks;
 
+    // TODO (MAJOR): Accept a TOptions generic and use it to define method overloads
+    // that allow passing in options.
```
anniel-stripe, line 7: I'm thinking we should update this TODO so that we aren't defining overloads, but rather replacing the existing `Get` methods with those that accept both options and request options.
```suggestion
    // TODO (MAJOR): Accept a TOptions generic and use it to define methods
    // that accept options parameters.
```

src/Stripe.net/Services/_interfaces/ISingletonRetrievable.cs
```diff
@@ -3,6 +3,8 @@ namespace Stripe
     using System.Threading;
     using System.Threading.Tasks;
 
+    // TODO (MAJOR): Accept a TOptions generic and use it to define methods
```
pakrym-stripe, line 6: I'm not sure whether this change is worth making. We can define another interface later if there is interest.

pakrym-stripe (approved):

src/Stripe.net/Services/_interfaces/ISingletonRetrievable.cs
anniel-stripe: Sounds good, I'll remove this TODO

PR #2807, opened 2023-12-07
Add 'Usage' to RequestTelemetry

richardm-stripe:
## Summary
Adds a public `Usage` property to `RequestTelemetry` and an internal `Usage` field to `RequestOptions` and `StripeRequest`.


src/Stripe.net/Infrastructure/Public/RequestTelemetry.cs
```diff
@@ -47,7 +47,8 @@ public void MaybeAddTelemetryHeader(IDictionary<string, string> headers)
         /// </summary>
         /// <param name="response">The HTTP response message.</param>
         /// <param name="duration">The request duration.</param>
-        public void MaybeEnqueueMetrics(HttpResponseMessage response, TimeSpan duration)
+        /// <param name="usage">Tracked behaviors.</param>
+        public void MaybeEnqueueMetrics(HttpResponseMessage response, TimeSpan duration, List<string> usage = null)
```
richardm-stripe, line 51: I think this is technically breaking because anybody who subclassed RequestTelemetry would now have to implement a different signature, but I think this is unlikely enough that we can disregard.

src/StripeTests/Functional/TelemetryTest.cs
```diff
@@ -61,7 +63,40 @@ public void TelemetryWorks()
...
+                            (_) => true)),
+                    ItExpr.IsAny<CancellationToken>());
+        }
+
+        [Fact]
+        public void TelemetryIncludesUsage()
+        {
+            this.MockHttpClientFixture.Reset();
+            var fakeServer = FakeServer.ForMockHandler(this.MockHttpClientFixture.MockHandler);
+            fakeServer.Delay = TimeSpan.FromMilliseconds(20);
```
pakrym-stripe, line 76: nit: rem?

pakrym-stripe (approved):

src/Stripe.net/Infrastructure/Public/RequestTelemetry.cs
pakrym-stripe: should we add an overload instead?

richardm-stripe: Done!

src/Stripe.net/Infrastructure/Public/RequestTelemetry.cs
```diff
@@ -48,6 +48,18 @@ public void MaybeAddTelemetryHeader(IDictionary<string, string> headers)
...
+        }
+
+        /// <summary>
+        /// If telemetry is enabled and the queue is not full, then enqueue a new metrics item;
+        /// otherwise, do nothing.
+        /// </summary>
+        /// <param name="response">The HTTP response message.</param>
+        /// <param name="duration">The request duration.</param>
+        /// <param name="usage">Tracked behaviors.</param>
+        public void MaybeEnqueueMetrics(HttpResponseMessage response, TimeSpan duration, List<string> usage = null)
```
pakrym-stripe, line 62:
```suggestion
        public void MaybeEnqueueMetrics(HttpResponseMessage response, TimeSpan duration, List<string> usage)
```

richardm-stripe:
Re-requesting review since I changed to the overload.

I think in the next major we should make the entire `RequestTelemetry` internal? It's fairly silly to have to think about external users of this.

src/StripeTests/Functional/TelemetryTest.cs
```diff
@@ -154,8 +191,36 @@ public void NoTelemetryWhenDisabled()
...
+                : base(client)
+            {
+            }
+
+            public override string BasePath => "/v1/test";
+
+            public virtual void MakeRequestWithUsage(RequestOptions requestOptions)
+            {
+                RequestOptions ro = requestOptions.Clone();
+                ro.Usage = new List<string> { "llama", "bufo" };
```
pakrym-stripe, line 221: for the actual implementation we should try and avoid creating new lists and throwing them away for every call. Maybe a static field somewhere?

pakrym-stripe (approved):

src/StripeTests/Functional/TelemetryTest.cs
richardm-stripe: Sounds good, I made the test reflect this to make it more likely to remember, but I'll bear in mind, too.

PR #2820, opened 2023-12-22
Run compat check only in pull-requests

pakrym-stripe:
Otherwise, it tries to check compatibility during the release and fails.


anniel-stripe (approved):
Why does the compat check fail during release? Because the version changes?

pakrym-stripe:
The compat check uses the current version in the `.csproj` as the base version to check agains. During the release, the current version in `.csproj` is the version being released. So the package can't be found.

PR #2825, opened 2024-01-04
Test: telemetry test reflect intended usage

richardm-stripe:
Intended to include this commit with https://github.com/stripe/stripe-dotnet/pull/2807/files/c584718081164fb6d17f2cecf502664d815e2e5f#diff-4693d4e413d7f17a62815d4a88bbaecaa9ae53431c67e34665d282d3ca982a64 but automerge messed me up


src/StripeTests/Functional/TelemetryTest.cs
```diff
@@ -213,12 +213,16 @@ public TestService(IStripeClient client)
             {
             }
 
+            // Real usage should mirror this and use a static readonly list, instead of creating
+            // a new list every time.
+            public static readonly List<string> testUsage = new List<string> { "llama", "bufo" };
```
pakrym-stripe, line 218:
```suggestion
            private static readonly List<string> testUsage = new List<string> { "llama", "bufo" };
```

pakrym-stripe (approved):

PR #2852, opened 2024-02-16
Shadow Created from ListOptionsWithCreated

helenye-stripe:
Created has been moved inline into all children classes for discoverability purposes. ListOptionsWithCreated will be deprecated in the next major version.

### DIFF
--- a/src/Stripe.net/Services/Accounts/AccountListOptions.cs
+++ b/src/Stripe.net/Services/Accounts/AccountListOptions.cs
@@ -1,7 +1,14 @@
 // File generated from our OpenAPI spec
 namespace Stripe
 {
+    using System;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
     public class AccountListOptions : ListOptionsWithCreated
     {
+        [JsonProperty("created")]
+        [JsonConverter(typeof(AnyOfConverter))]
+        public AnyOf<DateTime?, DateRangeOptions> Created { get; set; }
     }
 }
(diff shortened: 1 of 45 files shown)


src/Stripe.net/Services/Accounts/AccountListOptions.cs
```diff
@@ -1,7 +1,14 @@
...
 {
+    using System;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+
     public class AccountListOptions : ListOptionsWithCreated
     {
+        [JsonProperty("created")]
+        [JsonConverter(typeof(AnyOfConverter))]
+        public AnyOf<DateTime?, DateRangeOptions> Created { get; set; }
```
pakrym-stripe, line 12: In .NET there is no "shadowing" of properties. You end up defining a separate property with the same name and it has bad side-effects.

pakrym-stripe:
```C#
var options = new AccountListOptions();
options.Created = DateTime.Now;

Console.WriteLine(((ListOptionsWithCreated)options.Created); // prints null
```

helenye-stripe: Gotcha! Thanks for the insight. I think it'd be breaking to simply remove this from the ListOptionsWithCreated class and move it into the children instead, am I understanding correctly?

pakrym-stripe: Yep, that's breaking as well.

PR #2862, opened 2024-03-01
Explicitly define paths in manually maintained services

richardm-stripe:
In [#2971](https://github.com/stripe/stripe-dotnet/pull/2791) we switch codegenerated methods to avoid helpers like `CreateEntityAsync` and `InstanceUrl` and instead define their paths directly.

Unfortunately the library contains several files that, for one reason or another, aren't code-generated and are manually maintained. This PR makes the analogous change in these files.

### DIFF
--- a/src/Stripe.net/Services/BankAccounts/BankAccountService.cs
+++ b/src/Stripe.net/Services/BankAccounts/BankAccountService.cs
@@ -29,72 +29,72 @@ namespace Stripe
 
         public virtual BankAccount Create(string parentId, BankAccountCreateOptions options, RequestOptions requestOptions = null)
         {
-            return this.CreateNestedEntity(parentId, options, requestOptions);
+            return this.Request<BankAccount>(HttpMethod.Post, $"/v1/customers/{parentId}/sources", options, requestOptions);
         }
 
         public virtual Task<BankAccount> CreateAsync(string parentId, BankAccountCreateOptions options, RequestOptions requestOptions = null, CancellationToken cancellationToken = default)
         {
-            return this.CreateNestedEntityAsync(parentId, options, requestOptions, cancellationToken);
+            return this.RequestAsync<BankAccount>(HttpMethod.Post, $"/v1/customers/{parentId}/sources", options, requestOptions, cancellationToken);
         }
 
         public virtual BankAccount Delete(string parentId, string id, BankAccountDeleteOptions options = null, RequestOptions requestOptions = null)
         {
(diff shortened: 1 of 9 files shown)


src/Stripe.net/Services/BankAccounts/BankAccountService.cs
```diff
@@ -29,73 +28,72 @@ public BankAccountService(IStripeClient client)
 
         public virtual BankAccount Create(string parentId, BankAccountCreateOptions options, RequestOptions requestOptions = null)
         {
-            return this.CreateNestedEntity(parentId, options, requestOptions);
+            return this.Request<BankAccount>(HttpMethod.Post, $"/v1/customers/{parentId}/sources", options, requestOptions);
```
helenye-stripe, line 31: qq: isn't it sometimes true that it uses Account instead of sources?

richardm-stripe: I don't think so. `CreateNestedEntity` [gets the path by calling](https://github.com/stripe/stripe-dotnet/blob/d21cf27c6a19bdcd13a2630b0eb3e8a82900009e/src/Stripe.net/Services/_base/ServiceNested.cs#L31) `this.ClassUrl`, which [ultimately reads the path from](https://github.com/stripe/stripe-dotnet/blob/d21cf27c6a19bdcd13a2630b0eb3e8a82900009e/src/Stripe.net/Services/_base/ServiceNested.cs#L196-L206) `this.BasePath` which is [hardcoded to](https://github.com/stripe/stripe-dotnet/blob/d21cf27c6a19bdcd13a2630b0eb3e8a82900009e/src/Stripe.net/Services/BankAccounts/BankAccountService.cs#L28) `/v1/customers/.../sources`

richardm-stripe: The logic in other libraries e.g. 
https://github.com/stripe/stripe-ruby/blob/master/lib/stripe/resources/bank_account.rb#L28-L31 doesn't apply in `stripe-dotnet` because there are no "instance" methods here that read properties of the returned resource, e.g. you don't do `account_source.delete(...)` or `customer_source.delete(...)` ever, in stripe-dotnet you would

`new ExternalAccountService().delete(...)` if you wanted to `DELETE /v1/accounts/{}/external_accounts` or `new BankAccountService().delete(...)` if you wanted to `DELETE /v1/customers/{}/sources`.

helenye-stripe (approved):

PR #2869, opened 2024-03-15
Add obsolete attributes to deprecated APIs and doc comments to methods

pakrym-stripe:


Makefile
```diff
@@ -4,7 +4,7 @@ update-version:
 	@perl -pi -e 's|<Version>[.\-\d\w]+</Version>|<Version>$(VERSION)</Version>|' src/Stripe.net/Stripe.net.csproj
 
 codegen-format:
-	dotnet format src/Stripe.net/Stripe.net.csproj
+	dotnet format src/Stripe.net/Stripe.net.csproj --severity warn
```
anniel-stripe, line 7: why was this necessary?

pakrym-stripe: Makes format slightly faster.

pakrym-stripe: As it doesn't try to fix issues that won't result in an error.

anniel-stripe (approved):

PR #2899, opened 2024-04-24
[Next Major] prevent autopagination from mutating

richardm-stripe:
This PR causes `.ListAutoPaging` and `.SearchAutoPaging` *not* to mutate the user-provided `options` argument.

Before this change, there would be a subtle bug/confusing behavior if you ever re-used the same `options` across multiple list requests: e.g. suppose there are 4 customers ["a", "b", "c", "d"].
```c#
var options = new CustomerListOptions()
{
  Limit = 2
};
var service = new CustomerService();

// GET /v1/customers?limit=2 ["a", "b"]
// GET /v1/customers?starting_after=b&limit=2 ["c", "d"]
foreach (var cus of in service.ListAutoPaging(options)) {
  ...
}

// Would not start at the beginning because `options` was mutated.
// GET /v1/customers?starting_after=b&limit=2 ["c", "d"]
foreach (var cus of in service.ListAutoPaging(options)) {
  ...
}
```

There's a small chance that a user is dependent on this behavior, e.g. you can imagine somebody doing
```c#
foreach (var cus of in service.ListAutoPaging(options)) { ... }

var firstIdOfLastPage = options.StartingAfter;
var didAnyPagination = options.StartingAfter != null;
```

so out of conservatism I would wait to change this behavior until the next major version but I'm open to persuasion just to ship it.

## Changelog
* Fixes `ListAutoPaging` and `ListAutoPagingAsync` to not mutate options object passed in.


jar-stripe (approved):
LGTM!

PR #2955, opened 2024-08-23
urlencode ids in url paths

xavdid-stripe:
This was a regression from https://github.com/stripe/stripe-dotnet/pull/2791

## Changelog

- id strings passed to service methods (like `PaymentMethodService.Get("...")`) are now correctly URL Encoded

### DIFF
--- a/src/StripeTests/Services/PaymentMethods/PaymentMethodServiceTest.cs
+++ b/src/StripeTests/Services/PaymentMethods/PaymentMethodServiceTest.cs
@@ -2,9 +2,11 @@ namespace StripeTests
 {
     using System.Collections.Generic;
     using System.Linq;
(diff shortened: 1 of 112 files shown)


src/StripeTests/Services/PaymentMethods/PaymentMethodServiceTest.cs
```diff
@@ -122,6 +124,21 @@ public void Get()
             Assert.Equal("payment_method", payment_method.Object);
         }
 
+        // load bearing test - this ensures that autogen methods with ids in the url
```
xavdid-stripe, line 127: Added this test - everything else is auto-gen

helenye-stripe (approved):

PR #2957, opened 2024-08-29
Generate SDK for OpenAPI spec version 1230

ramya-stripe:
Manually generating the changes for this week's release as the latest OpenAPI version 1230 has issues that are being worked on

## Changelog
* Add support for `StatusDetails` on `TestHelpersTestClock`
* Added new const `IssuingRegulatoryReporting` with value `issuing_regulatory_reporting` to be used as a FilePurpose


src/Stripe.net/Constants/FilePurpose.cs
```diff
@@ -23,6 +23,8 @@ public static class FilePurpose
 
         public const string IdentityDocumentDownloadable = "identity_document_downloadable";
 
+        public const string IssuingRegulatoryReporting = "issuing_regulatory_reporting";
```
richardm-stripe, line 26: missing from changelog

richardm-stripe (approved):
LGTM, probably a bug/omission in the changelog generation script though?

src/Stripe.net/Constants/FilePurpose.cs
ramya-stripe: Added manual entry

PR #2963, opened 2024-09-06
Add .editorconfig file with minimal change to align with StypeCopAnalyzer

jar-stripe:
Context:
Our StyleCopAnalyzer rules require prefacing properties, methods, fields, and events with this. but the default VS Code editor config prefers no preface. This sets the editor config to align with StyleCopAnalyzer and adds a file to include additional rules in the future.

Key Changes:
- added .editorconfig file that enables preference for `this.` before events, properties, fields, and methods


xavdid-stripe (approved):
love tooling changes!

PR #2979, opened 2024-09-27
Adding support for the new Usage Billing APIs & add services on StripeClient

ramya-stripe:
Adding support for the new Usage Billing APIs & add services on StripeClient


src/Stripe.net/Infrastructure/Public/StripeRequest.cs
```diff
@@ -5,15 +5,17 @@ namespace Stripe
...
     /// </summary>
     public class StripeRequest
     {
+        // Either options or content can be set.
+        // If content is set, it will be used instead of serialized options.
         private readonly BaseOptions options;
+        private readonly string content;
 
         /// <summary>Initializes a new instance of the <see cref="StripeRequest"/> class.</summary>
         /// <param name="client">The client creating the request.</param>
```
jar-stripe, line 21: I assume we're keeping this around for backwards compatility; should we add to the comment that this can not be used to make V2 API calls?

src/Stripe.net/Services/Events/EventUtility.cs
```diff
@@ -15,7 +15,7 @@ public static class EventUtility
         internal static readonly UTF8Encoding SafeUTF8
             = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
 
-        private const int DefaultTimeTolerance = 300;
```
jar-stripe, line 18: just confirming this was intentional to change it from private to public?

src/Stripe.net/Services/_base/Service.cs
```diff
@@ -299,29 +123,51 @@ protected async Task<Stream> RequestStreamingAsync(
...
+                    this.V2ListRequestAutoPagingAsync<T>(url, options, requestOptions));
+            }
+
             return AsyncUtils.ToEnumerable(
-                this.ListRequestAutoPagingAsync<T>(url, options, requestOptions));
+                this.V1ListRequestAutoPagingAsync<T>(url, options as ListOptions, requestOptions));
 #endif
         }
 
+        private void SetNextPage(BaseOptions options, string nextPage)
```
jar-stripe, line 154: method is no longer used and can be removed

src/Stripe.net/Stripe.net.csproj
```diff
@@ -2,13 +2,13 @@
 
   <PropertyGroup>
     <Description>Stripe.net is a sync/async .NET 4.6.1+ client, and a portable class library for the Stripe API.  (Official Library)</Description>
-    <Version>45.14.0</Version>
+    <Version>0.11.0</Version>
```
jar-stripe, line 5: fix version

src/Stripe.net/Stripe.net.csproj
```diff
@@ -2,13 +2,13 @@
...
     <LangVersion>8</LangVersion>
     <Authors>Stripe, Jayme Davis</Authors>
     <TargetFrameworks>net5.0;net6.0;netcoreapp3.1;netstandard2.0;net461</TargetFrameworks>
     <PackageTags>stripe;payment;credit;cards;money;gateway;paypal;braintree</PackageTags>
     <PackageIcon>icon.png</PackageIcon>
-    <PackageProjectUrl>https://github.com/stripe/stripe-dotnet</PackageProjectUrl>
+    <PackageProjectUrl>https://github.com/stripe/stripe-dotnet-next</PackageProjectUrl>
```
jar-stripe, line 11: fix url

src/Stripe.net/Stripe.net.csproj
```diff
@@ -39,17 +42,19 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.2" PrivateAssets="All" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
```
jar-stripe, line 42: did this not get merged into sdk-release/next-major?

src/StripeTests/StripeTests.csproj
```diff
@@ -6,6 +6,9 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <CheckEolTargetFramework>false</CheckEolTargetFramework>
     <NuGetAudit>false</NuGetAudit>
+
+    <!-- Workaround JetBrains Rider bug https://youtrack.jetbrains.com/issue/RIDER-103207/Cannot-suppress-vulnerable-package-errors -->
```
jar-stripe, line 10: is this ok to merge into the public repo?

src/Stripe.net/Infrastructure/Public/StripeRequest.cs
ramya-stripe: In what scenarios would one want to use it in v1?

jar-stripe: I'm not sure; I assume since its public it was kept in for backwards compatibility (i.e. if a user was creating it directly), but we hardcode the V1 api mode inside this method so in that case if a user tried to make V2 calls with it they would likely fail.

jar-stripe (approved):

PR #2982, opened 2024-09-30
Updated Obsolete comments and test file names

jar-stripe:
### Why
As part of the 9/30 launch, for clarity/ease of use, several classes and files were renamed and others were split or refactored.  As part of this refactor, we included the old methods (that were refactored into split-off services) as Obsolete methods.  We also deprecated some constructors in favor of StripeClient-based accessors.  This PR ensures the Obsolete tags are correct and that the test files align with the new names of their associated services.

### What
- adds missing Obsolete attributes to methods moved to .partial.cs files
- added SourceService.partial.cs and CreditNoteService.partial.cs, and added CreditNoteService PreviewLines accessor
- corrects Obsolete tag references
- removed obsolete statements above service constructors that accept IStripeClient argument (to add back in later if we need to)
- renames service test files to match the new names of those services.


src/Stripe.net/Services/BankAccounts/BankAccountService.cs
```diff
@@ -23,7 +23,7 @@ internal BankAccountService(ApiRequestor requestor)
         {
         }
 
-        [Obsolete("Please use the constructor that accepts an `ApiRequestor` instead.")]
+        [Obsolete("This member is deprecated. Access this service directly via the client with client.V1.BankAccounts")]
```
ramya-stripe, line 26: We are not really referring to the StripeClient class here? Should this be

> Access this service directly via the StripeClient with client.V1.BankAccounts

or 

> Access this service directly via the client with StripeClient.V1.BankAccounts

Probably not the second coz it makes it look like a static method

ramya-stripe: Also, are we sure we want to change this comment? From my understanding, this constructor was for folks who wanted to dig into the http internals, change a few things there and then run their requests. Isnt ApiRequestor is the right replacement in that case?

ramya-stripe: https://github.com/stripe/stripe-dotnet/?tab=readme-ov-file#using-a-custom-httpclient

src/Stripe.net/Services/OAuth/OAuthTokenService.cs
```diff
@@ -19,7 +19,7 @@ internal OAuthTokenService(ApiRequestor requestor)
         {
         }
 
-        [Obsolete("Please use the constructor that accepts an `ApiRequestor` instead.")]
+        [Obsolete("This member is deprecated and will be removed in a future release")]
```
ramya-stripe, line 22: Why is this comment different from the others where we are marking the constructor that takes IStripeClient as obsolete?

src/Stripe.net/Services/BankAccounts/BankAccountService.cs
ramya-stripe: > This member is deprecated

It is not really a "member" though

ramya-stripe (approved):

Issue #3006, opened 2024-10-16
OAuthTokenService Uses Incorrect Base URL for OAuth Requests After SDK Update


luca1897:
Hello,

After updating to the latest version of the Stripe .NET SDK (version 46.2.0), I've encountered an issue where the OAuth methods no longer function correctly. It appears that the SDK is concatenating the API base URL (https://api.stripe.com) instead of the Connect base URL (https://connect.stripe.com) when making OAuth requests.

### To Reproduce

Use the OAuthTokenService to request OAuth tokens with the following code snippet:
```csharp
var requestOptions = new RequestOptions
{
    ApiKey = ApiKey // Your API Key
};
var code = ""; // Your authorization code
var tokenService = new OAuthTokenService();
var tokenOptions = new OAuthTokenCreateOptions
{
    GrantType = "authorization_code",
    Code = code,
};
var stripeToken = await tokenService.CreateAsync(tokenOptions, requestOptions);
```

Run the code and observe the exception thrown.

An exception is thrown with the following message:
`Unrecognized request URL (POST: /oauth/token). Please see https://stripe.com/docs or we can help at https://support.stripe.com/.`

Stack Trace:
```
Stripe.StripeException:
   at Stripe.LiveApiRequestor.ProcessResponse (Stripe.net, Version=46.2.0.0, Culture=neutral, PublicKeyToken=null: /_/src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs:245)
   at Stripe.LiveApiRequestor+<RequestAsync>d__33`1.MoveNext (Stripe.net, Version=46.2.0.0, Culture=neutral, PublicKeyToken=null: /_/src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs:110)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1+ConfiguredTaskAwaiter.GetResult
   at Stripe.StripeClient+<RequestAsync>d__36`1.MoveNext (Stripe.net, Version=46.2.0.0, Culture=neutral, PublicKeyToken=null: /_/src/Stripe.net/Infrastructure/Public/StripeClient.cs:154)
[...]

PR #3007, opened 2024-10-17
Fixes ApiRequestorAdapter requests with BaseAddress other than Api

jar-stripe:
### Why?
v46 of the SDK introduced a service accessors into `StripeClient` and a internal ApiRequestor classes for implementing the actual request issuing and response parsing logic used by the service classes.  The `ApiRequestorAdapter` class is used to support older-style SDK usage patterns, e.g.
```csharp
var svc = new CustomerService();
var customer = svc.Get("cust_...");
```
and also to support any custom implementations of `IStripeClient`.  When a service is created directly like this, Stripe.net wraps the `StripeConfiguration.StripeClient` client object with this adapter to adapt the client to an ApiRequestor.

A user reported a problem with using the OAuthTokenService service with the pattern described above (https://github.com/stripe/stripe-dotnet/issues/3006) where the SDK was not using the correct API base address.  The root issue appears to be that BaseAddress specified by the OAuthTokenService.Create method is getting replaced with the API base address when it passes through the ApiRequestorAdapter, because the IStripeClient `RequestAsync` method does not accept a BaseAddress argument.  To fix this would introduce a breaking change in `IStripeClient`; this PR implements a backwards compatible fix by utilizing the `RequestOption.BaseUrl` property.

### What?
- changes `ApiRequestorAdapter` to set the `RequestOptions.BaseUrl` if the `baseAddress` argument is anything other than API
- change s`LiveApiRequestor` `MakeStripeRequest` to use the `BaseUrl` to override `baseAddress`, if BaseUrl is specified
- adds unit tests for ApiRequestorAdapter RequestAsync with different base addresses
- adds unit tests for OAuthTokenService Create and CreateAsync using the older-style instantiation pattern

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/ApiRequestor.cs
+++ b/src/Stripe.net/Infrastructure/Public/ApiRequestor.cs
@@ -1,5 +1,6 @@
 namespace Stripe
 {
+    using System;
     using System.IO;
     using System.Net.Http;
     using System.Threading;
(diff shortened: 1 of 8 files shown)


src/Stripe.net/Infrastructure/Public/ApiRequestor.cs
```diff
@@ -38,6 +39,23 @@ internal abstract class ApiRequestor
...
+                case BaseAddress.Api:
+                    return this.ApiBase;
+                case BaseAddress.Files:
+                    return this.FilesBase;
+                case BaseAddress.Connect:
+                    return this.ConnectBase;
+                case BaseAddress.MeterEvents:
+                    return this.MeterEventsBase;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(baseAddress), baseAddress, null);
```
xavdid-stripe, line 55: `ArgumentOutOfRangeException ` seems like an odd error type to throw (I'd expect something closer to python's `ValueError`), but if it's a standard c# thing i'm fine with it

xavdid-stripe (approved):
Looks good! Should also add a `## Changlog` in the PR description so it shows up in release notes

src/Stripe.net/Infrastructure/Public/ApiRequestor.cs
jar-stripe: That was there when I got here, but I do think it's right for c#: the docs for it say "The exception that is thrown when the value of an argument is outside the allowable range of values as defined by the invoked method." which applies here, I think.

PR #3011, opened 2024-10-24
Do not allow setting API Version directly on StripeConfiguration

ramya-stripe:
## Why?

When we introduced beta SDKs, we allowed users to directly update the global configuration for API Version since they needed to pass beta headers in #2561. We soon realized that was not safe and was error prone if users didnt pass in the right format and so introduced a helper method in https://github.com/stripe/stripe-dotnet/pull/2855 and advertised that as the right way of doing things in the README

Proper deserialization of classes from Events can be guaranteed only when the Webhook API version matches the API version used to generate the SDKs. Therefore, in this PR we are dropping the ability to directly update the API version.

## What?

- Resolving merge conflicts coming from running the auto merge tool after #3010
- Make the setter on StripeConfiguration.ApiVersion private
- Remove tests that tested the setting of StripeConfiguration.ApiVersion

## Changelog
* `StripeConfiguration.ApiVersion` is no longer settable. If you were using this to set the beta headers, use the helper method `StripeConfiguration.AddBetaVersion()` instead.

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
+++ b/src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
@@ -38,7 +38,7 @@ namespace Stripe
         public static string ApiVersion
         {
             get => apiVersion;
-            set
+            private set
             {
                 apiVersion = value;
                 TrimmedApiVersion = StringUtils.TrimApiVersion(value);
--- a/src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
+++ b/src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
@@ -24,7 +24,6 @@ namespace Stripe
         private static int maxNetworkRetries = SystemNetHttpClient.DefaultMaxNumberRetries;
 
         private static IStripeClient stripeClient;
-        private static string apiVersion;
 
         static StripeConfiguration()
         {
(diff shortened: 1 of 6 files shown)


src/Stripe.net/Services/Events/EventUtility.cs
```diff
@@ -17,6 +17,22 @@ internal static readonly UTF8Encoding SafeUTF8
...
+            // a release train, there's no way its compatible with this
+            // version
+            if (!eventApiVersion.Contains("."))
+            {
+                return false;
+            }
+
+            // versions are yyyy-MM-dd.train
+            var eventReleaseTrain = eventApiVersion.Split('.')[1];
+            var currentReleaseTrain = ApiVersion.Current.Split('.')[1];
```
jar-stripe, line 32: This needs to be the trimmed API version (i.e. the version without the beta version info)

src/Stripe.net/Infrastructure/Public/StripeConfiguration.cs
```diff
@@ -38,7 +38,7 @@ static StripeConfiguration()
         public static string ApiVersion
         {
             get => apiVersion;
-            set
+            private set
             {
                 apiVersion = value;
                 TrimmedApiVersion = StringUtils.TrimApiVersion(value);
```
jar-stripe, line 44: We should be able to remove TrimmedApiVersion since we are now only allowing users to set the Beta version info

jar-stripe: I just pushed a commit to the dotnet pr to remove TrimmedApiVersion and also simplify the ApiVersion property (we no longer need to keep a separate variable for storage)

PR #3012, opened 2024-10-25
Fix user agent in .net sdks

prathmesh-stripe:
Currently .NET SDKs rely on AssemblyName.Version to extract the version of Stripe SDK. This works for major releases but breaks for beta releases because how AssemblyVersion are calculated.
In a nutshell, a version `42.1.0-beta.1` would get converted to `42.1.0.0` in AssemblyName.Version. 

This new approach of storing SDK version as a constant is in-line with how we extract and update version in all our other SDKs.

### DIFF
--- a/src/StripeTests/StripeMockFixture.cs
+++ b/src/StripeTests/StripeMockFixture.cs
@@ -110,8 +110,8 @@ namespace StripeTests
         /// <returns>-1 if a &gt; b, 1 if a &lt; b, 0 if a == b.</returns>
         private static int CompareVersions(string a, string b)
         {
-            var version1 = new Version(a);
-            var version2 = new Version(b);
+            var version1 = new System.Version(a);
+            var version2 = new System.Version(b);
             return version2.CompareTo(version1);
         }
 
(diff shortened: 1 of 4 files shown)


src/StripeTests/StripeMockFixture.cs
```diff
@@ -110,8 +110,8 @@ public string GetFixture(string path, string[] expansions = null)
         /// <returns>-1 if a &gt; b, 1 if a &lt; b, 0 if a == b.</returns>
         private static int CompareVersions(string a, string b)
         {
-            var version1 = new Version(a);
-            var version2 = new Version(b);
+            var version1 = new System.Version(a);
```
jar-stripe, line 113: Given the issue in the PR description, does this properly keep the -beta or does this convert it to .0?

prathmesh-stripe: System.Version does not treat -beta properly. We don't need it since stripe-mock version does not have beta flag. 
I had to change this file was because of the addition of new Stripe.Version class. System.Version was being used earlier and this change just resolves ambiguity for the  compilers.

jar-stripe: Gotcha.  I would suggest leaving a comment in the code about this, but otherwise that makes sense.

jar-stripe (approved):
These changes look good to me.  Non blocking notes: I would add that comment explaining the limitation of System.Version and why it doesn't matter, and I _think_ that regex in the update-version make step is right but it would be worth testing it a few times before hitting the merge button.

PR #3014, opened 2024-10-25
Prathmesh/merge dotnet beta

prathmesh-stripe:
Running merge script to merge master into beta.

### DIFF
--- a/src/Stripe.net/Constants/Version.cs
+++ b/src/Stripe.net/Constants/Version.cs
@@ -1,0 +1,7 @@
+namespace Stripe
+{
+    internal class Version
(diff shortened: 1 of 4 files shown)


src/Stripe.net/Constants/Version.cs
```diff
@@ -0,0 +1,7 @@
+namespace Stripe
+{
+    internal class Version
+    {
+        public const string Current = "46.2.1";
```
prathmesh-stripe, line 5: This should automatically get updated before the next release.

xavdid-stripe (approved):

PR #3021, opened 2024-10-30
Urlencode ids in url paths inside legacy services

jar-stripe:
### Why
This is a follow on to https://github.com/stripe/stripe-dotnet/pull/2955.  That PR added functionality to encode ids in `Get` service methods.  In v46, we added several service files to maintain support for legacy service patterns (e.g. sub-resource access from the resource service).  This PR applies the same url encoding to ids in those service files.

### What
- Replace {id} with {WebUtility.UrlEncode(id)} in manually maintained services and partials

## Changelog
* id strings passed to service methods for legacy services (like SessionService.ListLineItems("...")) are now correctly URL Encoded


src/Stripe.net/Services/BankAccounts/BankAccountService.cs
```diff
@@ -40,22 +41,22 @@ public virtual Task<BankAccount> CreateAsync(string parentId, BankAccountCreateO
 
         public virtual BankAccount Delete(string parentId, string id, BankAccountDeleteOptions options = null, RequestOptions requestOptions = null)
         {
-            return this.Request<BankAccount>(BaseAddress.Api, HttpMethod.Delete, $"/v1/customers/{parentId}/sources/{id}", options, requestOptions);
+            return this.Request<BankAccount>(BaseAddress.Api, HttpMethod.Delete, $"/v1/customers/{parentId}/sources/{WebUtility.UrlEncode(id)}", options, requestOptions);
```
prathmesh-stripe, line 44: Wouldn't `parentId` also need to be escaped?

jar-stripe: Yes it definitely does.  Let me take another pass and make sure I escape all these.  Good catch!

jar-stripe: Ok, escaped parentId along with other named ids in other files.

prathmesh-stripe (approved):

PR #3026, opened 2024-11-05
Fix V2 list options base class

jar-stripe:
### Why
API V2 lists have different semantics than lists in V1.  In this SDK, we utilize a ListOptions base class to hold common options.  We need to introduce a V2 specific ListOptions base class to ensure the V2 list semantics are correctly represented.  This PR introduces this base class and updates V2 list options subclasses to derive from it instead of the v1 ListOptions class.

### What
- added ListOptions in the Stripe.V2 namespace (in _base/V2/ListOptions.cs, close to the V1 ListOptions file)
- updated V2 EventListOptions and EventDestinationListOptions to derive from V2.ListOptions


jar-stripe:
The compat issues are expected because we're correcting the base class of two list options classes to use the V2 ListOptions base class.

xavdid-stripe (approved):

PR #3029, opened 2024-11-12
Add .NET 9.0 Support

mattbrailsford:
This PR adds a new `net9.0` compilation target and also updates the `HttpClient` configuration in `SystemNetHttpClient` to use a `SocketsHttpHandler` to configure TLS1.2 support due to `ServicePointManager.SecurityProtocol` becoming obsolete.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "projects": [ "src" ],
   "sdk": {
-    "version": "8.0.100",
+    "version": "9.0.100-rc.2.24474.11",
     "rollForward": "latestFeature"
   }
 }
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "projects": [ "src" ],
   "sdk": {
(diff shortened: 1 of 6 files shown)


mattbrailsford:
I get the feeling the `SocketsHttpHandler` implementation could work for other dotnet versions but it didn't want to change anything for the existing versions just in case.

jar-stripe:
Hi @mattbrailsford, thanks for submitting this PR!  I've not yet looked at .NET 9 in depth but I did a bit of research from your description, and it appears from https://learn.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager?view=net-9.0 that this may be necessary to run Stripe.net on .NET 9; is that your experience as well, and is this a blocking issue for you?

jar-stripe:
@mattbrailsford Thanks for the update, and will do!  Please let us know if any issues do arise and we'll take a look.

PR #3030, opened 2024-11-13
support `AnyOf` with a single type

xavdid-stripe:
Some polymorphic fields only support a single variant initially. In order to keep a consistent interface for this field under the expectation that new values will be added, we need to support `AnyOf` expressions with only a single inner type.

### Method

I mostly copied `AnyOf<T1, T2>`, removing references to `T2` and changing `T1` to `T`. I also copied the tests, dropping some of the inputs and keeping the same assertions.

See: https://go/j/DEVSDK-2291

### DIFF
--- a/src/Stripe.net/Services/_base/AnyOf.cs
+++ b/src/Stripe.net/Services/_base/AnyOf.cs
@@ -22,6 +22,85 @@ namespace Stripe
     }
 
     /// <summary>
+    /// <see cref="AnyOf{T}"/> is a generic class that can hold a value of one type.
+    /// It's useful for polymorphic fields that (so far) only have a single valid type.
+    /// It uses implicit conversion operators to seamlessly accept or return either type.
+    /// This is used to represent polymorphic request parameters, i.e. parameters that can
+    /// be different types (typically a string or an options class).
+    /// </summary>
+    /// <typeparam name="T">The only possible type of the value.</typeparam>
+    [SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1402:FileMayOnlyContainASingleType", Justification = "Generic variant")]
+    public class AnyOf<T> : AnyOf
+    {
+        private readonly T value;
+        private readonly Values setValue;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AnyOf{T}"/> class with type <c>T</c>.
+        /// </summary>
+        /// <param name="value">The value to hold.</param>
+        public AnyOf(T value)
+        {
+            this.value = value;
+            this.setValue = Values.Value;
+        }
+
+        private enum Values
+        {
+            Value,
+        }
+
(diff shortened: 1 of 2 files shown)


src/Stripe.net/Services/_base/AnyOf.cs
```diff
@@ -21,6 +21,85 @@ public abstract class AnyOf : IAnyOf
         public override string ToString() => this.Value == null ? "AnyOf(null)" : this.Value.ToString();
     }
 
+    /// <summary>
+    /// <see cref="AnyOf{T}"/> is a generic class that can hold a value of one type.
```
jar-stripe, line 25: It might make this type a bit more readable in the context of the file if we kept this as T1, with the idea being that we set the expectation that for any AnyOf, the first type is T1, etc.  Not a blocking thing, though.

jar-stripe (approved):

xavdid-stripe:
@jar-stripe Are those test failures expected? I branched off latest `master` but I'm getting failures that seem related to your `ListOptions` changes?

src/Stripe.net/Services/_base/AnyOf.cs
xavdid-stripe: That's fair. I think in the context of the internals of this class, `T` makes more sense because there'll never be a `T2` to account for? But I don't feel super strongly either way.

jar-stripe:
> @jar-stripe Are those test failures expected? I branched off latest `master` but I'm getting failures that seem related to your `ListOptions` changes?

Errors about ListOptions are expected (we haven't done a release since I merged those, right?)  Are there any other errors ?

xavdid-stripe:
I don't think we've done a release, no. but it's fixed in master, right? so i'd expect PRs in the dotnet repo to have passing tests (but let me know if i'm misunderstanding)

src/Stripe.net/Services/_base/AnyOf.cs
jar-stripe: Same; I think your rationale makes sense.  The only exception is _maybe_ the case where I'm wading thru several variants of AnyOf and consistency here might make that easier, but I suspect the fact that this specific case hasn't come up yet means this doesn't/won't get touched often.

jar-stripe:
> I don't think we've done a release, no. but it's fixed in master, right? so i'd expect PRs in the dotnet repo to have passing tests (but let me know if i'm misunderstanding)

Nah, we require a release here.  dotnet's compat check uses the package validation feature built into .NET, which checks a newly built .nupkg against the last package published to nuget (see https://learn.microsoft.com/en-us/dotnet/fundamentals/apicompat/package-validation/baseline-version-validator).

xavdid-stripe:
Ah, got it! makes sense. 

I'm going to hold off merging to double check that this work will still be needed, but will merge once I know it is.

xavdid-stripe:
Ok, we don't need this! The API shaped to be more correct.

Issue #2495, opened 2022-05-26
Support JSON serialization with System.Text.Json


cecilphillip-stripe:
### Is your feature request related to a problem? Please describe.

Stripe.NET takes a hard dependency on the Netwonsoft.Json serializer which is fine for some cases. Starting with [ASP.NET Core 3.0](https://docs.microsoft.com/en-us/aspnet/core/release-notes/aspnetcore-3.0?view=aspnetcore-6.0#new-json-serialization), the default JSON serializer has changed to System.Text.JSON. 

For ASP.NET Core applications running on v3.0+ that use our Stripe.NET nuget package, developers have to be concerned with the inconsistencies between the serializers when working with Stripe Objects like Products, Customers, etc. 

Serialized single Stripe Product w/ Newtonsoft
```json
{
  "object": "product",
  "active": true,
  "attributes": [],
  "caption": null,
  "created": 1650997792,
  "deactivate_on": null,
  "default_price": null,
  "description": "This is such an awesome product made out of Metal",
  "images": [
    "https://picsum.photos/640/480/?image=279"
  ],
  "livemode": false,
  "metadata": {
    "code": "XXFYF3V"
  },
  "name": "Gorgeous Plastic Pants",
  "package_dimensions": null,
  "shippable": true,
  "statement_descriptor": null,
  "tax_code": null,
  "type": "service",
  "unit_label": null,
  "updated": 1650997792,
  "url": null
}
```

Serialized single Stripe Product w/ System.Text.Json
```json
{
  "Object": "product",
  "Active": true,
  "Attributes": [],
  "Caption": null,
  "Created": "2022-04-26T18:29:52Z",
  "DeactivateOn": null,
  "DefaultPriceId": null,
  "DefaultPrice": null,
  "Deleted": null,
  "Description": "This is such an awesome product made out of Metal",
  "Images": [
    "https://picsum.photos/640/480/?image=279"
  ],
  "Livemode": false,
  "Metadata": {
    "code": "XXFYF3V"
  },
  "Name": "Gorgeous Plastic Pants",
  "PackageDimensions": null,
  "Shippable": true,
  "StatementDescriptor": null,
  "TaxCodeId": null,
  "TaxCode": null,
  "Type": "service",
  "UnitLabel": null,
  "Updated": "2022-04-26T18:29:52Z",
  "Url": null,
  "RawJObject": null,
  "StripeResponse": null
}
```

[...]

Issue #1979, opened 2020-04-03
The collection type 'Newtonsoft.Json.Linq.JObject' on 'Stripe.Checkout.Session.RawJObject' is not supported.


sstalder:
Stripe v35.11.0

#1970

This breaks my dotnet core 3 project not using Newtonsoft.Json as the serializer, and I would assume anyone using System.Text.Json.

This is due to the JsonIgnore attribute from Newtonsoft not being used by System.Text.Json as it has it's own attribute.

There is some more information under this section on the migration guide: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#conditionally-ignore-a-property

> An unhandled exception has occurred while executing the request.
> System.NotSupportedException: The collection type 'Newtonsoft.Json.Linq.JObject' on 'Stripe.Checkout.Session.RawJObject' is not supported.
> at System.Text.Json.JsonPropertyInfoNotNullable`4.GetDictionaryKeyAndValueFromGenericDictionary(WriteStackFrame& writeStackFrame, String& key, Object& value)
> at System.Text.Json.JsonPropertyInfo.GetDictionaryKeyAndValue(WriteStackFrame& writeStackFrame, String& key, Object& value)
> at System.Text.Json.JsonSerializer.HandleDictionary(JsonClassInfo elementClassInfo, JsonSerializerOptions options, Utf8JsonWriter writer, WriteStack& state)
> at System.Text.Json.JsonSerializer.Write(Utf8JsonWriter writer, Int32 originalWriterDepth, Int32 flushThreshold, JsonSerializerOptions options, WriteStack& state)
> at System.Text.Json.JsonSerializer.WriteAsyncCore(Stream utf8Json, Object value, Type inputType, JsonSerializerOptions options, CancellationToken cancellationToken)
> at Microsoft.AspNetCore.Mvc.Formatters.SystemTextJsonOutputFormatter.WriteResponseBodyAsync(OutputFormatterWriteContext context, Encoding selectedEncoding)
> at Microsoft.AspNetCore.Mvc.Formatters.SystemTextJsonOutputFormatter.WriteResponseBodyAsync(OutputFormatterWriteContext context, Encoding selectedEncoding)
[...]

PR #3033, opened 2024-11-27
System.Text.Json Serialization Support in .NET SDK Objects

jar-stripe:
### Why?
Users reported in https://github.com/stripe/stripe-dotnet/issues/2495 and https://github.com/stripe/stripe-dotnet/issues/1979 that Stripe.net objects did not behave as expected when serialized using System.Text.Json.  System.Text.Json has become the standard for modern .NET, and the Stripe.net SDK objects should support serializing using it.

### What?
- Added System.Text.Json attributes for each property in the Stripe Entity and Options types
- Added Wholesome tests to verify that each type has the correct properties
- Implemented System.Text.Json converters for AnyOf, ExpandableField, StripeObject, UnixDateTime, and V2.Event
- Implemented System.Text.Json custom converter for IEnumerable (for StripeList and StripeSearchResult)
- Updated readme to describe JSON serialization, its usage, and any limitations.

## Changelog
- Add System.Text.Json support for serializing Stripe.net entities and objects for applications running on .NET 6 and above. Now you can pass a Stripe.net object or collection of objects to the System.Text.Json serializer and it will produce the correct JSON string.


jar-stripe:
The build failure is not expected, and appears to be due to a significant difference between .NET 6 and .NET 8.

jar-stripe:
Fixed issue with .NET 6 and tests pass, but a new test that compares the output of Newtonsoft Json.NET and STJ takes a while to run across all supported versions.  Maybe it should only run on .NET 8 normally, and across all .NET versions on release?

src/Stripe.net/Entities/Events/EventData.cs
```diff
@@ -12,13 +15,19 @@ public class EventData : StripeEntity<EventData>
         /// </summary>
         [JsonProperty("object")]
         [JsonConverter(typeof(StripeObjectConverter))]
+#if NET6_0_OR_GREATER
+        [STJS.JsonPropertyName("object")]
```
pakrym-stripe, line 19: I think I'll learn the answer soon enough but won't the lack of StripeObjectConverter make this model not directly deserializeable?

src/Stripe.net/Entities/Invoices/InvoiceTaxAmount.cs
```diff
@@ -2,19 +2,31 @@ namespace Stripe
 {
     using Newtonsoft.Json;
     using Stripe.Infrastructure;
+#if NET6_0_OR_GREATER
+    using STJS = System.Text.Json.Serialization;
+#endif
 
+#if NET6_0_OR_GREATER
+    [STJS.JsonConverter(typeof(STJMemberSerializationOptIn))]
```
pakrym-stripe, line 10: Have you considered a custom converter that you put onto the public expandable field?

src/Stripe.net/Entities/V2/Events/Event.partial.cs
```diff
@@ -5,18 +5,29 @@ namespace Stripe.V2
...
+#endif
+
     using Stripe.Infrastructure;
 
     /// <summary>
     /// Manually-maintained convenience methods added to ThinEvent.
     /// </summary>
     [JsonConverter(typeof(V2EventConverter))]
+#if NET6_0_OR_GREATER
+    [NoSystemTextJsonAttributesNeeded("Converter is only needed for deserialization")]
```
pakrym-stripe, line 19: I don't understand this message. Event is being deserialized, right?

src/Stripe.net/Entities/_base/StripeEntity.cs
```diff
@@ -25,9 +31,15 @@ public abstract class StripeEntity : IStripeEntity
         /// </remarks>
         /// <returns>The raw <see cref="JObject">JObject</see>.</returns>
         [JsonIgnore]
+#if NET6_0_OR_GREATER
+        [STJS.JsonIgnore]
+#endif
         public JObject RawJObject { get; protected set; }
```
pakrym-stripe, line 37: consider throwing a nice error.

src/Stripe.net/Entities/Events/EventData.cs
pakrym-stripe: Found it, the converter is on IHasObject

src/Stripe.net/Infrastructure/JsonConverters/STJAnyOfConverter.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            object o = null;
+
+            // Try to deserialize with each possible type
+            var elem = JsonElement.ParseValue(ref reader);
+            foreach (var type in typeToConvert.GenericTypeArguments)
+            {
+                try
+                {
+                    o = JsonSerializer.Deserialize(elem.GetRawText(), type, options);
```
pakrym-stripe, line 37: do we have to go through raw text here?

src/Stripe.net/Infrastructure/JsonConverters/STJAnyOfConverter.cs
```diff
@@ -0,0 +1,96 @@
...
+
+                    // If deserialization succeeds, stop
+                    break;
+                }
+                catch (JsonException)
+                {
+                    // Do nothing, just try the next type
```
pakrym-stripe, line 44: Is this how newtonsoft works today as well?

src/Stripe.net/Infrastructure/JsonConverters/STJStripeObjectConverter.cs
```diff
@@ -0,0 +1,129 @@
...
+                var objectValue = jsonObject.GetProperty("object").GetString();
+
+                Type concreteType = StripeTypeRegistry.GetConcreteType(typeToConvert, objectValue);
+                if (concreteType == null)
+                {
+                    // Couldn't find a concrete type to instantiate, return null.
+                    return null;
+                }
+
+                return (T)JsonSerializer.Deserialize(jsonObject.GetRawText(), concreteType, options);
```
pakrym-stripe, line 77: can we go through reader / element overload?

src/Stripe.net/Infrastructure/JsonConverters/STJStripeObjectConverter.cs
```diff
@@ -0,0 +1,129 @@
...
+
+                            var converter = property.GetConverter(options);
+                            converter.Write(writer, valueToSerialize, options);
+
+                            break;
+                    }
+                }
+
+                writer.WriteEndObject();
+                /*
```
pakrym-stripe, line 116: nit:rem

src/Stripe.net/Infrastructure/JsonConverters/STJStripeObjectConverter.cs
```diff
@@ -0,0 +1,129 @@
...
+                }
+
+                var jsonObject = JsonElement.ParseValue(ref reader);
+                var objectValue = jsonObject.GetProperty("object").GetString();
+
+                Type concreteType = StripeTypeRegistry.GetConcreteType(typeToConvert, objectValue);
+                if (concreteType == null)
+                {
+                    // Couldn't find a concrete type to instantiate, return null.
+                    return null;
```
pakrym-stripe, line 74: Don't we fall back to StripeObject in this case?

src/Stripe.net/Entities/Events/EventData.cs
jar-stripe: Yea; STJ doesn't have the ItemConverterType parameter in JsonPropertyName for converting objects inside collections; the solution for that is to add the converter to the inner type itself.  Because of this, the wholesome tests are written to validate interfaces have converters (i.e. for polymorphic resources in lists in other resources); this one could have gone either way but I wanted to keep it consistent with the other interfaces.

src/Stripe.net/Entities/Invoices/InvoiceTaxAmount.cs
jar-stripe: I did, but I couldn't get this work within a custom converter that is placed on the Id or the object.  When `Write` is called for a property, the converter excepts a writer that's already positioned for that field; to handle multiple fields I had to lift the converter up to a common level, which would be the class itself.  There are likely other options that involve restructuring how we represent expandable fields but they would have been too drastic a change for this PR, I think.

src/Stripe.net/Entities/V2/Events/Event.partial.cs
jar-stripe: Not through STJ.  This PR only supports serializing with STJ, so any conversion logic needed to deserialize an API response into a StripeObject can be omitted.  In some cases, for testing purposes or because it was easy to do so, I implemented both sides of a converter; in this case I punted but I would be game to go ahead and implement this to remove this potential point of confusion.

src/Stripe.net/Infrastructure/JsonConverters/STJStripeObjectConverter.cs
jar-stripe: I thought I got all those 😅

src/Stripe.net/Infrastructure/JsonConverters/STJStripeObjectConverter.cs
jar-stripe: Nah, in this case we return null.  This part is copied almost verbatim from https://github.com/stripe/stripe-dotnet/blob/4d0c154eb7d4a399d21504f1b9ce1f5be14e8cc6/src/Stripe.net/Infrastructure/JsonConverters/StripeObjectConverter.cs#L58

src/Stripe.net/Infrastructure/JsonConverters/STJAnyOfConverter.cs
jar-stripe: yep!  https://github.com/stripe/stripe-dotnet/blob/4d0c154eb7d4a399d21504f1b9ce1f5be14e8cc6/src/Stripe.net/Infrastructure/JsonConverters/AnyOfConverter.cs#L80

for a lot of these converters, I copied the NS one and changed the objects/methods to their equivalents.

src/Stripe.net/Infrastructure/JsonConverters/STJStripeObjectConverter.cs
jar-stripe: This was the closest I could find to what we currently do in NS but I think we may be able to make this more efficient with a little work.

src/Stripe.net/Infrastructure/JsonConverters/STJAnyOfConverter.cs
jar-stripe: Same as with STJStripeObjectConverter; I think this is closely aligned with what we do in NS, but we may be able to make this better.

src/Stripe.net/Entities/V2/Events/Event.partial.cs
jar-stripe: An update here: the V2EventConverter here is only used to deserialize the v2.core.event object into the specific event class as indicated by the type property.  After that, it's just an object and our existing attributes should be able to handle this.  However, in proving this to myself, I realized that this does not hold true for when we're converting a StripeList<V2.Event> because of how STJ supports converting items inside of a collection, so this does need a converter after all.

src/Stripe.net/Entities/_base/StripeEntity.cs
jar-stripe: I thought about this some more.  The problem is that once we serialize the object to JSON, we lose access any unsupported fields that would have been only on the API response but not deserialized into a C# property.  If a user were to deserialize from the serialized JSON, there would be no value to accessing these fields via RawJObject.  I like the idea of reporting an error but STJ would require us to put a converter on every concrete type in order for it to be effective.  I think the best course for now is a comment here, and a section in the readme describing what the support looks like and what its limitations are, and then we wait for customer feedback.

src/Stripe.net/Infrastructure/JsonConverters/STJAnyOfConverter.cs
jar-stripe: Update: there are overrides that I didn't see initially that accept the JsonElement, so we don't need to go thru raw text.

src/Stripe.net/Infrastructure/JsonConverters/STJStripeObjectConverter.cs
jar-stripe: Update: there is an override that I initially didn't notice that lets us pass jsonObject itself without going thru raw text

src/Stripe.net/Entities/V2/StripeList.cs
```diff
@@ -3,26 +3,43 @@ namespace Stripe.V2
...
 
     [JsonObject]
+#if NET6_0_OR_GREATER
+    [STJS.JsonConverter(typeof(STJEnumerableObjectConverter))]
+#endif
     public class StripeList<T> : StripeEntity<StripeList<T>>, IEnumerable<T>
     {
         /// <summary>
         /// A list containing the actual response elements, paginated by any request parameters.
+        /// TODO(jar) does this need an ItemConverterType (like in Stripe.StripeList).
```
jar-stripe, line 19: delete this

helenye-stripe (approved):
Thanks for the detailed tests, looks awesome 🚀

jar-stripe:
Compat failures are expected and ok, and appear to be because of this change: https://github.com/stripe/sdk-codegen/pull/1954

PR #3040, opened 2024-12-20
Add justfile, remove coveralls

xavdid-stripe:
### Why?

In an effort to modernize and simplify our local tooling, we're moving our dev commands from makefiles to [justfiles](https://github.com/casey/just). This is intended to be mostly a drop-in replacement, but some command names may change per standardization efforts.

### What?

- added justfile
- added warning line to top of makefile
- updated readme
- tweaked CI to use `just`
- removed all the coveralls-related code from CI
- updated readme to mention `just` (& an alternative)

### See Also

- [DEVSDK-2325](https://go/j/DEVSDK-2325)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: extractions/setup-just@v2
       - uses: actions/checkout@master
 
       - name: Setup .NET
@@ -47,7 +48,7 @@ jobs:
       - uses: stripe/openapi/actions/stripe-mock@master
 
       - name: Run test suite
-        run: make ci-test
+        run: just ci-test
 
       - name: Collect coverage
         run: dotnet test --no-build -c Release -f netcoreapp3.1 src/StripeTests/StripeTests.csproj /p:CollectCoverage=true /p:CoverletOutputFormat=opencover /p:ExcludeByAttribute=CompilerGenerated
@@ -60,33 +61,9 @@ jobs:
         if: github.event_name == 'pull_request'
         run: echo "commitBranch=$(echo ${GITHUB_HEAD_REF} | tr / -)" >> $GITHUB_ENV
 
-      - name: Send code coverage report to coveralls.io
-        if: env.COVERALLS_REPO_TOKEN
-        run: |
-          export ARGS="--opencover \
-            -i src/StripeTests/coverage.netcoreapp3.1.opencover.xml \
-            --repoToken $COVERALLS_REPO_TOKEN \
-            --useRelativePaths \
-            --commitId $commitId \
-            --commitBranch $commitBranch \
-            --commitAuthor $commitAuthor \
-            --jobId $jobId"
-          if [ ! -z "${pullRequestId}" ];
-          then
-            export ARGS="$ARGS \
-              --pullRequest $pullRequestId"
-          fi
-          dotnet tool run csmacnz.Coveralls $ARGS
-        env:
-          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
-          commitId: ${{ github.sha }}
-          commitAuthor: ${{ github.actor }}
-          jobId: ${{ github.run_id }}
-          pullRequestId: ${{ github.event.pull_request.number }}
-
       - name: Pack
         run: dotnet pack src/Stripe.net -c Release --no-build --output nuget
-      - name: 'Upload Artifact'
+      - name: "Upload Artifact"
         uses: actions/upload-artifact@v4
         with:
           name: nuget
@@ -97,13 +74,13 @@ jobs:
     if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
-    - name: Setup .NET
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: 8.0.x
-    - name: Run backcompat check
-      run: dotnet pack src/Stripe.net -p:RunBaselineCheck=true
+      - uses: actions/checkout@master
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 8.0.x
+      - name: Run backcompat check
+        run: dotnet pack src/Stripe.net -p:RunBaselineCheck=true
 
   publish:
     name: Publish
@@ -111,18 +88,18 @@ jobs:
     needs: [build]
     runs-on: ubuntu-latest
     steps:
-    - name: Download all workflow run artifacts
-      uses: actions/download-artifact@v4
-      with:
-        name: nuget
-        path: nuget
-    - name: Setup .NET
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: 8.0.x
-    - name: Publish NuGet packages to NuGet
-      run: dotnet nuget push nuget/*.nupkg --api-key ${{ secrets.NUGET_KEY }} --source "nuget.org"
-    - uses: stripe/openapi/actions/notify-release@master
-      if: always()
-      with:
-        bot_token: ${{ secrets.SLACK_BOT_TOKEN }}
+      - name: Download all workflow run artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: nuget
+          path: nuget
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 8.0.x
+      - name: Publish NuGet packages to NuGet
+        run: dotnet nuget push nuget/*.nupkg --api-key ${{ secrets.NUGET_KEY }} --source "nuget.org"
+      - uses: stripe/openapi/actions/notify-release@master
+        if: always()
+        with:
+          bot_token: ${{ secrets.SLACK_BOT_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,11 +24,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Install just
-        run: |
-          sudo apt-get update
-          sudo apt-get install just
-
+      - uses: extractions/setup-just@v2
       - uses: actions/checkout@master
 
       - name: Setup .NET
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,10 +50,6 @@ jobs:
       - name: Run test suite
         run: just ci-test
 
-      # commented out because it's slow, but it's here if we want it
-      # - name: Verify formatting
-      #   run: just format-check
-
       - name: Pack
         run: dotnet pack src/Stripe.net -c Release --no-build --output nuget
       - name: "Upload Artifact"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,17 +50,6 @@ jobs:
       - name: Run test suite
         run: just ci-test
 
-      - name: Collect coverage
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -24,6 +24,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: extractions/setup-just@v2
```
xavdid-stripe, line 27: I had to allow this in repo settings. I gave the code a quick readthrough and it's super straightforward: it builds the name of a crate ([extractions/setup-just | index.ts](https://github.com/extractions/setup-just/blob/trunk/src/index.ts)) and feeds it into another extension by the same author, which reads github releases to download the correct binary for the platform ([extractions/setup-crate | index.ts](https://github.com/extractions/setup-crate/blob/trunk/setup-crate/src/index.ts)).

It's the action recommended in the [just README](https://github.com/casey/just?tab=readme-ov-file#github-actions), so I feel good about it.

.github/workflows/ci.yml
```diff
@@ -60,33 +61,9 @@ jobs:
         if: github.event_name == 'pull_request'
         run: echo "commitBranch=$(echo ${GITHUB_HEAD_REF} | tr / -)" >> $GITHUB_ENV
 
-      - name: Send code coverage report to coveralls.io
```
xavdid-stripe, line 63: removed entirely, since we're not using coveralls

.github/workflows/ci.yml
```diff
@@ -60,33 +61,9 @@ jobs:
...
-        env:
-          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
-          commitId: ${{ github.sha }}
-          commitAuthor: ${{ github.actor }}
-          jobId: ${{ github.run_id }}
-          pullRequestId: ${{ github.event.pull_request.number }}
-
       - name: Pack
         run: dotnet pack src/Stripe.net -c Release --no-build --output nuget
-      - name: 'Upload Artifact'
```
xavdid-stripe, line 89: the rest of this file was changed by `prettier`; no other manual changes

README.md
```diff
@@ -320,10 +320,13 @@ go install github.com/stripe/stripe-mock@latest
 stripe-mock
 ```
 
+Lastly, we use [just](https://github.com/casey/just) for running common development tasks. You can also read the `justfile` and run those commands directly.
```
xavdid-stripe, line 323: I added this and the updated scripts below; everything else was `prettier`

justfile
```diff
@@ -0,0 +1,29 @@
+set quiet
+
+import? '../sdk-codegen/justfile'
```
xavdid-stripe, line 3: the trailing `?` on `import` means to swallow the error if this file is missing. So we'll get bonus actions, but external contributors will still be able to use the recipes in this file.

justfile
```diff
@@ -0,0 +1,29 @@
...
+
+_default:
+    just --list
+
+test no_build="" config="Debug":
+    dotnet test {{no_build}} -f net8.0 src/StripeTests/StripeTests.csproj -c {{config}}
+
+# skip build and test in release mode
+test-release: (test "--no-build" "Release")
+alias ci-test := test-release
```
xavdid-stripe, line 13: I added an alias because CI calls this and it's nice to be explicit about which recipe CI will be using. But we could just as easily amend the GH actions to call `test-release` if we preferred.

justfile
```diff
@@ -0,0 +1,29 @@
...
+import? '../sdk-codegen/justfile'
+
+_default:
+    just --list
+
+test no_build="" config="Debug":
+    dotnet test {{no_build}} -f net8.0 src/StripeTests/StripeTests.csproj -c {{config}}
+
+# skip build and test in release mode
+test-release: (test "--no-build" "Release")
```
xavdid-stripe, line 12: This could be simplified if testing just always ran the build command, but I get the sense that in CI we want to build once and use that build repeatedly (for tests, packing, uploading, etc). 

We could also add a `just build` command that was used as a pre-req for packing and uploading, but I'm not too worried about it.

justfile
```diff
@@ -0,0 +1,39 @@
...
+_default:
+    just --list --unsorted
+
+# base test command
+[no-quiet]
+_test no_build framework config:
+    dotnet test {{no_build}} {{framework}} src/StripeTests/StripeTests.csproj -c {{config}}
+
+# run tests in debug mode
+[group('useful')]
```
xavdid-stripe, line 14: I added these groups to better show which commands users will use most often, but i'm not sold on them. That gives us:

```
Available recipes:
    [useful]
    test               # run tests in debug mode
    format *options    # format files as needed

    [CI]
    ci-test            # skip build and don't specify the dotnet framework
    format-check       # verify, but don't modify, the project's formatting
```

instead of:

```
Available recipes:
    test               # run tests in debug mode
    ci-test            # skip build and don't specify the dotnet framework
    format *options    # format files as needed
    format-check       # verify, but don't modify, the project's formatting
```

We could also use no groups and alphabetize everything, but that felt like the worst option because then everything really was jumbled together:

```
Available recipes:
    ci-test            # skip build and don't specify the dotnet framework
    format *options    # format files as needed
    format-check       # verify, but don't modify, the project's formatting
    test               # run tests in debug mode
```

also happy to change names of groups.

justfile
```diff
@@ -0,0 +1,39 @@
...
+
+import? '../sdk-codegen/justfile'
+
+_default:
+    just --list --unsorted
+
+# base test command
+[no-quiet]
+_test no_build framework config:
+    dotnet test {{no_build}} {{framework}} src/StripeTests/StripeTests.csproj -c {{config}}
```
xavdid-stripe, line 11: This could be simplified if testing just always ran the build command, but I get the sense that in CI we want to build once and use that build repeatedly (for tests, packing, uploading, etc). 

We could also add a `just build` command that was used as a pre-req for packing and uploading, but I'm not too worried about it. Since this recipe isn't called directly, I think `test` and `ci-test` are clear enough on intention.

justfile
```diff
@@ -0,0 +1,39 @@
...
+alias codegen-format := format
+
+# verify, but don't modify, the project's formatting
+[group('CI')]
+format-check: (format "--verify-no-changes")
+
+# called by tooling
+[private]
+update-version version:
+    echo "{{ version }}" > VERSION
```
xavdid-stripe, line 37: This is the same as the makefile, but uses an action recipe argument and the just-style variable replacement (`{{ varname }}`). I verified that it works manually locally, though we'll have to tweak the code that calls this (the release script) once all the repos are done.

xavdid-stripe:
Ok, this is ready for review @jar-stripe! I over-commented in the PR since this is the first one and I want to explain my reasoning, but the actual changes should be fairly uncontroversial.

justfile
jar-stripe: Id agree with that.  unless we have pages of options (which I'm guessing we dont), i think its easier to digest as a simple list.  maybe include the group as part of the description, if it would add value there?

.github/workflows/ci.yml
jar-stripe: It would be nice to not have to rely on a 1 person organization for this.  is there value in us forking it and using the fork?  it doesn't look super complicated so I wouldn't expect it to get updated often, and having the fork gives us some stability if something changes for them

.github/workflows/ci.yml
```diff
@@ -47,46 +48,15 @@ jobs:
...
-              --pullRequest $pullRequestId"
-          fi
-          dotnet tool run csmacnz.Coveralls $ARGS
-        env:
-          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
-          commitId: ${{ github.sha }}
-          commitAuthor: ${{ github.actor }}
-          jobId: ${{ github.run_id }}
-          pullRequestId: ${{ github.event.pull_request.number }}
+      # commented out because it's slow, but it's here if we want it
```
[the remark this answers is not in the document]
jar-stripe: We format the code when generating, and StyleCop ensures that most (if not all) formatting issues are errors, so I'm guessing at least for dotnet we can assume if it builds its formatted correctly.  Also, performance seems to be an issue: on my machine, `just format-check` runs in ~78.38s and it looks like its taking 7-8 cores to do so.

justfile
xavdid-stripe: Yeah, that's a good idea. I'll play with starring the most useful commands (in the docstring)

.github/workflows/ci.yml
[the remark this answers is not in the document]
xavdid-stripe: sounds good, I can pull!

.github/workflows/ci.yml
xavdid-stripe: I'm not too worried about it, but I understand the concern. This approach takes care of supporting any platform, but given that we're only running on linux, we can probably get away with an `apt-get install just` ([docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners)), which I should have thought of in the first place.

Do you feel better about that? I don't really want to be in the business of maintaining a GH action, even if it's just to check and sync our fork periodically. It just smells like maintenance we're not going to do.

justfile
jar-stripe: I like this.  so a user that does not have codegen will still have access to repo-specific commands but perhaps not any commands that apply to or invoke the code generator?

.github/workflows/ci.yml
jar-stripe: oh right because this is just for CI.  i think thats reasonable, and if we end up having to support another platform (i.e. if we have to change a runner to be a macos runner for codesigning or something) its reasonable to assume we can find a way to install just thru an os-specific official channel there too.

justfile
xavdid-stripe: yep, exactly! progressive enhancement, basically

.github/workflows/ci.yml
xavdid-stripe: Yep, CI only. For local purposes, I can update our shared `mise` config and it'll "just work". 😁

jar-stripe: famous last words... 😆 
but yea, that sounds good to me!

justfile
jar-stripe: i think that is true re: reusing the build when packing (or previously when computing coverage).  replacing `dotnet build src -c Release /p:ContinuousIntegrationBuild=true` with `just ci-build` makes sense to me and we could then assume anything with a ci- scope has to conform to certain rules (e.g. ci-build first, and then everything else is nobuild).  but these are the short strokes, i think.

justfile
jar-stripe: ah, because the Makefile pulled it from an environment variable?  The change makes sense to me (and I prefer an argument to an env var anyway).

justfile
xavdid-stripe: let's leave it as written for now so we don't go down the rabbit hole of debugging the whole CI setup. I think these test commands are mostly set-and-forget, so i'm not worried about a _little_ harder to read

justfile
xavdid-stripe: exactly- just slightly different syntax. And yeah, we get real arguments now!

justfile
```diff
@@ -0,0 +1,36 @@
...
+
+# ⭐ run tests in debug mode
+test: (_test "" "-f net8.0" "Debug")
+
+# skip build and don't specify the dotnet framework
+ci-test: (_test "--no-build" "" "Release")
+
+# ⭐ format all files
+format *args:
+    TargetFramework=net5.0 dotnet format src/Stripe.net/Stripe.net.csproj --severity warn {{args}}
```
jar-stripe, line 22: I realize there wasn't a comment in the Makefile (and so this may be out of scope for this PR) but it would be nice to have a comment here explaining why we set TargetFramework=, something like:
> This sets TargetFramework because of a race condition in dotnet format when it tries to format to multiple targets at a time, which could lead to code with compiler errors after it completes

jar-stripe: Or if there's a better place to store that kind of domain knowledge, that works too

jar-stripe (approved):
This looks good to me!

PR #3051, opened 2025-02-04
Deprecate ServiceNested class

jar-stripe:
### Why?
The ServiceNested used to contain methods that mirrored functionality from the Service base class but for nested services. These methods have been removed and ServiceNested is now effectively empty. This PR marks it for future deletion.

When working on this issue, I discovered Service<T> is no longer needed as well, as all Service type parameters in use are defined at the method level.

### What?
- added Obsolete tag and message to ServiceNested class and Service<T> class

## Changelog
- Deprecates `ServiceNested` and `Service<T>` classes.  To implement custom services, use the request methods defined on `StripeClient`


jar-stripe:
Compat issues are expected, since we're changing the base class of all of the Service classes.

PR #3052, opened 2025-02-05
Remove RawRequestOptions#ApiMode 

jar-stripe:
### Why?
When we shipped RawRequest functionality in the 9/30 release, most LiveApiRequestor calls were implemented to determine the API mode from the path.  The exception to this was RawRequestAsync, which accepted ApiMode as an option.  This is not a configurable option (it is either V1 or V2 depending on the API) and inspecting the path is a reliable way to determine ApiMode; specifying the wrong ApiMode means that responses may be interpreted incorrectly.  Because of this, we are removing this option to avoid confusion and the potential for subtle errors.

### What?
- removed  RawRequestOptions@ApiMode argument
- changed LiveApiRequestor#RawRequestAsync to use GetApiMode to determine apiMode
- updated tests

## Changelog
* Removes `ApiMode` from `RawRequestOptions`.  ApiMode is automatically determined from the path passed to RawRequest.

### DIFF
--- a/src/Stripe.net/Services/_common/RawRequestOptions.cs
+++ b/src/Stripe.net/Services/_common/RawRequestOptions.cs
@@ -1,10 +1,12 @@
 namespace Stripe
 {
+    using System;
     using System.Collections.Generic;
 
     public class RawRequestOptions : RequestOptions
     {
         /// <summary>Gets or sets the API mode to use for the request.</summary>
+        [Obsolete("This property is no longer used and will be removed in a future release")]
         public ApiMode ApiMode { get; set; }
 
         /// <summary>Gets or sets additional headers for the request.</summary>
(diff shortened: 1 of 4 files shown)


helenye-stripe (approved):

src/Stripe.net/Services/_common/RawRequestOptions.cs
```diff
@@ -1,12 +1,10 @@
 namespace Stripe
 {
+    using System;
```
ramya-stripe, line 3: There is no net new code in this file... why this new import?

helenye-stripe (approved):

src/Stripe.net/Services/_common/RawRequestOptions.cs
jar-stripe: Good catch!  That was a carry-over from when I marked the field as obsolete.  Removed

PR #3054, opened 2025-02-11
Improved examples

jar-stripe:
### Why?
The Examples folder contains an example project to make it easy to run small code snippets against the local SDK codebase.  We are improving these examples to make them easy to use and run in order to reduce friction when testing SDK functionality

### What?
- Implemented example resolution in Program.cs, so one can pass an example name on the command line
- updated ExampleTemplate to list customers as an example, and pull api key from the environment
- updated V2 meter event stream example to pull configuration from environment
- added justfile recipe for running examples


xavdid-stripe (approved):
nice! I like the just -> example script

PR #3058, opened 2025-03-10
Support for StripeAccount in StripeClient and StripeContext in StripeClientOptions

jar-stripe:
### Why?
The SDK should support setting StripeAccount and StripeContext at the client level instead of requiring it to be passed into each request.

### What?
- added StripeAccount and StripeContext to StripeClientOptions
- added code to RequestOptionsExtensions WithClientOptions to copy over StripeContext and StripeAccount if it is not set in the request options
- added tests


jar-stripe:
Compat failure is expected; we changed the public API a few PRs ago and so every pr merged until we make a release that includes that change will have the same failure.  For this month, that's ok because 2025-03-31 is a major release.

mbroshi-stripe (approved):

PR #3071, opened 2025-03-26
Remove fields and tests for removed endpoints and fields

helenye-stripe:
### Why?

Remove various fields and tests for endpoints that are being removed in the upcoming major version.

The fields are for manually maintained files (List params for InvoiceLineItem) and we should do an audit afterwards to make sure all relevant manually maintained options/fields/resources are removed as well.

### What?

Remove fields/tests for various billing changes, Invoice.Upcoming, UsageRecord, and UsageRecordSummary.

I verified the fields that were being removed in the API release changes list.


jar-stripe (approved):
LG

Issue #2227, opened 2020-10-15
ReportTypeService List fails if given parameters


clement911:
When using `ReportTypeService.ListAsync`, passing a `Limit` parameter fails.
Indeed, the docs documents 0 parameter.

clement911:
I see.
I'd much from that you keep generic objects such as `IListable` and `ListOptions` because we use them heavily to write generic code that works with multiple services.
It's not really a big deal to be honest that the `ReportTypeService` fails there. I'll close.

remi-stripe:
I think we could still do this and just have `ListOptions` be empty really but keep it as an interface if that makes sense. Or I could push internally so that every List API supports pagination (which I thought it did so I think this one is special and should be fixed :p)

clement911:
An empty interface wouldn't work for us as we actually use the members of the interface...
So, personally, I'd prefer the second option to push every List API to support pagination.
But again, we can live with the current state of things...

mbroshi-stripe:
👋 Hi @clement911! This was a very old issue, but I just wanted to point out that we at last put in a permanent fix where `ReportTypeService` no longer extends `ListOptions` in the [v48.0.0 release](https://github.com/stripe/stripe-dotnet/releases/tag/v48.0.0).

PR #3078, opened 2025-03-28
Clean up ListOptions

mbroshi-stripe:
### Why?

The `ListOptionsWithCreated` base class is superfluous, so we are removing it. There are also a couple of List methods, namely `ReportType` and `AccountCapability` that do not actually auto-paginate, so those services should not implement `IListable`.

### What?

- Removes `ListOptionsWithCreated` base class and pushes the `created` field into the children.
- `ReportTypeOptions` and `AccountCapabilityOptions` no longer extend `ListOptions`
- `ReportTypeService` and `AccountCapabilityService` no longer implement the `IListable` interface, and the `ListAutoPaging` and `ListAutoPagingAsync` methods are removed

### See Also

https://github.com/stripe/stripe-dotnet/issues/2227

## Changelog
- Remove `ListOptionsWithCreated` base class and pushes the `created` field into the children.
- Classes `ReportTypeOptions` and `AccountCapabilityOptions` no longer extend `ListOptions`
- Classes`ReportTypeService` and `AccountCapabilityService` no longer implement the `IListable` interface, and the `ListAutoPaging` and `ListAutoPagingAsync` methods are removed

### DIFF
--- a/src/Stripe.net/Services/AccountCapabilities/AccountCapabilityListOptions.cs
+++ b/src/Stripe.net/Services/AccountCapabilities/AccountCapabilityListOptions.cs
@@ -1,7 +1,7 @@
 // File generated from our OpenAPI spec
 namespace Stripe
 {
-    public class AccountCapabilityListOptions : ListOptions
+    public class AccountCapabilityListOptions : BaseOptions
     {
     }
 }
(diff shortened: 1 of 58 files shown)


jar-stripe (approved):
LGTM

PR #3084, opened 2025-03-31
Removed unused manually maintained options class

jar-stripe:
### Why?
The `Upcoming` and `UpcomingAsync` methods in `InvoiceService` are removed in the upcoming major release.  This removes the manually maintained `ListOptions` class for those methods.

### What?
- removed `UpcomingInvoiceOptions` and related test code


jar-stripe:
Note: no changelog entry needed as the changelog already indicates the removed methods, and we don't add changelog entries for method param classes.

PR #3091, opened 2025-04-05
Xavdid/merge dotnet beta & fix tests

xavdid-stripe:
### Why?

This is fixing both a merge conflict and a failing test.

### What?

- `just merge dotnet-beta`
- add a skip in a test that doesn't work in beta SDKs anymore

PR #3093, opened 2025-04-08
copy fixed test from beta

xavdid-stripe:
### Why?

Backports a fix from https://github.com/stripe/stripe-dotnet/pull/3091

### What?

- skips a test when it's a preview SDK (which won't happen in GA, but it's nice for the code to be consistent)


jar-stripe (approved):
👍

PR #3106, opened 2025-04-17
Jar/merge dotnet beta

jar-stripe:

PR #3111, opened 2025-04-18
Require merge commit for merge PR (+ new rules.yml file)

jar-stripe:
### Why?
Our merge script requires that PRs be merged with a merge commit to maintain git history and avoid conflicts in the future.  Our normal PR process is to squash before merging, usually with auto-merge enabled.  Since Github defaults to the last merge method selected, its very easy to accidentally enable auto-squash in this one situation where we really need a merge commit.

Neither Github's branch rulesets nor branch protection rules let us define rules in the context of the HEAD ref, so this PR adds a CI that can.

### What?
- add rules.yml to define PR/branch rules that we cannot define in the Github UI
- added Merge script PRs must create merge commits rule to run when auto merge is enabled and when the branch contains `/merge-`, and check that the auto merge method is 'merge'

#### Demo
When auto-merge is enabled on a merge PR like https://github.com/stripe/stripe-dotnet/pull/3106, it runs the rule as part of its checks.  If the auto-merge method is not 'merge', it fails:
image

The error provides a brief description about why it failed as a reminder to go fix the auto-merge method:
image

Once fixed, the merge PR check is green, and everything's good to go:
image

Note that this must exist in the repo and enabled as a blocking status check to get the full benefit!


jar-stripe:
Dont approve this just yet, please!

xavdid-stripe (approved):

PR #3115, opened 2025-04-25
Use json files instead of stripe mock

mbroshi-stripe:
### Why?

When moving away from stripe-mock's builtin fixtures to the fixtures that ship with the same version of openapi used by this repo, we found that some of the tests were failing. They were trying to get fixtures from stripe-mock that are in private beta, so not available.

### What?

- Adds hard-coded fixture JSON files to replace the calls to stripe-mock

### See Also

TEST USING LATEST OPENAPI


helenye-stripe (approved):

prathmesh-stripe (approved):

prathmesh-stripe:
LGTM but build tests are still failing.

PR #3122, opened 2025-04-30
Update generated code for beta

helenye-stripe:
Codegen for openapi v1719.

## Changelog
Release specs are identical.


src/Stripe.net/Services/Privacy/RedactionJobValidationErrors/RedactionJobValidationErrorService.cs
```diff
@@ -26,22 +25,6 @@ public RedactionJobValidationErrorService(IStripeClient client)
         {
         }
 
-        /// <summary>
-        /// <p>Retrieve validation error method</p>.
-        /// </summary>
```
jar-stripe, line 31: just to confirm, is this not showing up in the diff because it was added since the last release and then removed before this release?

jar-stripe (approved):
LGTM, small question about the chanagelog

src/Stripe.net/Services/Privacy/RedactionJobValidationErrors/RedactionJobValidationErrorService.cs
helenye-stripe: It was added this release but they want to remove it. It should never be used by the public

PR #3137, opened 2025-06-08
Updated stripeclient snippets in Readme.md

prathmesh-stripe:
### Why?
We are soon releasing v1 namespaces under StripeClient. 

### What?
- Updated all snippets to include the new v1 namespace.

### See Also
https://jira.corp.stripe.com/browse/DEVSDK-2506

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,8 @@ var options = new CustomerCreateOptions
     Email = "customer@example.com"
 };
 
-var service = new CustomerService();
+var client = new StripeClient("sk_test_...");
+var service = client.V1.Customers;
 Customer customer = service.Create(options);
 
 // Newly created customer is returned
@@ -75,7 +76,8 @@ Console.WriteLine(customer.Email);
 The `Retrieve` method of the service class can be used to retrieve a resource:
 
 ```C#
-var service = new CustomerService();
+var client = new StripeClient("sk_test_...");
+var service = client.V1.Customers;
 Customer customer = service.Get("cus_1234");
 
 Console.WriteLine(customer.Email);
@@ -91,7 +93,8 @@ var options = new CustomerUpdateOptions
     Email = "updated-email@example.com"
 };
 
-var service = new CustomerService();
+var client = new StripeClient("sk_test_...");
+var service = client.V1.Customers;
 Customer customer = service.Update("cus_123", options);
 
 // The updated customer is returned
@@ -103,7 +106,8 @@ Console.WriteLine(customer.Email);
 The `Delete` method of the service class can be used to delete a resource:
 
 ```C#
-var service = new CustomerService();
+var client = new StripeClient("sk_test_...");
+var service = client.V1.Customers;
 Customer customer = service.Delete("cus_123", options);
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,7 +63,8 @@ var options = new CustomerCreateOptions
     Email = "customer@example.com"
 };
 
-var service = new CustomerService();
+var client = new StripeClient("sk_test_...");
+var service = client.V1.Customers;
```
helenye-stripe: do we not just want `client.V1.Customers.Create()`?

prathmesh-stripe: This is to directly replicate what a user will see in API reference to avoid confusion when jumping from readme to apiref. 
One liner is simpler but it would require a refactor of .NET examples.

helenye-stripe: Hm, ok. Do we have that on our radar?

prathmesh-stripe: Not planned atm. Might be easier to do when we drop support for resources based snippets.

ramya-stripe: Post offline discussions:
- Lets move to using `client.V1.Customers.Create()` here rather than create an intermediate "service"
- Lets create a ticket to do the same for generated code snippets which we can come back to in July

ramya-stripe: Add a new section "Using StripeClient" and talk about the benefits while maintaining that the old way of instantiating services with global config will still work

Advantages to call out
- No global config - you can create multiple clients with different configuration
- A single root entry point from where you can discover all available services

README.md
```diff
@@ -43,14 +43,32 @@ how to use the library.
 
 ## Usage
 
+### Using StripeClient
```
prathmesh-stripe, line 46: @ramya-stripe Can you please take a look at this?

README.md
```diff
@@ -43,14 +43,32 @@ how to use the library.
...
+
+Stripe .NET SDK offers a more flexible approach of working with the Stripe API through the `StripeClient` class. 
+StripeClient pattern has several advantages over the traditional approach of instantiating services with a global configuration: 
+- Client based configuration: Allows you to instantiate several clients with different configurations(API key, network retries etc.). 
+- Single Root Entry Point: The StripeClient provides a centralized entry point to all available Stripe services. This improves discoverability through IDE auto-completion and creates a more intuitive developer experience for you.
```
ramya-stripe:
```suggestion
In version 46 of the Stripe .NET SDK, we have enhanced the `StripeClient` class to be the entry point to access all services that had to be previously independently instantiated with global configuration. This improves discoverability during IDE auto-completion and creates a more intuitive developer experience for you. 

Each client instantiation can have its own configuration so you can access Stripe API with different API keys or different configuration (like number of retries) on a per client basis and without changing a global configuration.
```

ramya-stripe: @jar-stripe Does the above read right to you?

jar-stripe: Content wise it looks good; I might suggest something like this as an alternate wording for the second paragraph:

> Each client instantiation can have its own configuration so you can access Stripe API with different API keys or different configuration (like number of retries) on a per client basis and without changing a global configuration.

ramya-stripe: done

README.md
```diff
@@ -43,14 +43,31 @@ how to use the library.
...
+var client = new StripeClient("sk_test_...");
+Customer customer = client.V1.Customers.Get("cus_1234");
+
+// Global Configuration pattern (Legacy)
 StripeConfiguration.ApiKey = "sk_test_...";
+var service = new CustomerService();
+Customer customer = service.Get("cus_1234");
+```
+
+The traditional Global Configuration pattern will continue to work for everyone.
```
ramya-stripe:
```suggestion
The previous global configuration pattern will continue to be supported.
```

PR #3150, opened 2025-07-21
Separate the default StripeConfiguration.StripeClient from feature-full StripeClient class, and add telemetry string to StripeClient

jar-stripe:
### Why?
Stripe.NET supports two different service usage patterns: direct Service class construction, and accessing services via StripeClient.  We want to be able to differentiate these patterns in our client telemetry.

### What?
- adds defaultUsage construction parameter to LiveApiRequestor to let callers specify telemetry usage strings for each request
- adds telemetry usage strings to LiveApiRequestor constructor in StripeClient
- add DefaultStripeClient IStripeClient implementation (without a usage string)
- modifies StripeConfiguration.BuildDefaultStripeClient to create a DefaultStripeClient
- adds tests to verify correct telemetry usage strings

## Changelog
* Adds usage string to telemetry on API calls made through `StripeClient` service accessors

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/ApiRequestorAdapter.cs
+++ b/src/Stripe.net/Infrastructure/Public/ApiRequestorAdapter.cs
@@ -35,6 +35,11 @@ namespace Stripe
                 return stripeClient.Requestor;
             }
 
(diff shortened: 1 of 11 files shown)


helenye-stripe (approved):
👍

PR #3167, opened 2025-09-05
⚠️ Build SDK w/ V2 OpenAPI spec

xavdid-stripe:
### Why?

While SDK support for our v1 APIs comes from our [public openapi spec](https://github.com/stripe/openapi), our v2 endpoints have been sources from an internal dump of the protobuf definitions of those APIs. 

This PR marks our cutover point for our generation source: now we're using the (private, for now) V2 OpenAPI spec! There are a few changes/bugfixes as a result (documented below).

> note: entire PR is codegen

### What?

- generated SDK using v2 openapi spec.

### See Also

[DEVSDK-2750](https://go/j/DEVSDK-2750)

## Changelog

- ⚠️ V2 delete methods now return a `V2DeletedObject`


src/Stripe.net/Entities/V2/DeletedObjects/DeletedObject.cs
jar-stripe, line 1: small nit but the pluralized folder is a little weird

jar-stripe (approved):
Other than the DeletedObjects folder, it looks good!

PR #3168, opened 2025-09-07
⚠️ Add strongly typed EventNotifications

xavdid-stripe:
### Why?

As part of the Next-Gen Event Handling project, we're adding fully typed classes to represent all the ~~thin events~~ Event Notifications a user could receive. There are also helper methods that help with the event handling experience.

### What?

- (see changelog below for user-facing changes)
- add tests
- add justfile recipe for `test-one`
- generate event notification classes
- added and internal `GetConcreteType` to the type registry (though we probably don't need it; see note)

### See Also

[DEVSDK-2662](https://go/j/browse/DEVSDK-2662)

## Changelog

We've overhauled how V2 Events are handled in the SDK! This approach should provide a lot more information at authoring and compile time, leading to more robust integrations. As part of this process, there are a number of changes to be aware of.

- Added matching `EventNotification` classes to every v2 `Event`. For example, there's now a `V1BillingMeterErrorReportTriggeredEventNotification` to match the existing `V1BillingMeterErrorReportTriggeredEvent`. Each notification class defines a `fetchEvent()` method to retrieve its corresponding event. For events with related objects, there's a `fetchRelatedObject()` method that performs the API call and casts the response to the correct type.
- ⚠️ Rename function `StripeClient.parseThinEvent` to `StripeClient.parseEventNotification` and remove the `Stripe.ThinEvent` class.
    - This function now returns a `Stripe.V2.Core.EventNotification` (which is the shared base class that all of the more specific `Stripe.Events.*EventNotification` classes  share) instead of `Stripe.ThinEvent`. When applicable, these event notifications will have the `relatedObject` property and a `fetchRelatedObject()` function. They also have a `fetchEvent()` method to retrieve their corresponding `Stripe.Event.*Event` instance.
    - If you parse an event the SDK doesn't have types for (e.g. it's newer than the SDK you're using), you'll get an instance of `Stripe.Events.UnknownEventNotification` instead of a more specific type. It has both the `relatedObject` property and the `FetchRelatedObject()` function (but they may be/return `null`)

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/EventNotification.cs
+++ b/src/Stripe.net/Infrastructure/Public/EventNotification.cs
@@ -78,7 +78,10 @@ namespace Stripe.V2
         protected StripeClient? Client { get; set; }
 
         /// <summary>
-        /// Not suitable for production code because it doesn't validate event signature. It's used internally for testing and from StripeClient.
+        /// Helper for constructing an Event Notification. Doesn't perform signature validation, so you
+        /// should use <see cref="StripeClient.ParseEventNotification"/> instead for
+        /// initial handling. This is useful in unit tests and working with EventNotifications that you've
+        /// already validated the authenticity of.
         /// </summary>
         /// <param name="payload">JSON body for an EventNotification.</param>
         /// <param name="client">A StripeClient instance that will be used to make requests.</param>


src/Stripe.net/Entities/V2/Events/Event.partial.cs
```diff
@@ -66,7 +66,7 @@ protected virtual Task<T> FetchRelatedObjectAsync<T>(EventRelatedObject relatedO
             RequestOptions opts = null;
             if (this.Context != null)
             {
-                opts = new RequestOptions { StripeAccount = this.Context };
```
jar-stripe, line 69: is this considered a breaking change?  if so we need to add a changelog entry

src/Stripe.net/Infrastructure/JsonConverters/STJV2EventNotificationConverter.cs
```diff
@@ -0,0 +1,41 @@
+#if NET6_0_OR_GREATER
+namespace Stripe.Infrastructure
```
jar-stripe, line 2: is this class needed?  it doesn't look like it's doing anything here

src/StripeTests/Events/V2/EventTest.cs
```diff
@@ -175,89 +178,115 @@ private Stripe.V2.Event DoParseSignedEvent(string payload)
...
 
-            var bte = this.stripeClient.ParseThinEvent(payload, GenerateSigHeader(payload), WebhookSecret);
+            var bte = this.DoParseSignedEventNotification(payload);
+
+            // fetch full event
+            return bte.FetchEvent<Stripe.V2.Event>();
+        }
+
+        /// <summary>
+        /// Parse a ThinEvent object.  This uses StripeClient.ParseThinEvent to
```
jar-stripe: error in comment

jar-stripe: also in the `<returns>`

src/StripeTests/Events/V2/EventTest.cs
```diff
@@ -175,89 +178,115 @@ private Stripe.V2.Event DoParseSignedEvent(string payload)
...
+        /// <param name="o">Your object.</param>
+        /// <returns>The input object, cast to type T.</returns>
+        private T AssertAndCast<T>(object o)
+        {
+            Assert.IsType<T>(o);
+            return (T)o;
         }
 
         [Fact]
         public void ParseThinEventWithoutRelatedObject()
```
jar-stripe: name should be updated

src/Stripe.net/Infrastructure/Public/UnknownEventNotification.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            return this.FetchEvent<V2.Event>();
+        }
+
+        public Task<V2.Event> FetchEventAsync(CancellationToken cancellationToken = default)
+        {
+            return this.FetchEventAsync<V2.Event>(cancellationToken);
+        }
+
+        public T FetchRelatedObject<T>()
```
jar-stripe: this is interesting; is the plan to allow custom objects to be specified (as long as they implement IStripeEntity)?

src/Stripe.net/Entities/V2/Events/Event.partial.cs
xavdid-stripe: It's got no user impact; for now, we treat the account and context headers identically. this is just marginally more correct / future proof.

src/Stripe.net/Infrastructure/JsonConverters/STJV2EventNotificationConverter.cs
xavdid-stripe: I'm not sure, I was just coping the matching `EventConverter` 😅 

The only notification specific thing is `CanConvert`, which references `Stripe.V2.EventNotification`. If STJ serialization works works without this class, happy to ditch it.

src/Stripe.net/Infrastructure/JsonConverters/V2EventConverter.cs
```diff
@@ -18,6 +18,11 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
             throw new NotSupportedException("V2EventConverter should only be used while deserializing.");
         }
 
+        protected virtual Type GetConcreteType(string typeValue)
```
xavdid-stripe: I can probably back out this whole file - I did this when I thought I could subclass this for the event notification, but that didn't work the way I wanted.

src/Stripe.net/Infrastructure/Public/UnknownEventNotification.cs
xavdid-stripe: Not custom so much as unknown - if there's a `AccountExploded` event that the SDK doesn't know about, we want users to be able to fetch the related object (which they know is an `Account` even if the SDK doesn't

src/Stripe.net/Infrastructure/JsonConverters/V2EventNotificationConverter.cs
```diff
@@ -0,0 +1,53 @@
+namespace Stripe.Infrastructure
+{
+    using System;
+    using System.Reflection;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+
+    /// <summary>
+    /// This converter deserializes Stripe thin events, which are polymorphic and discriminated by the value
```
jar-stripe: thin events should be changed to event notifications

src/Stripe.net/Infrastructure/Public/EventNotification.cs
```diff
@@ -0,0 +1,148 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Newtonsoft.Json;
+    using Stripe.Infrastructure;
+#if NET6_0_OR_GREATER
+    using STJS = System.Text.Json.Serialization;
+#endif
+
+    /// <summary>
+    /// A pushed thin event.  Use this the Id with the <see cref="Stripe.V2.Core.EventService"/>
```
jar-stripe: "thin event"

src/Stripe.net/Infrastructure/JsonConverters/STJV2EventNotificationConverter.cs
jar-stripe: i think its needed for now but i also think there's a better way to implement this type of converter that ill have to look into.  for now though this is good!

src/Stripe.net/Infrastructure/JsonConverters/V2EventConverter.cs
jar-stripe: fewer changes the better!

src/Stripe.net/Infrastructure/Public/UnknownEventNotification.cs
jar-stripe: In that situation, I'm not sure I like this over requiring the user to go to the necessary service, e.g. client.V1.Accounts.Get(...).  This feels like it could be easy to accidentally mess up; like what happens if the caller supplied T is wrong, where maybe the user specifies the wrong Account type?

xavdid-stripe: That's fair. the solution in python is to return a `StripeObject` because we know it's some API resource, but don't know what it is exactly. I'll see if that's possible here.

xavdid-stripe: looks like it is! good call, that's much better.

xavdid-stripe:
(ignore red CI; docker hub is having a problem so all tests are down. Everything besides compat was good before i merged master back into this and trigger the CI run.)

src/Stripe.net/Infrastructure/Public/UnknownEventNotification.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            return this.FetchEvent<V2.Event>();
+        }
+
+        public Task<V2.Event> FetchEventAsync(CancellationToken cancellationToken = default)
+        {
+            return this.FetchEventAsync<V2.Event>(cancellationToken);
+        }
+
+        public IStripeEntity FetchRelatedObject()
```
jar-stripe: if we can we should use StripeEntity here (sorry I missed this before). The interface is bare bones; the abstract class has e.g. RawJObject and various other helper methods

jar-stripe (commented):
One more change; sorry I missed this earlier!

jar-stripe (approved):

PR #3170, opened 2025-09-10
Adds public BaseUrl to RawRequestOptions

jar-stripe:
### Why?
[RawRequest](https://github.com/stripe/stripe-dotnet?tab=readme-ov-file#custom-requests) can be used to make custom requests to Stripe APIs not yet supported by the SDK, but it was limited to APIs exposed via api.stripe.com.  This limits the ability to call some of our APs.

### What?
- renames internal `BaseUrl` in RequestOptions to `InternalBaseUrl`
- adds public BaseUrl to RawRequestOptions which delegates to InternalBaseUrl for safety
- adds test


helenye-stripe:
lg but can we get a changelog entry?

PR #3194, opened 2025-09-25
Add `StripeContext` object

xavdid-stripe:
### Why?

As the `stripe-context` header is evolving, we want to give better tools for creating and managing contexts. This PR adds a new class, `StripeContext`. It can be used anywhere the context option is supplied and gets serialized to a string when making requests. It's also found on the `EventNotification.context` property.

### What?

- add `StripeContext` class
- allow it to be supplied for RequestOptions.context; serialize it into strings
- add tests

## Changelog

- Add the `StripeContext` class
- ⚠️ Change `EventNotification` (formerly known as `ThinEvent`)'s `context` property from `string` to `StripeContext`

### DIFF
--- a/src/Stripe.net/Services/_common/RequestOptions.cs
+++ b/src/Stripe.net/Services/_common/RequestOptions.cs
@@ -71,10 +71,8 @@ namespace Stripe
                 clone.ApiKey = clientOptions.ApiKey;
             }
 
-            if (string.IsNullOrEmpty(clone.StripeContext))
-            {
-                clone.StripeContext = clientOptions.StripeContext;
-            }
+            // an empty string in the clone means we should _not_ overwrite here
+            clone.StripeContext ??= clientOptions.StripeContext;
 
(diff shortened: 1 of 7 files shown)


src/Stripe.net/Services/_common/RequestOptions.cs
```diff
@@ -71,10 +71,8 @@ internal static T WithClientOptions<T>(this T requestOptions, StripeClientOption
                 clone.ApiKey = clientOptions.ApiKey;
             }
 
-            if (string.IsNullOrEmpty(clone.StripeContext))
-            {
-                clone.StripeContext = clientOptions.StripeContext;
-            }
+            // an empty string in the clone means we should _not_ overwrite here
+            clone.StripeContext ??= clientOptions.StripeContext;
```
jar-stripe, line 75: small nit but it would be nice to keep this stylistically the same as the others

PR #3202, opened 2025-09-29
Speed up just format in CI for private-preview branch

jar-stripe:
### Why?
Recently we noticed that the format stage of updating generated code for testing was taking > 30 min on the private preview build.  This PR updates the `just format` command to split out whitespace formatting in a more efficient manner, which also speeds up the regular formatting (run immediately after).

### What?
- separated formatting whitespace from other formatting, for performance

### DIFF
--- a/justfile
+++ b/justfile
@@ -22,6 +22,7 @@ ci-test: (_test "--no-build" "" "Release")
 
 # ⭐ format all files
 format *args:
+    dotnet format whitespace --folder {{args}}
     # This sets TargetFramework because of a race condition in dotnet format when it tries to format to multiple targets at a time, which could lead to code with compiler errors after it completes
     TargetFramework=net5.0 dotnet format src/Stripe.net.sln --severity warn {{args}}
 


prathmesh-stripe (approved):

justfile
```diff
@@ -22,6 +22,7 @@ ci-test: (_test "--no-build" "" "Release")
 
 # ⭐ format all files
 format *args:
+    dotnet format whitespace --folder {{args}}
```
xavdid-stripe, line 25: does this work when `{{ args }}` is `""`? I would expect an error about "no value provided for flag `--folder`"

xavdid-stripe: (if so, you can copy the approach above in `_test`)

jar-stripe: `dotnet format whitespace --folder` is the actual command; the real question is does this work with just format-check.  lemme confirm that

jar-stripe: `just format-check` runs correctly

PR #3222, opened 2025-10-14
Add support for building on .NET 9

jar-stripe:
### Why?
As per https://docs.stripe.com/sdks/versioning?lang=dotnet#language-policies, we are supporting .NET 9 in Stripe.net.  This PR migrates our builds and tests to use .NET 9.  Additionally, to support our updated version policy, we needed to refactor our github workflow files to make it easier to update our supported versions.

### What?
- moves list of supported versions out of ci.yml and adds step to set supported versions and latest supported version (for use in downstream jobs)
- updates test project TargetFrameworks to include net9.0 and `just test` to use net9.0 target
- adds conditional compilation flag to `SystemNetHttpClient` to not use deprecated pattern in .net 9.0
- adds `just install` and `just print-supported-versions` and `just print-latest-supported-version` commands and `dotnet-install.sh` script to help with set up and with reporting supported versions
- updates `just test` and `just test-one` to always test against the latest supported version
- removes global.json; instead, dotnet will use the latest version installed in the `just install` command

## Changelog
* Adds support for building Stripe.net on .NET 9.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         id: supported-dotnet-versions
         run: |
           echo "dotnet_versions<<EOF" >> $GITHUB_OUTPUT
-          just print_supported_dotnet_versions
+          just print_supported_dotnet_versions >> $GITHUB_OUTPUT
           echo "EOF" >> $GITHUB_OUTPUT
       - name: Get latest .NET version
         id: latest-dotnet-version


scripts/dotnet-install.sh
```diff
@@ -0,0 +1,1892 @@
+#!/usr/bin/env bash
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+# Downloaded from https://learn.microsoft.com/en-us/dotnet/core/install/linux-scripted-manual#scripted-install
```
mbroshi-stripe, line 6: I'm fine this this, but OOC, is there a "packaged" or versioned script we can pull in?

justfile
```diff
@@ -5,17 +5,36 @@ import? '../sdk-codegen/utils.just'
...
+  then
+    echo "Installing dotnet via mise..."
+    mise install dotnet
+  fi
+
+  export -p DOTNET_ROOT="$(mise where dotnet)"
+
+  echo "Ensuring all required versions are installed"
+  just install-sdk 9.0
+  just install-sdk 8.0
```
mbroshi-stripe: Why are we installing just the 2 most recent?

jar-stripe: thats a good question!  honestly, it probably doesn't need to even be 2.  generally speaking, a .net version can build at its version or versions below it, and you only need the specific version to run a project built for that version.  i think i may refactor this PR a bit to make it make a bit more sense.

scripts/dotnet-install.sh
jar-stripe: There isn't.  This is the blessed way to install w/o using an interactive installer (as provided by microsoft at the mentioned url).  IMHO the right way to do this is to enhance the mise extension for dotnet to place installed versions correctly in its install directory but i think thats a bigger undertaking.

justfile
jar-stripe: Ok, I changed the strategy a bit: i moved the supported version calculation into just and set it up so that tests always run the latest version (as computed); and then replaced the scripts in ci.yml with calls to just to move logic out of the yml file.  Can you give it a once over and see if this makes more sense?

scripts/dotnet-install.sh
mbroshi-stripe: SGTM!

mbroshi-stripe (approved):

PR #3223, opened 2025-10-14
Fixes STJDefaultConverter to safely ignore unknown properties

jar-stripe:
### Why?
There is a bug in STJDefaultConverter where extra properties in the JSON could cause an exception when it is being deserialized.  We currently do not support capturing extra properties into a separate object with STJ (like we do with Newtonsoft Json.NET) so for now we can safely discard these properties.

### What?
- fixes STJDefaultConverter to advance the reader even if the property is not present in the type being deserialized
- changes loop to find the target property to a Filter call, for readability

## Changelog
* Fixes a bug when using System.Text.Json to deserialize JSON that has properties not present in the target object.


prathmesh-stripe (approved):
I'll follow this up with a test case.

PR #3224, opened 2025-10-15
Added STJ tests for deserializing null values

prathmesh-stripe:
### Why?
Add some tests as followup to https://github.com/stripe/stripe-dotnet/pull/3223

### What?
- Add test to serialize and deserialize basic object with System.Text.Json
- Add serialization+deserialization test for Newtonsoft that has null fields assigned to non-nullable fields.

### DIFF
--- a/src/StripeTests/Infrastructure/SerializationTest.cs
+++ b/src/StripeTests/Infrastructure/SerializationTest.cs
@@ -44,6 +44,14 @@ namespace StripeTests
                 JsonConvert.DeserializeObject<Event>(serialized),
                 Formatting.Indented);
             Assert.Equal(serialized, reserialized);
+
+            json = GetResourceAsString("api_fixtures.customer_with_expansions.json");
+            var customer = JsonConvert.DeserializeObject<Customer>(json);
+            serialized = JsonConvert.SerializeObject(customer, Formatting.Indented);
+            reserialized = JsonConvert.SerializeObject(
+                JsonConvert.DeserializeObject<Customer>(serialized),
+                Formatting.Indented);
+            Assert.Equal(serialized, reserialized);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/StripeTests/Infrastructure/SerializationTest.cs
```diff
@@ -44,6 +44,14 @@ public void Serialize()
                 JsonConvert.DeserializeObject<Event>(serialized),
                 Formatting.Indented);
             Assert.Equal(serialized, reserialized);
+
+            json = GetResourceAsString("api_fixtures.customer_with_expansions.json");
```
prathmesh-stripe, line 48: `description` is nullable in the JSON, required primitive in Customer class. This is handled gracefully by Newtonsoft. This same test does not pass with SystemTextJson.

jar-stripe: this confuses me because strings should work like objects, not other primatives.  do we know why this is failing?

src/StripeTests/Infrastructure/StjDeserializationTest.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                Date = roundedDate,
+            };
+
+            var reloaded = JsonSerializer.Deserialize<TestObjectDateTime>(JsonSerializer.Serialize(obj));
+            Assert.Equal(reloaded.Date, obj.Date);
+        }
+
+        [Fact]
+        public void HandleNull()
```
jar-stripe, line 28: i think we should move this into https://github.com/stripe/stripe-dotnet/pull/3158/files

src/StripeTests/Infrastructure/SerializationTest.cs
prathmesh-stripe: This is passing with Newtonsoft. Skipped it in STJ test

src/StripeTests/Infrastructure/StjDeserializationTest.cs
prathmesh-stripe: But this is a passing test so just wanted to get this merged.

PR #3250, opened 2025-11-11
Add EventNotificationHandler

xavdid-stripe:
### Why?

We've been designing a streamlined approach to handling incoming events that is easy to get right and hard to get wrong. This PR has the initial implementation of this new system.

The only other pending item is to add a method to allow handling a webhook without verifying the signature. This is good for testing and for Event Bridge, which doesn't use the signature-based verification. Otherwise, this is ready for review.

### What?

- add `EventHandler` class
- add event handler constructor on `StripeClient`
- add tests
- add method for changing the stripe context on a client

### Example usage

```cs
[Route("api/[controller]")]
[ApiController]
public class EventNotificationWebhookHandler : ControllerBase
{
    private readonly StripeClient client;
    private readonly StripeEventNotificationHandler handler;

    public EventNotificationWebhookHandler()
    {
        client = new StripeClient(Environment.GetEnvironmentVariable("STRIPE_API_KEY"));
        handler = client.NotificationHandler(Environment.GetEnvironmentVariable("WEBHOOK_SECRET") ?? string.Empty, FallbackCallback);

        // register handlers
        handler.V1BillingMeterErrorReportTriggered += HandleBillingMeterErrorReportTriggeredEventNotification;
    }

    private void HandleBillingMeterErrorReportTriggeredEventNotification(object sender, Stripe.StripeEventNotificationEventArgs<Stripe.Events.V1BillingMeterErrorReportTriggeredEventNotification> e)
    {
        var meter = e.EventNotification.FetchRelatedObject();
        Console.WriteLine($"Meter {meter.DisplayName} had an error");
    }

    private void FallbackCallback(object sender, Stripe.StripeUnhandledEventNotificationEventArgs e)
    {
        Console.WriteLine($"Received unhandled event notification type: {e.EventNotification.Type}");
    }

    [HttpPost]
    public async Task<IActionResult> Index()
    {
        var json = await new StreamReader(HttpContext.Request.Body).ReadToEndAsync();
        handler.Handle(json, Request.Headers["Stripe-Signature"]);
        return null;
    }
}
```

### See Also

- [DEVSDK-2822](https://go/j/DEVSDK-2822)
- https://github.com/stripe/stripe-dotnet/pull/3168

### DIFF
--- a/src/StripeTests/Infrastructure/Public/StripeEventNotificationHandlerTest.cs
+++ b/src/StripeTests/Infrastructure/Public/StripeEventNotificationHandlerTest.cs
@@ -74,7 +74,7 @@ namespace StripeTests
             }
 
             var handler = new StripeEventNotificationHandler(this.stripeClient, WebhookSecret, UnhandledHandler);
-            handler.V1BillingMeterErrorReportTriggeredEventNotification += Handler;
+            handler.V1BillingMeterErrorReportTriggered += Handler;
 
             var sigHeader = StripeTests.V2.EventTest.GenerateSigHeader(this.V1BillingMeterPayload);
             handler.Handle(this.V1BillingMeterPayload, sigHeader);
@@ -115,8 +115,8 @@ namespace StripeTests
             }
 
             var handler = new StripeEventNotificationHandler(this.stripeClient, WebhookSecret, UnhandledHandler);
-            handler.V1BillingMeterErrorReportTriggeredEventNotification += BillingHandler;
-            handler.V1BillingMeterNoMeterFoundEventNotification += NoMeterHandler;
+            handler.V1BillingMeterErrorReportTriggered += BillingHandler;
+            handler.V1BillingMeterNoMeterFound += NoMeterHandler;
 
             var sigHeader1 = StripeTests.V2.EventTest.GenerateSigHeader(this.V1BillingMeterPayload);
             handler.Handle(this.V1BillingMeterPayload, sigHeader1);
@@ -150,7 +150,7 @@ namespace StripeTests
             }
 
             var handler = new StripeEventNotificationHandler(this.stripeClient, WebhookSecret, (s, e) => { });
-            handler.V1BillingMeterErrorReportTriggeredEventNotification += Handler;
+            handler.V1BillingMeterErrorReportTriggered += Handler;
 
             var sigHeader = StripeTests.V2.EventTest.GenerateSigHeader(this.V1BillingMeterPayload);
             handler.Handle(this.V1BillingMeterPayload, sigHeader);
@@ -176,11 +176,11 @@ namespace StripeTests
             }
 
             var handler = new StripeEventNotificationHandler(this.stripeClient, WebhookSecret, (s, e) => { });
-            handler.V1BillingMeterErrorReportTriggeredEventNotification += Handler1;
+            handler.V1BillingMeterErrorReportTriggered += Handler1;
 
             var exception = Assert.Throws<InvalidOperationException>(() =>
             {
-                handler.V1BillingMeterErrorReportTriggeredEventNotification += Handler2;
+                handler.V1BillingMeterErrorReportTriggered += Handler2;
             });
 
             Assert.Contains("already registered", exception.Message);
@@ -194,11 +194,11 @@ namespace StripeTests
             }
 
             var handler = new StripeEventNotificationHandler(this.stripeClient, WebhookSecret, (s, e) => { });
-            handler.V1BillingMeterErrorReportTriggeredEventNotification += Handler;
+            handler.V1BillingMeterErrorReportTriggered += Handler;
 
             var exception = Assert.Throws<InvalidOperationException>(() =>
             {
-                handler.V1BillingMeterErrorReportTriggeredEventNotification -= Handler;
+                handler.V1BillingMeterErrorReportTriggered -= Handler;
             });
 
             Assert.Contains("not supported", exception.Message);
@@ -235,7 +235,7 @@ namespace StripeTests
             }
 
             var handler = new StripeEventNotificationHandler(this.stripeClient, WebhookSecret, (s, e) => { });
-            handler.V1BillingMeterErrorReportTriggeredEventNotification += Handler;
+            handler.V1BillingMeterErrorReportTriggered += Handler;
 
             var types = handler.HandledEventTypes();
 
@@ -253,9 +253,9 @@ namespace StripeTests
             var handler = new StripeEventNotificationHandler(this.stripeClient, WebhookSecret, (s, e) => { });
 
             // Register in non-alphabetical order
-            handler.V2CoreAccountUpdatedEventNotification += Handler;
-            handler.V1BillingMeterErrorReportTriggeredEventNotification += Handler;
-            handler.V2CoreAccountCreatedEventNotification += Handler;
+            handler.V2CoreAccountUpdated += Handler;
+            handler.V1BillingMeterErrorReportTriggered += Handler;
+            handler.V2CoreAccountCreated += Handler;
 
             var types = handler.HandledEventTypes();
 
@@ -346,7 +346,7 @@ namespace StripeTests
             }
 
             var handler = new StripeEventNotificationHandler(this.stripeClient, WebhookSecret, UnhandledHandler);
-            handler.V1BillingMeterErrorReportTriggeredEventNotification += Handler;
+            handler.V1BillingMeterErrorReportTriggered += Handler;
 
             var sigHeader = StripeTests.V2.EventTest.GenerateSigHeader(this.V1BillingMeterPayload);
             handler.Handle(this.V1BillingMeterPayload, sigHeader);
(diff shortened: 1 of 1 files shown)


src/Stripe.net/Infrastructure/Public/StripeEventRouter.cs
```diff
@@ -0,0 +1,800 @@
...
+
+        // private-event-handlers: The end of the section generated from our OpenAPI spec
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="StripeEventRouter"/> class.
+        /// </summary>
+        /// <param name="client">The StripeClient instance to use for parsing and API requests.</param>
+        /// <param name="webhookSecret">The webhook secret used for signature verification.</param>
+        /// <param name="unhandledEventHandler">TODO: ADD UNHANDLED DETAILS.</param>
+        public StripeEventRouter(StripeClient client, string webhookSecret, EventHandler<StripeUnhandledEventNotificationEventArgs> unhandledEventHandler)
```
xavdid-stripe: Since `unhandledEventHandler` i'm not convinced it actually needs to be an event handler? but i'm not sure if that's what users expect anyway

justfile
```diff
@@ -29,7 +29,7 @@ _mise_install_dotnet:
   export DOTNET_ROOT="$(mise where dotnet)"
 
 _dotnet_install tpv:
-  ./scripts/dotnet-install.sh --channel {{ tpv }} --install-dir $DOTNET_ROOT
+  ./scripts/dotnet-install.sh --channel {{ tpv }} --install-dir "$(mise where dotnet)"
```
xavdid-stripe: Env isn't shared between recipes, so this doesn't work if you don't have the root set

jar-stripe:
Not sure if the comment should go here or in your doc, but HandleUnhandledError seems like the wrong name in your example.  That's just HandleUnhandledEventNotification, right?

src/Stripe.net/Infrastructure/Public/StripeEventNotificationEventArgs.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <typeparam name="TEventNotification">The type of EventNotification.</typeparam>
+    public class StripeEventNotificationEventArgs<TEventNotification> : EventArgs
+        where TEventNotification : V2.Core.EventNotification
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="StripeEventNotificationEventArgs{TEventNotification}"/> class.
+        /// </summary>
+        /// <param name="eventNotification">The event notification instance.</param>
+        /// <param name="client">The StripeClient instance.</param>
+        public StripeEventNotificationEventArgs(TEventNotification eventNotification, StripeClient client)
```
jar-stripe, line 18: will this ever get constructed outside of our SDK code?  If not, I would consider changing this constructor to be internal

src/Stripe.net/Infrastructure/Public/StripeEventNotificationEventArgs.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            this.EventNotification = eventNotification ?? throw new ArgumentNullException(nameof(eventNotification));
```
jar-stripe: its a small nit, but can you pull the null check and throw into a separate line/conditional?  I find that easier to read

src/Stripe.net/Infrastructure/Public/StripeEventRouter.cs
jar-stripe: I dont think I've seen any classes take an event handler as a constructor arg, but also I don't think I've seen too many patterns where an event is expected to have a handler when it is fired or else its an error.  I think the shape of this is fine, but if you wanted to make it a little more "this is just a function"-like, you could do:
```
public StripeEventRouter(StripeClient client, string webhookSecret, Action<object, StripeUnhandledEventNotificationEventArgs> unhandledEventHandler)
{
    this.client = client ?? throw new ArgumentNullException(nameof(client));
    this.webhookSecret = webhookSecret ?? throw new ArgumentNullException(nameof(webhookSecret));
    this.UnhandledEventHandler += new EventHandler<StripeUnhandledEventNotificationEventArgs>(unhandledEventHandler);
}
```

This has the added benefit of making it easy to pass anonymous functions in; but they also could do the `new EventHandler...` adapter call if you wanted to keep the argument the way it is.

Absent user feedback, I don't think one way is necessarily better than the other.

justfile
jar-stripe: we may want to fail if DOTNET_ROOT isnt set here; iirc a lot of stuff doesnt work right if that env var isn't already set.

src/Stripe.net/Infrastructure/Public/StripeEventRouter.cs
```diff
@@ -0,0 +1,800 @@
...
+            var requestor = this.client.Requestor as LiveApiRequestor;
+            if (requestor == null)
+            {
+                // If we can't cast to LiveApiRequestor, just dispatch without context override
+                this.DispatchEvent(eventNotification);
+                return;
+            }
+
+            // Save the original context and temporarily set the event's context
+            var originalContext = requestor.CurrentStripeContext;
```
jar-stripe: I don't think you can safely swap out and back the StripeContext in an environment with concurrent API calls, if this.client represents the StripeClient that created the router.  The EventRouter should hold its own StripeClient and requestor, if it is expected to make API calls with different context than e.g. a user's "normal" StripeClient instance.

src/Stripe.net/Infrastructure/Public/StripeEventNotificationEventArgs.cs
xavdid-stripe: Wouldn't users need it for tests? Like, if they write their function and want to test it in isolation, that function takes these args and they'd nee to be able to construct it?

src/Stripe.net/Infrastructure/Public/StripeEventRouter.cs
jar-stripe: We should probably check this in the other SDKs too; we're not always consistent about which request objects are reused vs recreated in different scenarios.

src/Stripe.net/Infrastructure/Public/StripeEventNotificationHandler.cs
```diff
@@ -0,0 +1,800 @@
...
+                }
+                else if (eventNotification is Stripe.Events.V2MoneyManagementTransactionUpdatedEventNotification)
+                {
+                    this.v2MoneyManagementTransactionUpdatedEventNotification.Invoke(this, new StripeEventNotificationEventArgs<Stripe.Events.V2MoneyManagementTransactionUpdatedEventNotification>((Stripe.Events.V2MoneyManagementTransactionUpdatedEventNotification)eventNotification, this.client));
+                }
+
+                // event-handler-dispatch: The end of the section generated from our OpenAPI spec
+                else
+                {
+                    throw new Exception("unexpected state, please file a bug");
```
jar-stripe, line 812: this message needs some love 😆

jar-stripe: but actually, in all seriousness could this just fall thru to the UnhandledEventHandler?  at least then the user might see it and have context in which to raise an issue back to us and/or recover on their side if we have to fix something.

justfile
xavdid-stripe: I don't disagree, but I don't have `$DOTNET_ROOT` set and everything works fine except that this install script fails 😅 I can pull it though

src/Stripe.net/Infrastructure/Public/StripeEventNotificationHandler.cs
xavdid-stripe: it could, but it's indicative of a codegen bug. it means we generated a registration method for an event type, but not the matching if-branch. Not to say it _can't_ happen, it just seems unlikely anyone will hit this. I can put something that links to the GH if you prefer though.

xavdid-stripe:
ok @jar-stripe this is ready for re-review!

My biggest open question is about supporting async - in Node, all of the callbacks return promises for easier integration with making stripe client calls. In python we haven't added support for async (yet) since i'm not positive how we'll do it. For dotnet, I know users can call async functions syncronously. So, is it worth having `Handle` be async and all of the callbacks be too?

Also re: naming. Dotnet having a built-in event thing called a "handler" (that other languages call a "callback") muddles our naming a bit. Is it worth diverging from the other languages here for consistency? Basically, the fill-in-the-blank exercise we did for the other languages is:

> A user instantiates a _____, registers 1+ ______, and then invokes that [1st blank] from their (framework-specific) webhook endpoint.

Today, it's

> A user instantiates an event (notification) handler, registers 1+ callback functions, and then invokes that handler from their (framework-specific) webhook endpoint.

But should dotnet be different?

src/Stripe.net/Infrastructure/Public/StripeEventNotificationHandler.cs
jar-stripe: ahh i see; i think the handledEventTypes name was throwing me off; yea i think expanding on the message in all directions would be good e.g. "we are unable to process the event x.y.z, this is likely a bug in the SDKs.  please file an issue in the stripe-dotnet github respository", or something like that.  as a developer id be able to see what was missed and what I should do about it.

PR #3359, opened 2026-04-01
Add test coverage for 2D list form encoding

jar-stripe:
### Why?

The content encoder correctly handles nested lists (`List<List<T>>`) via recursive dispatch through `FlattenParamsList`. However, there was no explicit test for this behavior, making it possible for a future refactor to break it silently.

### What?

- Add nested `List<object>` test case to `ContentEncoderTest.cs` verifying `list[0][0]=1&list[0][1]=2&list[1][0]=3&list[1][1]=4` encoding

### See Also

- https://jira.corp.stripe.com/browse/RUN_DEVSDK-2304
- Bug fix PRs in stripe-ruby, stripe-python, stripe-php
- Test-only PR in stripe-go


prathmesh-stripe (approved):
Nice! Thanks for adding the tests.

Issue #3364, opened 2026-04-11
Serialization issues in version 51.0.0


mirlayne-ia:
### Describe the bug

After updating to version 51.0.0, when trying to create a Setup Intent we get an error and the operation is not performed. The `SetupIntentCreateOptions` is created with the property `MandateData = null` and `Usage = "off_session"`.

Below I have pasted the exact error:
`Stripe.StripeException: You passed an empty string for 'mandate_data'. We assume empty values are an attempt to unset a parameter; however 'mandate_data' cannot be unset. You should remove 'mandate_data' from your request or supply a non-empty value.`

### To Reproduce

1. Update to version 51.0.0
2. Try to create a Setup Intent without confirmation. Use `SetupIntentCreateOptions` with the following properties values `MandateData = null` and `Usage = "off_session"`

### Expected behavior

The SetupIntent is created successfully as in versions prior to version 51.0.0

### Code snippets

```csharp
`SetupIntentCreateOptions setupIntentCreateOptions = new()
{
    Customer = input.StripeCustomerId,
    PaymentMethodTypes =
    [
        "card"
    ],
    Usage = "off_session"
};`
```

### OS

macOS

### .NET version

.NET 10

### Library version

stripe-dotnet v51.0.0

### API version

2026-03-25.dahlia

remi-stripe:
@mirlayne-ia Thanks for the report. This is expected behaviour in that latest major as it's one of the breaking changes. See [this entry](https://github.com/stripe/stripe-dotnet/wiki/Migration-guide-for-v51#clearing-fields-with-explicit-null-assignment) of the migration guide which explains the change. I recommend that you remove `MandateData = null` entirely from the code you shared in this case.

mirlayne-ia:
Hello @remi-stripe 

Thank you for the quick response. After reading the documentation I understood the idea behind the change, however let me explain the situation in detail:

1- I'm creating the `SetupIntentCreateOptions` object without the `MandateData` property in a microservice **A** as follows:
`SetupIntentCreateOptions setupIntentCreateOptions = new()
{
    Customer = input.StripeCustomerId,
    PaymentMethodTypes =
    [
        "card"
    ],
    Usage = "off_session"
};`

2- Microservice **A** invokes an AWS Lambda function in microservice **B** sending as input the `SetupIntentCreateOptions`. Internally AWS serialises the object. This is the payload received in **B** with field `"mandate_data": null`: 


`{
  "attach_to_self": null,
[...]

PR #3366, opened 2026-04-14
Fix emptyable property serialization for third-party serializers

jar-stripe:
### Why?

See #3364. When a third-party serializer (e.g. AWS Lambda) serializes an Options object, unset emptyable properties were written as `"mandate_data": null`. On deserialization the setter fires, `SetTracker.Track()` marks the property as explicitly set, and the SDK sends `mandate_data=` to the API — unintentionally clearing fields.

The risk extends beyond Create calls — on Update APIs this could silently clear fields like `invoice_settings` without the user realizing.

### What?

- Regenerated Options classes with `NullValueHandling.Ignore` (Newtonsoft) and `WhenWritingNull` (STJ) annotations on emptyable properties, so third-party serializers skip unset nulls
- Refactored `STJDefaultConverter.Write` null-handling to dispatch on the property's ignore condition: `WhenWritingNull` properties only write null when `IsPropertySet` is true
- Added round-trip serialization tests covering both Newtonsoft and STJ paths
- Updated `TestOptions` to match the new codegen output

## Changelog

* Fixes a bug where serializing Options objects through a third-party serializer (e.g. AWS Lambda) could cause the SDK to unintentionally clear fields on the API. Emptyable properties now have null-skipping annotations so third-party serializers omit unset properties during round-trips.


xavdid (approved):
fix looks good, but make sure to call out this behavior / pitfalls in the README!

PR #3381, opened 2026-05-08
Migrate formatting from dotnet-format to CSharpier

mbroshi-stripe:
### Why?

`dotnet format` requires package restore and full compilation before formatting, making it slow (minutes) and prone to OOM kills/timeouts in CI. CSharpier parses C# syntax directly without compilation and formats the full repo (~5400 files) in ~3 seconds.

### What?

- Added CSharpier 1.0.3 as a local dotnet tool (`.config/dotnet-tools.json`)
- Added `.csharpierrc.yaml` with `printWidth: 100` (matches existing `.vscode/settings.json` ruler)
- Updated `just format` and `just format-check` to use CSharpier
- Applied CSharpier formatting to all source files (purely whitespace changes)


jar-stripe:
Closing stale PR.  We may pick this back up in the future, but CSharpier is not a drop in replacement for `dotnet format` and we can also speed up format performance by changing how we generate code.

PR #3385, opened 2026-05-21
Emit warning when `stripe-notify` header is present in response

xavdid:
### Why?

We've noticed a trend where users (and their agents) will sometimes reach for suboptimal API endpoints/methods because of outdated information. For example, using `/v1/accounts` when `/v2/core/accounts` is available and has a superset of the functionality from v1.

In an effort to help steer users towards best practices, we're introducing a new header to provide feedback during the development process. The SDK is responsible for surfacing this header, if present.

Once enabled internally, the header will be returned by the server for:

- all testmode calls
- livemode calls made by an agent

We may tweak that going forwards. But no matter the conditions, the SDKs will act as a thin pipe.

### What?

- Emit warning when we see the `Stripe-Notice` header in a response. **Note** this is the first warning we're emitting in this SDK, so make sure it's the correct way to do that!
- add tests

### See Also

- https://go/agent-steering
- [DEVSDK-2430](https://go/j/browse/DEVSDK-2430)

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs
+++ b/src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs
@@ -2,10 +2,12 @@ namespace Stripe
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
     using System.IO;
     using System.Linq;
     using System.Net;
     using System.Net.Http;
+    using System.Net.Http.Headers;
     using System.Text.Json;
     using System.Threading;
     using System.Threading.Tasks;
@@ -225,6 +227,18 @@ namespace Stripe
             }
         }
 
+        private static void MaybeEmitStripeNotice(HttpResponseHeaders headers)
+        {
+            if (headers != null && headers.Contains("Stripe-Notice"))
+            {
+                var notice = headers.GetValues("Stripe-Notice").FirstOrDefault();
+                if (notice != null)
+                {
+                    Trace.TraceWarning(notice);
+                }
+            }
+        }
+
         // Note: BaseOptions options really means query params here
         private StripeRequest MakeStripeRequest(
             BaseAddress baseAddress,
@@ -273,6 +287,8 @@ namespace Stripe
 
                 throw BuildStripeException(response);
             }
+
+            MaybeEmitStripeNotice(response.Headers);
 
             T obj;
             try
@@ -361,6 +377,8 @@ namespace Stripe
             {
                 throw BuildStripeException(response);
             }
+
+            MaybeEmitStripeNotice(response.Headers);
 
             return response;
         }
--- a/src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs
+++ b/src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs
@@ -2,7 +2,6 @@ namespace Stripe
 {
     using System;
     using System.Collections.Generic;
-    using System.Diagnostics;
     using System.IO;
     using System.Linq;
     using System.Net;
@@ -232,9 +231,9 @@ namespace Stripe
             if (headers != null && headers.Contains("Stripe-Notice"))
             {
                 var notice = headers.GetValues("Stripe-Notice").FirstOrDefault();
-                if (notice != null)
+                if (!string.IsNullOrEmpty(notice))
                 {
-                    Trace.TraceWarning(notice);
+                    Console.Error.WriteLine(notice);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs
```diff
@@ -225,6 +227,18 @@ private static StripeException BuildV2StripeException(StripeResponse response)
             }
         }
 
+        private static void MaybeEmitStripeNotice(HttpResponseHeaders headers)
+        {
+            if (headers != null && headers.Contains("Stripe-Notice"))
+            {
+                var notice = headers.GetValues("Stripe-Notice").FirstOrDefault();
+                if (notice != null)
```
jar-stripe: What does it mean if the header is present but empty?  Is that something to flag?

src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs
```diff
@@ -225,6 +227,18 @@ private static StripeException BuildV2StripeException(StripeResponse response)
...
+                {
+                    Trace.TraceWarning(notice);
```
jar-stripe: The tests show a trace listener being registered before firing this; does this need some kind of registered listener to work at all?  Or will it go to stdout or stderr absent a listener?

jar-stripe (changes requested):
The code looks good; the changes I'd like to see are at the PR level, specifically, can you attach results of a test run that shows what an agent will see when this is hit?  Even better if we can verify that the agent will see that on the correct output channel.

src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs
xavdid: that would be a malformed shape from the server, but claude was just being a little extra defensive here. Can remove the check or use or use `!IsNullOrEmpty` instead. I don't think we have to code around it though, mostly.

In other languages we just check if the header is present at all and could probably do so here.

src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs
xavdid: from claude:

> Regarding the output channel: `Trace.TraceWarning` routes to `Trace.Listeners`. By default, `DefaultTraceListener` directs output to the debugger output window — it does **not** write to stderr. An application wanting stderr output would need to add a `TextWriterTraceListener(Console.Error)`. If there's a preference for stderr by default, I can add `Trace.Listeners.Add(new TextWriterTraceListener(Console.Error))` guidance to the PR description or even make that the implementation. But `Trace` is the idiomatic .NET mechanism for this pattern — it's what Microsoft recommends for SDK-level diagnostics, and it integrates naturally with whatever logging infrastructure the host has configured (e.g., `TraceSource` listeners, ASP.NET Core's `DiagnosticsEventListenerObserver`, etc.).

I don't have a lot of context around if/how an integration would have their error logging set up. I do see there's a `Console.Error` method though, so maybe we should print to that instead. In either case, it seems like this is the first place we're doing any sort of warning in the SDK, so I defer to you on what devs will be most receptive to.

It may be academic though, since I don't imagine many agents are setting up dotnet projects to do small stripe write operations. IMO this change is most useful in python + node (the most common steps up from a series of curl calls), where we print to stderr by default.

jar-stripe: in .NET the debug channel is only active if a debugger is attached, which wont likely show up if its invoked from an agent.  We should probably have this write to stderr so that the agent can watch for the warning

src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs
```diff
@@ -225,6 +226,18 @@ private static StripeException BuildV2StripeException(StripeResponse response)
             }
         }
 
+        private static void MaybeEmitStripeNotice(HttpResponseHeaders headers)
+        {
+            if (headers != null && headers.Contains("Stripe-Notice"))
```
jar-stripe, line 231: sanity check: are headers expected to be cased like this all the time?

jar-stripe (approved):

src/Stripe.net/Infrastructure/Public/LiveApiRequestor.cs
xavdid: yes and no- the server is going to send it exactly like that. But also, header lookups in most languages/frameworks (dotnet included) are case insensitive! so we're all good.

PR #3394, opened 2026-06-03
Add "source" field to user-agent header

xavdid:
### Why?

Earlier this year, we removed the output of `uname` from the user-agent header for privacy reasons. It over-collected data we usually didn't need.

Since then, our tech support team has reached out asking us to re-add some of the information to help with complex debugging tasks: sometimes they need to help users identify exactly which machine produced an API call.

Rather than just revert the previous change and reintroduce the same privacy issues that caused its removal in the first place, we're taking a more privacy-focused approach. Rather than send all the PII in plaintext, we hash it first. This way it can still be used to confirm the source of an API call if a user generates the hashes themselves, but isn't sensitive in and of itself.

### What?

- add `source` key to user-agent hash. It's contents are an md5 hash of `uname -a`
- add tests

### See Also

- [DEVSDK-3131](https://go/j/DEVSDK-3131)

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
+++ b/src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
@@ -8,6 +8,8 @@ namespace Stripe
     using System.Net;
     using System.Net.Http;
     using System.Net.Http.Headers;
+    using System.Security.Cryptography;
+    using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
     using Newtonsoft.Json;
@@ -133,6 +135,9 @@ namespace Stripe
         /// </summary>
         public static TimeSpan MinNetworkRetriesDelay => TimeSpan.FromMilliseconds(500);
 
+        /// <summary>Gets or sets the cached source hash. Internal for testing.</summary>
+        internal static string SourceHash { get; set; } = ComputeSourceHash();
+
         /// <summary>
         /// Gets whether telemetry was enabled for this client.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
```diff
@@ -297,6 +331,10 @@ private string BuildStripeClientUserAgentString()
                 { "lang", ".net" },
                 { "stripe_net_target_framework", StripeNetTargetFramework },
             };
+            if (!string.IsNullOrEmpty(SourceHash))
+            {
+                values["source"] = SourceHash;
```
jar-stripe, line 336: are we assuming a missing key means it was omitted due to error or missing value?

src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
```diff
@@ -133,6 +135,9 @@ public SystemNetHttpClient(
         /// </summary>
         public static TimeSpan MinNetworkRetriesDelay => TimeSpan.FromMilliseconds(500);
 
+        /// <summary>Gets or sets the cached source hash. Internal for testing.</summary>
```
jar-stripe: i think this comment is a bit misleading; its not just for testing, its also cached for use when preparing requests, right?

jar-stripe (approved):
sanity check and one comment to fix, but otherwise looks good

src/Stripe.net/Infrastructure/Public/SystemNetHttpClient.cs
xavdid: yeah, exactly. that was the pattern we followed in other languages- we include the key as a "best effort"

PR #3408, opened 2026-07-10
Make Error fields generated

xavdid:
### Why?

Our main API Error class has always been handwritten. As new fields have been added to `api_errors` in the spec over time (e.g. `advice_code`, `network_advice_code`, `network_decline_code`), we haven't picked them up in the SDKs.

This makes most of those properties generated so they stay in sync.

### What?

- Makes the fields on the API error object generated from the OpenAPI spec
- Adds section markers so codegen can maintain these fields going forward
- add tests

### See Also

- [RUN_DEVSDK-2568](https://go/j/RUN_DEVSDK-2568)
- https://github.com/stripe/stripe-go/issues/2384

### DIFF
--- a/src/StripeTests/Infrastructure/StripeExceptionTest.cs
+++ b/src/StripeTests/Infrastructure/StripeExceptionTest.cs
@@ -1,7 +1,7 @@
 namespace StripeTests
 {
+    using System.Text.Json;
     using System.Threading.Tasks;
-    using Newtonsoft.Json;
     using Stripe;
     using Xunit;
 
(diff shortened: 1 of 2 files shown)


src/StripeTests/Infrastructure/StripeExceptionTest.cs
```diff
@@ -1,6 +1,7 @@
 namespace StripeTests
 {
     using System.Threading.Tasks;
+    using Newtonsoft.Json;
```
jar-stripe: can you change this to use System.Text.Json?  Newtonsoft is there for compatibility, but all of our tests should use STJ

jar-stripe (changes requested):

src/StripeTests/Infrastructure/StripeExceptionTest.cs
xavdid: fixed!

jar-stripe (approved):

PR #3409, opened 2026-07-10
Replace source hash with Telemetry UUID

xavdid:
### Why?

Recently, we introduced the `source` hash in the user-agent header to help our support team track down exactly where incoming requests originated from.

They've reported back to us that this hasn't been as useful as we wanted, so we're going a different direction.

Instead, we're adopting an industry-standard method of persisting a small UUID on disk that we include with subsequent requests. It's easy for users to read this file and it contains no PII. It also honors our existing telemetry opt outs.

It's meant to be lightweight and failure tolerant.

### What?

- remove `source` key from user-agent header
- add `telemetry_id` to the user-agent header
- adjust tests accordingly
- (unrelated) remove unused coverage badge from readme

### See Also

- original request: [DEVSDK-3131](https://go/j/DEVSDK-3131)
- update request: [RUN_DEVSDK-2563](https://go/j/RUN_DEVSDK-2563)

### DIFF
--- a/src/Stripe.net/Infrastructure/TelemetryId.cs
+++ b/src/Stripe.net/Infrastructure/TelemetryId.cs
@@ -11,6 +11,9 @@ namespace Stripe.Infrastructure
         private static string cachedId;
         private static bool loaded;
 
+        // For testing: overrides GetConfigDir() when set.
+        internal static string ConfigDirOverride { get; set; }
+
         public static string Get()
         {
             if (loaded)
@@ -64,7 +67,7 @@ namespace Stripe.Infrastructure
 
         private static string Resolve()
         {
-            var configDir = GetConfigDir();
+            var configDir = ConfigDirOverride ?? GetConfigDir();
             if (configDir == null)
             {
                 return null;
@@ -122,6 +125,7 @@ namespace Stripe.Infrastructure
             {
                 cachedId = null;
                 loaded = false;
+                ConfigDirOverride = null;
             }
         }
     }
--- a/src/Stripe.net/Infrastructure/TelemetryId.cs
+++ b/src/Stripe.net/Infrastructure/TelemetryId.cs
@@ -1,0 +1,128 @@
+namespace Stripe.Infrastructure
+{
+    using System;
+    using System.IO;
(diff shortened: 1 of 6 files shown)


src/Stripe.net/Infrastructure/TelemetryId.cs
```diff
@@ -0,0 +1,132 @@
...
+                }
+
+                cachedId = Resolve();
+                loaded = true;
+            }
+
+            return cachedId;
+        }
+
+        internal static string GetConfigDir()
```
jar-stripe, line 38: im wondering if maybe this should be its own utility class; maybe not blocking for this PR but worth a jira ticket, for if/when we need to store other information locally?

src/StripeTests/Infrastructure/TelemetryIdTest.cs
```diff
@@ -0,0 +1,158 @@
...
+                {
+                    Directory.Delete(tmpDir, recursive: true);
+                }
+            }
+        }
+
+        [Fact]
+        public void GetConfigDir_UnixUsesXdgConfigHomeWhenSet()
+        {
+            if (Environment.OSVersion.Platform == PlatformID.Win32NT)
```
jar-stripe, line 203: another follow up: now that we're adding in OS specific behavior we should consider CI tests that test on windows

jar-stripe (approved):
Looks good!  A couple of follow up things, but overall 👍

src/StripeTests/Infrastructure/TelemetryIdTest.cs
xavdid: at least for dotnet, we have that! https://github.com/stripe/stripe-dotnet/blob/f850f16cb7d390ada5d34ea780d02488b03ad689/.github/workflows/ci.yml#L146

that's a good suggestion for other langs though.

src/Stripe.net/Infrastructure/TelemetryId.cs
xavdid: I think let's move it if/when we need to!

PR #3413, opened 2026-07-21
Add discriminated union converter infrastructure

jar-stripe:
### Why?

Adds discriminated union infrastructure to stripe-dotnet: both the runtime converters for polymorphic JSON deserialization and comprehensive tests validating serialization in both directions.

### What?

**Runtime infrastructure:**
- `DiscriminatedUnionConverter` (Newtonsoft): Attribute-based polymorphic JSON reader — reads the discriminator field from the JSON, looks up the matching variant class via `[StripeUnionVariant]` attribute, and deserializes into the correct concrete subclass
- `STJDiscriminatedUnionConverterFactory` (System.Text.Json): Same dispatch logic for STJ
- `StripeDiscriminatedUnionAttribute`: Marks a class as a DU base type, specifying the discriminator property name. Targets both interfaces (request-side) and classes (response-side), with `Inherited = false` to prevent subclass inheritance
- `StripeUnionVariantAttribute`: Marks a variant class with its discriminator value. Same target/inheritance settings

**Tests (3 test files):**
- `DiscriminatedUnionConverterTest` (Newtonsoft): 5 tests — variant dispatch by discriminator, unknown discriminator fallback, nested DU in container
- `STJDiscriminatedUnionConverterTest` (System.Text.Json): Same 5 tests against the STJ converter
- `DiscriminatedUnionSerializationTest`: 6 tests covering request standalone (Options interface pattern), request inline (discriminator + variant fields), response standalone (base class + variant subclasses via converter), and response inline (flat fields with nullable variants)

All test fixtures use the class hierarchy pattern that codegen actually produces (base class with `[StripeDiscriminatedUnion]` + variant subclasses with `[StripeUnionVariant]`).

## Changelog

- Added `DiscriminatedUnionConverter` and `STJDiscriminatedUnionConverterFactory` for attribute-based polymorphic JSON deserialization of discriminated union types. Added `[StripeDiscriminatedUnion]` and `[StripeUnionVariant]` attributes for annotating discriminated union type hierarchies.


src/Stripe.net/Infrastructure/JsonConverters/DiscriminatedUnionConverter.cs
```diff
@@ -0,0 +1,102 @@
...
+        private static readonly ConcurrentDictionary<Type, UnionMetadata> MetadataCache = new ConcurrentDictionary<Type, UnionMetadata>();
+
+        public override bool CanWrite => false;
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            throw new NotSupportedException(
+                "DiscriminatedUnionConverter should only be used while deserializing. " +
+                "Serialization is handled by the concrete variant type's default serializer.");
+        }
```
zacchua-stripe, line 26: Do we still need to override `WriteJson` given that `CanWrite` is false?

zacchua-stripe (approved):
Looks good! Just had a non-blocking question.

src/Stripe.net/Infrastructure/JsonConverters/DiscriminatedUnionConverter.cs
jar-stripe: We do, unfortunately; its abstract on JsonConverter so we have to give it some implementation.

PR #3431, opened 2026-08-22
Add `.preHandle` method to `EventNotificationHandler`

xavdid:
### Why?

In final testing, users have expressed interest in being able to centrally run code before any other handler executes. It can be used for centralized logging, event deduplication, and more. I don't want to go full middleware, but this felt like a reasonable way to dip our toe into the waters.

Because users are responsible for [deduplicating events](https://docs.stripe.com/webhooks#handle-duplicate-events), this pre-handle hook (if present) should return a `bool`, signifying whether handling should continue.

If the function sets `.Cancel = false` (or is missing entirely), the handler proceeds like normal (running at-most-one other callback). If it sets `.Cancel = true`, no further callback is run and the original `.handle()` call finishes cleanly.

### What?

- add the `.pre_handle` method
- & associated tests
- update some error messages with correct verbs & consistency
- refactored some internal error handling
- updated example
- dotnet: reject nil handler case and and test

### See Also

- http://go/j/DEVSDK-3249

### DIFF
--- a/src/Examples/V2/EventNotificationHandlerEndpoint.cs
+++ b/src/Examples/V2/EventNotificationHandlerEndpoint.cs
@@ -17,8 +17,7 @@ namespace Examples.V2
     ///     - write a fallback callback to handle unrecognized event notifications
     ///     - create a StripeClient called client
     ///     - Initialize an EventNotificationHandler with the client, webhook secret, and fallback callback
-    ///     - register a PreHandle hook that deduplicates events we've already processed, so retried
-    ///       deliveries don't trigger the callback (or the fallback) a second time
+    ///     - register a PreHandle hook that deduplicates events by id before any callback runs
     ///     - register a specific handler for the "v1.billing.meter.no_meter_found" event notification type
     ///     - use handler.handle() to process the received notification webhook body.
     /// </summary>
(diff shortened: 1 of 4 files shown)


src/Examples/V2/EventNotificationHandlerEndpoint.cs
```diff
@@ -31,17 +33,38 @@ public class EventNotificationHandlerEndpoint : ControllerBase
...
         public EventNotificationHandlerEndpoint()
         {
             client = new StripeClient(Environment.GetEnvironmentVariable("STRIPE_API_KEY"));
             handler = client.NotificationHandler(Environment.GetEnvironmentVariable("WEBHOOK_SECRET") ?? string.Empty, FallbackCallback);
             unverifiedHandler = client.NotificationHandlerWithoutVerification(FallbackCallback);
 
-            // register handlers
+            // PreHandle runs after Handle parses the payload but before any callback fires.
+            // Returning false skips both the registered handler and the fallback for this event.
+            handler.PreHandle(SkipAlreadyProcessedEvents);
```
jar-stripe: I would suggest using a property, e.g. `handle.PreHandle = SkipAlreadyProcessedEvents`, and then implement that property's `get` to perform the same steps as the PreHandle method.

jar-stripe (changes requested):

src/Examples/V2/EventNotificationHandlerEndpoint.cs
jar-stripe: This is a medium held opinion; I can be convinced otherwise but I think its better ergonomics to use a property.

xavdid: fair, i've updated! 

one question though- should it use the same `+=` approach as the individual event handlers? In the other languages I made sure that the UX around registering the pre handler was identical to event types. Is it weird that these diverge?

jar-stripe: Actually, that would probably make sense.  My initial thought was it could go either way and wouldnt be weird, but the .NET core classes uses event semantics for lifecycle callbacks all the time.

xavdid: ok, updated! I like that much more. Feels more consistent with the rest of the class.

`.Cancel` is cool, too!

jar-stripe (approved):

PR #3441, opened 2026-08-28
Harden API requestor code against malicious URLs

xavdid:
### Why?

If an attacker got access to a user's webhook secret (or the user wasn't validating incoming webhook signatures) _and_ a user was calling the built-in `event_notification.fetch_related_object()` method, a malicious webhook payload could cause a user integration to make an authenticated request to an attacker-controlled domain (leaking the secret key)

To solve, we're being more careful during url construction to ensure user input can't be used to send requests to non-`stripe.com` domains (on a per-request basis).

### What?

- add path validation method and call it before the final URL is built
- add tests

### See Also

- [RUN_DEVSDK-2808](https://go/j/RUN_DEVSDK-2808)

### DIFF
--- a/src/Stripe.net/Infrastructure/Public/StripeRequest.cs
+++ b/src/Stripe.net/Infrastructure/Public/StripeRequest.cs
@@ -139,6 +139,41 @@ namespace Stripe
                 this.Uri.ToString());
         }
 
+        /// <summary>
+        /// Asserts that a request path is origin-relative: that it begins with a single
+        /// <c>"/"</c>.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// The absolute URL is built by concatenating a base URL onto this path, and no base URL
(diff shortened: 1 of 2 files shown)


src/Stripe.net/Infrastructure/Public/StripeRequest.cs
```diff
@@ -139,13 +139,43 @@ public override string ToString()
                 this.Uri.ToString());
         }
 
+        /// <summary>
+        /// Asserts that a request path is origin-relative: that it begins with a single
+        /// <c>"/"</c>.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// The absolute URL is built by concatenating a base URL onto this path, and no base URL
```
jar-stripe, line 148: similar to comment in stripe-java, this feels like a lot of exposition for ValidatePath.  is there a better place to put this?

jar-stripe (approved):
approving to unblock, but consider relocating the content of that long ValidatePath comment

